PR #16, opened 2018-10-30
Add EditorConfig

### LINKED ISSUE #15: Add an EditorConfig
We should add an [EditorConfig](https://editorconfig.org/) to this project. I suggest we clone the one used by corefx or such.

markusweimer:
This adds a `.editorconfig` to the root of the repo. To get us started,
I copied the one found here:

https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference?view=vs-2017#example-editorconfig-file

This closes #15


migueldeicaza:
This needs to be verified that it does not reformat the code.

Issue #18, opened 2018-10-31
Ponder: Tensor type?


migueldeicaza:
Perhaps there should be a `TorchSharp.Tensor` type that proxies to the right storage tensor, with an abstract interface, so that people can write generic-ish code that deals with Tensors, rather than having different data types.

Bonus points - even better would be to have the `Tensor` not surface a `Tensor<T>`, as that would defeat the reusability at that point.   This would have the downside that operations would have to dynamically check for type compatibility.

dsyme:
We can close this, as there is a TorchTensor type?

interesaaat:
I think that this issue was about adding generics to the `Tensor` type. I tried and failed miserably. I think we can close this if we think that generics are not necessary.

migueldeicaza:
The work on @PartyDonk could have helped.

PR #14, opened 2018-10-30
Added constructor for multidimensional tensors with params long[]

interesaaat:
Fix #9

PR #13, opened 2018-10-30
[WIP] SNT Wrapper for TorchTensors

interesaaat:
NativeMemoryManager.cs is taken from SNT/tests

PR #28, opened 2018-11-01
[WIP] Adding Generic abstract class for tensors

interesaaat:
Trying to solve #18.

This will also help with PR #13 to avoid `object` type for inner tensors. 

This PR will go after PR #14 is merged.

### DIFF
--- a/TorchSharp/TypeGeneration.cs
+++ b/TorchSharp/TypeGeneration.cs
@@ -4,7 +4,63 @@ using Microsoft.Win32.SafeHandles;
 using System.Text;
 
 namespace TorchSharp {
-    public partial class ByteTensor : IDisposable {
+public abstract class Tensor<T> : IDisposable
+    {
+        public SafeHandle handle;
+        public abstract void Dispose();
+        public abstract void ZeroFill();
+        public abstract void Fill(T value);
+        public abstract int Dimensions { get; }
+        public abstract long GetTensorDimension(int dim);
+        public abstract long[] Shape { get; }
+        public abstract long GetTensorStride(int dim);
+        public abstract unsafe void* Data { get; }
+        public abstract Tensor<T> Clone();
+        public abstract Tensor<T> Select(int dim, long slideIndex);
+        public abstract Tensor<T> Narrow(int dim, long firstIndex, long size);
+        public abstract Tensor<T> Transpose(int dim1, int dim2);
+        public abstract Tensor<T> Unfold(int dim, long size, long step);
+        public abstract void Resize1d(long size);
+        public abstract void Resize2d(long size0, long size1);
+        public abstract void Resize3d(long size0, long size1, long size2);
+        public abstract void Resize4d(long size0, long size1, long size2, long size3);
+        public abstract void Resize5d(long size0, long size1, long size2, long size3, long size4);
+        public abstract void ResizeAs(Tensor<T> src);
+        public abstract void Set(Tensor<T> src);
+        public abstract T this [long x0] { get; set; }
+        public abstract T this[long x0, long x1] { get; set; }
+        public abstract T this[long x0, long x1, long x2] { get; set; }
+        public abstract T this[long x0, long x1, long x2, long x3] { get; set; }
+        public abstract void Random(RandomGenerator source, long n);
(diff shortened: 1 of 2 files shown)


migueldeicaza:
I believe that this needs some discussion, which is why I had opened issue #18 as "Ponder".

I am not sure that `Tensor<T>` is the way to go, I mentioned that on #18, the reason being that Tensor<T> is not that different from DoubleTensor, in that it is difficult to change the code once a decision has made - the change has to take place everywhere.

The convenience of the Python API is that the tensor is typed at creation time, and checked when the operations are used, not at compile time.   So it is simple to change the models by merely changing the creation of the tensor, rather than changing every occurrence of the type (now, Pytorch also supports a global variable to change the default, but I think that is sort of gross).

One alternative would be to have a `Tensor` type, plus an additional `Tensor<T>`, to let folks pick their poison.

interesaaat:
> I believe that this needs some discussion, which is why I had opened issue #18 as "Ponder".
> 
> I am not sure that `Tensor<T>` is the way to go, I mentioned that on #18, the reason being that Tensor is not that different from DoubleTensor, in that it is difficult to change the code once a decision has made - the change has to take place everywhere.
> 
> The convenience of the Python API is that the tensor is typed at creation time, and checked when the operations are used, not at compile time. So it is simple to change the models by merely changing the creation of the tensor, rather than changing every occurrence of the type (now, Pytorch also supports a global variable to change the default, but I think that is sort of gross).
> 
> One alternative would be to have a `Tensor` type, plus an additional `Tensor<T>`, to let folks pick their poison.

Ok now I got your "even better would be to have the `Tensor` not surface a `Tensor<T>`". I would rather say that this is related to the fact that Python is dynamically typed. Yes it is definitely convenient being able to easily swap types, but my experience (or, I would say, what I learned from ML.NET folks) is that it is more practical having strictly typed code with errors thrown at compile time, rather then runtime errors (eventually thrown after minutes, especially for training).

migueldeicaza:
More often than not, the problem in these situations is mismatches on shape of the tensor, and we have no way of expressing this class of problems.   I agree with you that the type system would help, but in this case, it is catching a minor problem, while leaving the real problem unsolved.

interesaaat:
I think we might be able to address the shape problem using value tuples in c#. We can definitely specify labels for dimensions in this way, but probably also lengths and related operation checks.

migueldeicaza:
I do not understand what you have in mind, can you share?

interesaaat:
> I do not understand what you have in mind, can you share?

I was thinking about something like adding a second constraint type to the tensor class in order to force static type checking of operators. Something like `Tensor<T, U> where U : Tuple<Dimension>`. I tried to implement something but the number of classes blows up because we need to enumerate all possible number of dimensions (unless there is a way to just say that `U` is a tuple, without explicitly saying with how many elements. You can do this in [scala](https://github.com/ctongfei/nexus) for instance.)

NiklasGustafsson:
In order to capture shape as part of the type using tuples, you'd need something like C++'s variadic constructs, which isn't likely to happen in C#. Even Julia, which I consider almost ideal for numerics, doesn't try to make the actual shape part of the type, just the number of dimensions.

dpthayer:
This problem is effectively resolved by using a Dependently Typed Language of which one of the premier examples is C++.  If you build the Tensor Library as a C++ Class you could pass template parameters to it such as the Number of dimensions and the size of each dimension and have lightning fast tensor operations at the machine level.

dpthayer:
Something like the following
Tensor CreateTensor(uint dtype, uint rank, uint shape)
{
      //call to low level library
      return new NumericTenser<dtype, rank,shape>() as Tenser;
}

dsyme:
@interesaaat I think we can close this?  Adding the generic baseclass is not particularly easy.

It could be done, but most of the operations would be TorchTensor -> TorchTensor with no generics statically propagated.

dsyme:
p.s. I notice there was a use of SafeHandle which may be of interest here to put back into the code separately

interesaaat:
Yes let's close this one. Even because now `TorchTensor` is a `struct` for performance reasons.

Issue #34, opened 2018-11-02
Revamp folder structure


NiklasGustafsson:
We should have a traditional src/test/docs structure.

interesaaat:
Plus namespace names. Should / can we have Microsoft.ML as in ML.NET?

markusweimer:
> Plus namespace names. Should / can we have Microsoft.ML as in ML.NET?

I don't think that is a good idea. Success to me for this project would be acceptance into the PyTorch code base. Maybe we can use `Torch` as the prefix?

migueldeicaza:
Few topics:

Agreed, we do not want "Microsoft" on the prefix.   Makes sense to use "Torch" instead of "TorchSharp", let us do that.

As for the layout of the project, you probably mean the "corefx" lineage, which is only traditional as "in the last couple of years", the tradition of cross platform .NET goes some 17 years back and that is a mixed bag.

I do not mind creating a toplevel for "test" and adding or moving various tests there, the rest I would keep as is.

In particular, "docs" is harder, given that I am publishing the API documentation from GitHub, the formatted docs have to go there, while the source documentation has to go elsewhere.

interesaaat:
I think that all docs related can stay as they are. However it is probably better to factor source code into their proper folders: all tests into the test dir, and the remaining code into src. I also suggest to rename Tester to Sample. 

So the final structure I have in mind is
/docs
/docfx
/ecmadocs
/test
/src
/src/TorchSharp
/src/Sample
(/src/TorchTensor) -> this is the mapping to SNT

For the namespace, I think we should have Torch as root. So TorchSharp will become Torch.TorchSharp (or Torch.Sharp?), and the bindings to SNL as Torch.Sharp.Tensor (I am open to suggestions here).

WDYT?

motus:
@interesaaat looks good - although we have to double check the location of `test/` folder - it can be e.g. under `/src/`. Also, do we plan to keep using `Microsoft.VisualStudio.TestTools.UnitTesting` or switch to, say, `XUnit` or soething?

interesaaat:
`test` should be its own folder IMO.  `UnitTesting` is ok since we already started with that.

interesaaat:
@NiklasGustafsson any specific reason for having the test project not being part of the solution?

PR #51, opened 2018-11-06
Dense SNT Wrapper for TorchTensors

### LINKED ISSUE #3: Implement System.Numerics.Tensor using libtorch
`System.Numerics.Tensor` is the tensor type of .NET. We should have an implementation of it using `libtorch`.

interesaaat:
This closes #3. We will need a different issue for the operators. This PR *only* extends the methods in the 
 `DenseTensor<T>` class using Torch tensors as backends.


SNT/NativeMemoryManager.cs
```diff
@@ -0,0 +1,112 @@
+﻿using System.Buffers;
+using System.Runtime.InteropServices;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System;
+
+namespace Torch.SNT
```
NiklasGustafsson: Should we be using non-descriptive names like 'SNT' in namespaces? It does not aid in discovery.

SNT/NativeMemoryManager.cs
```diff
@@ -0,0 +1,112 @@
...
+            return new NativeMemory<T>(memory, length);
+        }
+
+        public bool IsDisposed { get; private set; } = false;
+
+        public unsafe override Span<T> GetSpan() => new Span<T>((void*)memory, length);
+
+        protected bool IsRetained => refCount > 0;
+
+        public override MemoryHandle Pin(int elementIndex = 0)
```
NiklasGustafsson: Doc comments on all public methods, please.

SNT/SNT.csproj
```diff
@@ -0,0 +1,21 @@
...
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <LangVersion>7.1</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Numerics.Tensors" Version="0.1.0-preview2-181031-3" />
+    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.2" />
```
NiklasGustafsson: Since I don't do unsafe C# too often, I keep forgetting -- does it not complicate using the package from applications/services that don't allow unsafe code? Is there a way to avoid that?

SNT/TypeGenerator.tt
```diff
@@ -0,0 +1,157 @@
...
+    new { ApiName = "Byte", Storage = "byte", AccReal="long" }, 
+    new { ApiName = "Short", Storage = "short", AccReal="long" },   
+    new { ApiName = "Int", Storage = "int",  AccReal="long"},
+    new { ApiName = "Long", Storage = "long", AccReal="long" },
+    new { ApiName = "Double", Storage = "double", AccReal="double" },
+    new { ApiName = "Float", Storage = "float", AccReal="double" },
+     }) {
+        string tname = type.ApiName;
+#>
+    public sealed class <#=tname#>TorchTensor : DenseTensor<<#=tname.ToLower()#>>
```
NiklasGustafsson: Even if we don't split up the Razor template file, maybe we should avoid adding new logic to the existing one?

SNT/Utils.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            if (dimensions.Length == 0)
+            {
+                return 0;
+            }
+
+            int product = 1;
+            for (int i = startIndex; i < dimensions.Length; i++)
+            {
+                if (dimensions[i] < 0)
```
NiklasGustafsson: Can we avoid this check by using unsinged integers for the dimensions?

Test/FloatTorchTensorUnitTest.cs
```diff
@@ -0,0 +1,368 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
```
NiklasGustafsson: The unit tests look good.

SNT/TypeGenerator.tt
```diff
@@ -0,0 +1,157 @@
...
+            this.inner = inner;
+        }
+
+        /// <summary>
+        ///   Utility method to create a TorchTensor.
+        ///   This is currently failing if the input parameter is empty because SNT 
+        ///   does not support zero-size tensors.
+        /// </summary>
+        /// <param name="sizes">The desired sizes for the dimensions of the tensor.</param>
+        public static unsafe <#=tname#>TorchTensor Create(params int[] sizes)
```
NiklasGustafsson: See my earlier question about using unsafe code.

SNT/NativeMemoryManager.cs
interesaaat: I agree. Any suggestion? I can definitely unroll SNT as SystemNumericTensor but it becomes really long.

SNT/TypeGenerator.tt
interesaaat: I didn't get this. This TypeGenerator is different than the one in TorchSharp and is used to generate SNT wrappers around the original torch tensors.

Test/FloatTorchTensorUnitTest.cs
interesaaat: Thanks :)

SNT/SNT.csproj
interesaaat: Mmm I don't think I can avoid this. I need to map the unmanaged memory from TorchSharp into C# stuff; I don't think I can avoid the usage of pointers.

SNT/TypeGenerator.tt
interesaaat: Unfortunately I don't think I can void unsafe.

SNT/SNT.csproj
interesaaat: Unsafe is also already used in TorchSharp when returning the pointer to data.

SNT/NativeMemoryManager.cs
interesaaat: Fixed

SNT/Utils.cs
interesaaat: I think this is difficult to do, unless we can convince the SNT folks to use unsigned ints in their API. (Torch is actually using long BTW)

SNT/NativeMemoryManager.cs
```diff
@@ -0,0 +1,114 @@
...
+
+namespace Torch.SNT
+{
+    /// <summary>
+    ///   Class used to wrap the native memory of a Torch Tensor.  
+    /// </summary>
+    internal class NativeMemory<T> : MemoryManager<T>
+    {
+        private int refCount = 0;
+        private IntPtr memory;
```
markusweimer: can this be `readonly`?

SNT/NativeMemoryManager.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        public NativeMemory(IntPtr memory, int length)
+        {
+            this.memory = memory;
+            this.length = length;
+        }
+
+        public unsafe NativeMemory(void* memory, int length)
+        {
+            this.memory = (IntPtr)memory;
```
markusweimer: Can this conversion fail? Also, this constructor could call the other one.

SNT/NativeMemoryManager.cs
```diff
@@ -0,0 +1,114 @@
...
+    /// <summary>
+    ///   Class used to wrap the native memory of a Torch Tensor.  
+    /// </summary>
+    internal class NativeMemory<T> : MemoryManager<T>
+    {
+        private int refCount = 0;
+        private IntPtr memory;
+        private int length;
+
+        public NativeMemory(IntPtr memory, int length)
```
markusweimer: This probably should do some defensive parameter checks of `memory` not being null and such, and `length` being greater than 0.

SNT/NativeMemoryManager.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System;
+
+namespace Torch.SNT
+{
+    /// <summary>
+    ///   Class used to wrap the native memory of a Torch Tensor.  
+    /// </summary>
+    internal class NativeMemory<T> : MemoryManager<T>
```
markusweimer: `sealed`?

SNT/NativeMemoryManager.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// <summary>
+        /// Returns a handle to the memory that has been pinned and hence its address can be taken.
+        /// </summary>
+        /// <param name="elementIndex">The offset to the element within the memory at which the returned <see cref="MemoryHandle"/> points to. (default = 0)</param>
+        public override MemoryHandle Pin(int elementIndex = 0)
+        {
+            unsafe
+            {
+                Retain();
+                if ((uint)elementIndex > length) throw new ArgumentOutOfRangeException(nameof(elementIndex));
```
markusweimer: Maybe add a more descriptive error message to the exception?

SNT/TypeGenerator.cs
```diff
@@ -0,0 +1,800 @@
...
+
+using System;
+using System.Linq;
+using System.Numerics.Tensors;
+using TorchSharp;
+
+namespace Torch.SNT 
+{
+
+    public sealed class ByteTorchTensor : DenseTensor<byte>
```
markusweimer: Please add xml docs to all `public` types, methods and fields.

SNT/TypeGenerator.cs
```diff
@@ -0,0 +1,800 @@
...
+    {
+        internal readonly object inner;
```
markusweimer: Why not `private`?

SNT/Utils.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+
+namespace Torch.SNT
+{
+    internal class Utils
+    {
+        public static int GetTotalLength(ReadOnlySpan<int> dimensions, int startIndex = 0)
```
markusweimer: Please add xml docs.

markusweimer: Also, couldn't this be a property on the Tensor object?

Test/FloatTorchTensorUnitTest.cs
```diff
@@ -0,0 +1,368 @@
...
+using System;
+using System.Linq;
+using System.Numerics.Tensors;
+using Torch.SNT;
+
+namespace Torch.SNT
+{
+    [TestClass]
+    public class FloatTorchTensorUnitTest
```
markusweimer: Should we auto-generate the unit tests for all types?

Test/Test.csproj
```diff
@@ -14,6 +14,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\TorchSharp\TorchSharp.csproj" />
+    <ProjectReference Include="..\SNT\SNT.csproj" />
```
markusweimer: Maybe we should have our own test project for the SNT wrapper? That way, we keep the two layers completely separate.

SNT/NativeMemoryManager.cs
interesaaat: Nope. memory get set to IntPtr.Zero when disposed.

SNT/NativeMemoryManager.cs
interesaaat: Length could be zero if you want to set up an empty tensors I presume. I can add the check that length is not negative and memory is not null.

SNT/NativeMemoryManager.cs
interesaaat: I redirected the constructor to the previous one. The conversion is ok, I checked online and it seems that it cannot fail, it is simply boxing a pointer into a managed object.

SNT/NativeMemoryManager.cs
interesaaat: Fixed

SNT/Utils.cs
interesaaat: There is a Length property in Tensor<T> (which I will map into a call to Torch nElement). But this is different because it is used before the tensor is created.

Test/FloatTorchTensorUnitTest.cs
interesaaat: Yes! Let me do it.

SNT/Utils.cs
interesaaat: Ops. My bad. Fixed.

Test/Test.csproj
interesaaat: I agree. At this point we should address #34 and create a unique test folder (either in source or on its own).

SNT/TypeGenerator.cs
interesaaat: Fixed.

interesaaat:
Fixed all comments so far.

SNT/TypeGenerator.tt
NiklasGustafsson: Aha! I didn't pay attention to the folder name.

SNT/SNT.csproj
NiklasGustafsson: So, looking at this some more -- TorchSharp.NNTensor.Data takes an IntPtr and casts it to a pointer, which makes it unsafe code. Then, your code takes the pointer and casts it back to an IntPtr for storage. If Tensor.Data instead returns the IntPtr it got from native code, you shouldn't have to handle pointers, right?

interesaaat: I can definitely remove some of the unsafe in that way. However I will still need Unsafe because the Pin method accepts as input an index and therefore I need to do pointer + index to return the correct slice of memory. Actually that Unsafe dependency is used for this.

NiklasGustafsson: Okay, that's good. We can get rid of the unsafe attribute on TorchSharp, then. I have a PR coming with some of those changes.

migueldeicaza:
Ouch - rerunning the tests still crashes.   I suspect that this indicates a real memory corruption bug on this PR.

interesaaat:
> Ouch - rerunning the tests still crashes. I suspect that this indicates a real memory corruption bug on this PR.

Yes there is something going on with the memory clean up. I have spent the last day in trying to figure this out with no luck. Please don't merge until I am able to fix the bug.

interesaaat:
I think I have addressed all comments and fixed the test error (thanks to @NiklasGustafsson).

NiklasGustafsson:
Should we still hold off on merging this one?

interesaaat:
> Should we still hold off on merging this one?

If you want we can wait until System.Numerics.Tensor NuGet get released. Beside this, it can be merged as is I think.

markusweimer:
> If this relies on a third party NuGet that is not released, let us wait until then.

The NuGet is available on MyGet. The .NET team is working on pushing it to Nuget.org soon. I'm OK holding off till that happens.

SNT/NativeMemoryManager.cs
ericstj: FWIW this isn't a public type.

SNT/NativeMemoryManager.cs
```diff
@@ -0,0 +1,133 @@
...
+            {
+                throw new InvalidOperationException("Disposing over unmatched Release/Retain");
+            }
+
+            IsDisposed = true;
+
+            if (disposing)
+            {
+                // Typically this would call into a native method appropriate for the platform
+                Marshal.FreeHGlobal(memory);
```
ericstj: I would expect this to PInvoke into torch to tell it to free the memory, or decrement some refcount if it happens to be shared.

SNT/TypeGenerator.tt
```diff
@@ -0,0 +1,160 @@
...
+    new { ApiName = "Long", Storage = "long", AccReal="long" },
+    new { ApiName = "Double", Storage = "double", AccReal="double" },
+    new { ApiName = "Float", Storage = "float", AccReal="double" },
+     }) {
+        string tname = type.ApiName;
+#>
+    /// <summary>
+    ///   Wrapper class used to surface a Torch <#=tname#>Tensor as a System.Numerics DensorTensor of <#=tname.ToLower()#>
+    /// </summary>
+    public sealed class <#=tname#>TorchTensor : DenseTensor<<#=tname.ToLower()#>>
```
ericstj: Can you summarize briefly why you needed to subclass DenseTensor rather than use it as is?  Was it only so that you could make the Clone give you back a tensor with memory owned by torch rather than managed memory?

SNT/NativeMemoryManager.cs
interesaaat: Thant's a good point. In practice the memory should never be disposed from `NativeMemory` (and in fact disposing is `false`).

ericstj: I guess that raises a different question then: which managed object is responsible for freeing this memory?

ericstj: Some managed object **must** maintain a refcount to the Torch tensor.  When that managed object is disposed or GC'ed it needs to tell Torch that it is done with that Tensor so that Torch knows it is safe to free it at that point.  What's the mechanism that handles this?

interesaaat: The refcount is directly managed by Torch, that's why I am forwarding all operations to the Torch layer. Like, in case of clone or resize, Torch maintain the refcount of the storage (memory) so that multiple tensors can point to the same memory and Torch take track of when is safe to free the memory.

ericstj: It looks to me like you're handling this with HType, so that HType is the thing with a finalizer responsible for telling Torch it is OK to free the memory.  In that case then I believe this type should be dealing in terms of HType's as SafeHandles instead of IntPtrs.  That way your NativeMemory objects will keep the SafeHandle's rooted.

interesaaat: You are referring to the memory pointer, right? HType and IntPtr are 2 different things here though. HType is a pointer to the tensor (storage) object in Torch, while IntPtr here is the actual pointer to the memory. HType is maintained by the tensor wrapper, not NativeMemory, which instead only surface the native memory. That's also why memory is not cleaned by NativeMemory, it will be freed by the tensor wrapper (by calling the proper free method on the Torch side). Does it make sense?

interesaaat:
I have update the dependency to the Nuget so this PR can now be merged (once @ericstj agrees)

SNT/NativeMemoryManager.cs
```diff
@@ -0,0 +1,133 @@
...
+            {
+                throw new InvalidOperationException ("Disposing over unmatched Release/Retain");
+            }
+
+            IsDisposed = true;
+
+            if (disposing)
+            {
+                // Typically this would call into a native method appropriate for the platform
+                Marshal.FreeHGlobal (memory);
```
ericstj: You must remove this, regardless of our other discussion, you're memory is not allocated on the global heap.

SNT/TypeGenerator.tt
ericstj: Thank you for this.  Did you consider making your derived type generic as well?

interesaaat: I think my initial implementation was generic but we figured that a non-generic version was simpler. What do you think? Do you think that a generic type will be more "compatible" with the `Tensor<T>` design?

SNT/NativeMemoryManager.cs
interesaaat: Make sense. Will fix this.

SNT/NativeMemoryManager.cs
interesaaat: I see what you mean. The workflow is like this:
1) create a tensor
2) retrieve the memory of the tensor by the Buffer property
3) the tensor is destroyed
4) try to do some operation on the memory.
Let me write a test case for this and if this is the case I will write a fix.

SNT/TypeGenerator.tt
ericstj: If you have a generic type / methods it enables generic callers, if you don't have them then that scenario is not possible (unless the caller creates a type-switch).  Ideally we want to handle the type switching in the libraries rather than force all the callers to do type switching.

We also try to design generic APIs so that the generic type can be inferred: this is why we have the static creation methods on Tensor (note: not Tensor<T>) rather than forcing folks to call the constructor.

interesaaat: In our case we force users to create tensors by calling the proper static object (e.g., `FloatTorchTensor.Create`). The main problem we had with the generic approach is that since the inner torch tensors are not generic, each function call would have to contain a switch statement. I think until with don't have a proper generic type hierarchy for the inner torch tensor type (issue #18) having a not generic type is the only reasonable possibility. But I agree that we should try to have a generic API.

NiklasGustafsson: The underlying C APIs are element-type-specific, so without something like C++'s template specialization available to us, how would we implement a generic concrete Tensor class?

interesaaat: Like here #28.

ericstj: > The main problem we had with the generic approach is that since the inner torch tensors are not generic, each function call would have to contain a switch statement.

Tensor<T> handles this specialization already in other cases.  Consider the arithmetic functions where we needed a bridge from generic type to type-specific IL in all function calls.  We have a mapping that happens per-managed type and gets stored in a singleton: https://github.com/dotnet/corefx/blob/master/src/System.Numerics.Tensors/tests/TensorArithmetic.cs#L44.  This isn't the most elegant solution but it works and is relatively lightweight.  I believe @tannergooding also had some ideas to improve this.

motus: FWIW, libtorch C++ API *is not* generic - that is, it has a single non-parameterized `at::Tensor` class with methods like `.type()`, `.device()`, `.is_sparse()` etc. All dispatching to the proper backend and function specialization happens at runtime in `Type` and `TensorImpl` classes. So maybe if even libtorch developers were not able to make the `Tensor` class parameterized - not even by the scalar type and in less restrictive C++ template system, we should not worry about it too much, either?

P.S. That said, I'd be delighted to see a strongly typed tensor library for C# - ideally something in the spirit of Haskell's [HMatrix](http://dis.um.es/~alberto/hmatrix/hmatrix.html). The question is how much effort we want to put into it. (Maybe if we have a nice low-level API now, we can build an alternative strongly-typed parameterized API on top of it later?) Just thinking out loud.

interesaaat: These are two different concerns tho. I think that the `Tensor<T>` wrapper for TorchSharp should be following the `Tensor<T>` paradigm, i.e., try to use generics. For the main TorchSharp story, I totally agree with you: it is fine to be non-generic and consistent with the Torch API, even if a generic API will be better (and Issue #18 is exactly for this).

SNT/NativeMemoryManager.cs
interesaaat: I fixed this. Now `Native Memory` points to the Torch storage instead of directly to the data pointer. Torch manages the ref count etc.

SNT/TypeGenerator.tt
interesaaat: I was looking into following the approach of System.Numerics.Tensors for dispatching types but it is not going to work in our case at the moment. The problem is that we don't have a root type for the tensors in TorchSharp (i.e., the `ITensorArithmetic<T>` in https://github.com/dotnet/corefx/blob/master/src/System.Numerics.Tensors/tests/TensorArithmetic.cs) therefore I am not able to create a singleton object for type dispatching.

SNT/NativeMemoryManager.cs
ericstj: I see.  I'm taking a closer look at these storage classes and they are actually very close to NativeMemory in terms of functionality.  I wonder if you could collapse the two and have the storage classes derive from MemoryManager<T>?

SNT/TypeGenerator.tt
ericstj: That's just an internal interface, could't you create one?

ericstj:
I'm concerned with the overall approach that you now have two tensor types exposed instead of just one.  This PR creates Torch.SNT.*TorchTensor, but you still have TorchSharp.*Tensor.  Is this PR a step along the way, or do you imagine having both of these tensor types remain publicly exposed?

interesaaat:
> I'm concerned with the overall approach that you now have two tensor types exposed instead of just one. This PR creates Torch.SNT.*TorchTensor, but you still have TorchSharp.*Tensor. Is this PR a step along the way, or do you imagine having both of these tensor types remain publicly exposed?

I think that the idea was to have TorchSharp as a thin layer with minimal dependencies to provide a way to access Torch functionalities from C#. Torch.SNT should instead try to back System.Numerics.Tensors with Torch tensors so that, from one side, we can exploit the efficient implementation of tensor operations that Torch provide, and from the other we can converge to "the" tensor type in .Net without providing a new one (TorchSharp.Tensor). I agree that in this way some of the functionalities will have duplicates (e.g., NativeMemory and Storage and 2 tensor types) but @migueldeicaza was clear that he wants to separate the 2 efforts (am I wrong?).

SNT/TypeGenerator.tt
```diff
@@ -0,0 +1,257 @@
...
+            }
+
+            /// <summary>
+            /// Destructor for the native memory.
+            /// The Tensor memory lifecycle is managed by Torch. Disposing the memory
+            /// from here will through a "double free or corruption" error.
+            /// <summary>
+            ~<#=tname#>NativeMemory ()
+            {
+                Dispose (true);
```
ericstj: Shouldn't this be false?

interesaaat: In this version I actually have to dispose the memory. If you look at line 146, I am retain the storage pointer every time a new native memory is created (or pinned). Therefore I need to `Dispose (true)` to properly decrement the reference count. I have to fix the comment :)

ericstj: I don't believe you can safely do that.  The GC does not guarantee that `storage` (another managed object) has not yet been finalized at the time the finalizer for this NativeMemory runs.  Only one object should be in charge of maintaining the native state and that object should be the one that has the finalizer that does work.  (ps, consider https://github.com/xamarin/TorchSharp/pull/51#discussion_r234733245)

interesaaat: There is something I am missing on C# GC then. Native storage maintains a reference to `storage`, why the GC should destroy `storage` if refcount > 0?

ericstj: Both references are out of scope at the same time.  This is [documented](https://docs.microsoft.com/en-us/dotnet/api/system.object.finalize) behavior

>The finalizers of two objects are not guaranteed to run in any specific order, even if one object refers to the other. That is, if Object A has a reference to Object B and both have finalizers, Object B might have already been finalized when the finalizer of Object A starts.

I suspect this has to do with GC performance and how it calculates live objects in a single pass and puts them on the finalization queue in the order they are encountered (not necessarily in the order they appear in an object graph), also likely has to do with not needing to deal with potentially deep object cycles.  @Maoni0 could explain better.

ericstj: That sounds right.  Of course in the case that no-one calls Dispose, the finalizer for storage will still clean up the native resources, therefore you can technically remove the finalizer from NativeMemory since it does nothing.  You can also remove the public Dispose method unless you need to explicitly call it, since the base type implements IDisposable.

Did you consider unifying the storage and NativeMemory classes?  I think that actually might reduce the number of object allocations per tensor and feels like the right conceptual pairing.

interesaaat: > Did you consider unifying the storage and NativeMemory classes? I think that actually might reduce the number of object allocations per tensor and feels like the right conceptual pairing.

Yes I was looking into that but this will require to add a dependency to the main TorchSharp project which I believe @migueldeicaza and @markusweimer want to avoid.

ericstj: >  this will require to add a dependency to the main TorchSharp project

You're talking about a dependency on System.Memory?  Seems like that's not such a bad depdendency to have since it would let TorchSharp use Span, but I can see why you may want to avoid it.

I think this is leading to confusion in how this NativeMemory class is implemented.  I was taking a closer look and now I see that Dispose is calling Free, which but I don't think this is correct.  It looks to me like you're using Free to represent "decrement a reference count" however the reference count may not have been previously incremented since I don't see a case where this is happening during construction.  Since the NativeMemory itself isn't the owner of the memory at all, I wonder if it even makes sense to do anything in its Dispose 😕: in fact I don't even see where you call this.

In the Tensor/NativeMemory sample we were using the NativeMemory object as the "owner" of the memory.  As such, it was the one implementing the reference count and it was the one responsible for eventually freeing that memory in its finalizer/dispose.

In this case, you're using NativeMemory as an adapter for your storage class which is ultimately the owner of the memory.  The adapter acts as the mechanism for constructing spans over the storage.

Furthermore you have added IDisposable to your derived Tensors and it looks to me like this will dispose the backing torch tensor which will dispose the backing storage.  I would have imagined you'd instead want to have decremented a reference count here and dispose the backing storage when the ref count gets to zero.  As its implemented I think you'll end up freeing the memory behind a shallow-copied reshaped tensor.

Perhaps you can add some test cases to further flesh out how this should behave?   

@tannergooding perhaps we could expand our Tensor samples to cover this scenario?

danmosemsft:
@tannergooding owns S.N.T and can help make sure that the right integration is happening here. It's important to us that S.N.T is valuable for TorchSharp.

SNT/TypeGenerator.tt
interesaaat: > It looks to me like you're using Free to represent "decrement a reference count" however the reference count may not have been previously incremented since I don't see a case where this is happening during construction.

I have pushed the call to Retain into the Nativ Memory constructor. Before it was outside when the storage was retrieved from the tensor to create the NativeMemory object.

> Since the NativeMemory itself isn't the owner of the memory at all, I wonder if it even makes sense to do anything in its Dispose 😕: in fact I don't even see where you call this.

Dispose is required because NameMemory extends MemoryManager. 

> Furthermore you have added IDisposable to your derived Tensors and it looks to me like this will dispose the backing torch tensor which will dispose the backing storage. I would have imagined you'd instead want to have decremented a reference count here and dispose the backing storage when the ref count gets to zero. As its implemented I think you'll end up freeing the memory behind a shallow-copied reshaped tensor.

This is exactly the behavior I implemented (I hope :) ). The fact is that NativeMemory is required to dispose the storage because it may happen the case where a span is retrieved and used outside a tensor context (e.g., in [this](https://github.com/xamarin/TorchSharp/pull/51/files#diff-9d09654dea42c82f2b91e9afc1c739a1R1881) test case).

SNT/TypeGenerator.cs
```diff
@@ -0,0 +1,1264 @@
...
+
+            /// <summary>
+            /// Returns a handle to the memory that has been pinned and hence its address can be taken.
+            /// </summary>
+            /// <param name="elementIndex">The offset to the element within the memory at which the returned <see cref="MemoryHandle"/> points to. (default = 0)</param>
+            public override MemoryHandle Pin (int elementIndex = 0)
+            {
+                unsafe
+                {
+                    storage.Retain();
```
eerhardt: Should you call `Retain()` after doing the argument check? I wouldn't expect needing to call `Unpin` if `Pin` threw an exception.

SNT/TypeGenerator.cs
```diff
@@ -0,0 +1,1264 @@
...
+            protected override void Dispose (bool disposing)
+            {
+                if (disposing)
+                {
+                    storage.Free ();
+                }
+            }
+        }
+
+        private readonly object inner;
```
eerhardt: Why is this field held as `object`? why not `ByteTensor`?

SNT/TypeGenerator.cs
interesaaat: Yes. Fixed.

SNT/TypeGenerator.cs
interesaaat: This is something coming from the previous version using generics. Fixed.

Issue #21, opened 2018-10-31
Binding: TH.*Blas APIs


migueldeicaza:
* [x] Create `DllImport` bindings for the following functions:
   - [x] THByteBlas_axpy
   - [x] THByteBlas_copy
   - [x] THByteBlas_dot
   - [x] THByteBlas_gemm
   - [x] THByteBlas_gemv
   - [x] THByteBlas_ger
   - [x] THByteBlas_scal
   - [x] THByteBlas_swap
   -
   - [ ] THCharBlas_axpy
   - [ ] THCharBlas_copy
   - [ ] THCharBlas_dot
   - [ ] THCharBlas_gemm
   - [ ] THCharBlas_gemv
   - [ ] THCharBlas_ger
   - [ ] THCharBlas_scal
   - [ ] THCharBlas_swap
   -
   - [x] THDoubleBlas_axpy
   - [x] THDoubleBlas_copy
   - [x] THDoubleBlas_dot
   - [x] THDoubleBlas_gemm
   - [x] THDoubleBlas_gemv
   - [x] THDoubleBlas_ger
   - [x] THDoubleBlas_scal
   - [x] THDoubleBlas_swap
   -
   - [x] THFloatBlas_axpy
   - [x] THFloatBlas_copy
   - [x] THFloatBlas_dot
   - [x] THFloatBlas_gemm
   - [x] THFloatBlas_gemv
   - [x] THFloatBlas_ger
   - [x] THFloatBlas_scal
   - [x] THFloatBlas_swap
   -
   - [x] THIntBlas_axpy
   - [x] THIntBlas_copy
   - [x] THIntBlas_dot
   - [x] THIntBlas_gemm
   - [x] THIntBlas_gemv
   - [x] THIntBlas_ger
   - [x] THIntBlas_scal
   - [x] THIntBlas_swap
   -
   - [x] THLongBlas_axpy
   - [x] THLongBlas_copy
   - [x] THLongBlas_dot
   - [x] THLongBlas_gemm
   - [x] THLongBlas_gemv
   - [x] THLongBlas_ger
   - [x] THLongBlas_scal
   - [x] THLongBlas_swap
   -
   - [x] THShortBlas_axpy
   - [x] THShortBlas_copy
   - [x] THShortBlas_dot
   - [x] THShortBlas_gemm
   - [x] THShortBlas_gemv
   - [x] THShortBlas_ger
   - [x] THShortBlas_scal
   - [x] THShortBlas_swap
[...]

PR #57, opened 2018-11-07
Create bindings for raw TH*Blas_* functions

motus:
This PR addresses the first part of issue #21 
I will implement high-level functions for BLAS operations like dot product of matrix-vector multiplication on the next PR.


TorchSharp/Raw/Blas.tt
```diff
@@ -0,0 +1,188 @@
...
+        // *** Level 1
+
+        /// <summary>
+        /// <a href="http://www.netlib.org/blas/#_blas_routines">BLAS</a> <c>swap</c> function for <#=tName#> type.
+        /// Swap x and y.
+        /// </summary>
+        /// Corresponds to the following TH definition:
+        /// <code>
+        /// TH_API void THBlas_(swap)(
+        ///     int64_t n, scalar_t *x, int64_t incx, scalar_t *y, int64_t incy);
```
interesaaat: This is great! We should use this as standard for the documentation!

migueldeicaza: The above documentation is not correct, and produces a warning, this should be fixed.

I do not think we want to add to the documentation a link to the C API, not sure what value it adds - it is not like the C API has any documentation that could be useful for people anyways.

Our documentation should cover what the API does, so I rather spend the effort on digging through the documentation for how these operations are surfaced in Python and put that documentation there.

motus: @migueldeicaza I put those references mostly for the sanity check - if function has a name like `gemm` and 13 parameters, all equally cryptic and with minimal type checks, it is important to make sure that nothing gets lost in translation. Anyway, I have no problem with removing that stuff. Is it OK if I just keep one-line summary for each function? We can add more docs in subsequent PRs

motus:
OK I've removed the BLAS link and moved the C declaration to the `<remarks>` section. @migueldeicaza , what do you think?

**P.S.** I have no idea why the CI build fails - looks like a server problem(?)

TorchSharp/Raw/Blas.tt
migueldeicaza: We can keep it with regular "//" comments, let us not put it there for the comments that go into the API docs (the "///").

As they currently stand, they also do not compile.

motus:
@migueldeicaza oh I got it. will send a fix in a minute

TorchSharp/Raw/Blas.tt
```diff
@@ -0,0 +1,158 @@
...
+namespace TorchSharp {
+<#
+foreach (var type in TorchTypeDef.Types) {
+#>
+
+    public partial class <#=type.Name#>Tensor : IDisposable {
+
+        // *** Level 1
+
+        // Swap x and y.
```
interesaaat: I think this should go into summary, no?

interesaaat: Unless you put them in // so that you can add the summary into the actual C# API. Is this your plan?

motus: @interesaaat I plan to have a short description like this one both here and in xmldoc for high-level API. That is, when you look at the code, you have a rough idea what the function is for, yet we don't clutter the generated docs with low-level  API that we don't expose anyway.

TorchSharp/Raw/Blas.tt
motus: @migueldeicaza OK, I've switched to regular `//` comments and removed all XML tags. Both Debug and Release builds worked for me until I've merged in the master. Current master d1bb43fd42e25d222f2eb01f0c327ab1feb7cee1 fails to build the Release, and also has some xmldoc issues:
```
DiskFile.cs(9,18): warning CS1591: Missing XML comment for publicly visible type or member 'DiskFile' [/mnt/c/Users/sergiym/devel/TorchSharp/TorchSharp/TorchSharp.csproj]
THMemoryFile.cs(148,26): warning CS1572: XML comment has a param tag for 'handle', but there is no parameter by that name [/mnt/c/Users/sergiym/devel/TorchSharp/TorchSharp/TorchSharp.csproj]
THMemoryFile.cs(150,39): warning CS1573: Parameter 'storage' has no matching param tag in the XML comment for 'MemoryFile.MemoryFile(MemoryFile.CharStorage, string)' (but other parameters do) [/mnt/c/Users/sergiym/devel/TorchSharp/TorchSharp/TorchSharp.csproj]
/usr/share/dotnet/sdk/2.1.403/Sdks/Microsoft.NET.Sdk/targets/Microsoft.PackageDependencyResolution.targets(198,5): error NETSDK1004: Assets file '/mnt/c/Users/sergiym/devel/TorchSharp/TorchSharp/C:/Users/sergiym/devel/TorchSharp/TorchSharp/obj/project.assets.json' not found. Run a NuGet package restore to generate this file. [/mnt/c/Users/sergiym/devel/TorchSharp/TorchSharp/TorchSharp.csproj]
    3 Warning(s)
    1 Error(s)
```
so it probably isn't related to this PR

motus:
@migueldeicaza Does it still produce xmldoc errors on your machine? could you please check and let me know (and maybe copy the message that you get?)
**P.S.** I've also moved all functions into internal static class, as per @NiklasGustafsson suggestion.

migueldeicaza:
I have seen issues like the one described about project.json before in other projects, usually I do a clean, and an `msbuild /t:restore` to get back to a working state.   You can try that.

On the documentation issue beyond your code, I will fix those manually.

PR #58, opened 2018-11-08
File APIs and minor changes to NNNTensor.

NiklasGustafsson:
Added File APIs to write entire storages, arrays, and tensors.
Changed how Tensor.Storage was used.
Changed Tensor.Data to return IntPtr.
Changed Tensor.NumElements to read-only property.

Added unit tests.


NiklasGustafsson:
This branch mine started showing up with errors. I'll pull back and submit a partial PR.

Issue #22, opened 2018-10-31
Binding: TH.*Vector APIs


migueldeicaza:
- [x] THByteVector_adds
- [x] THByteVector_cadd
- [x] THByteVector_cdiv
- [x] THByteVector_cmul
- [x] THByteVector_copy
- [x] THByteVector_divs
- [x] THByteVector_fill
- [x] THByteVector_muls
- [x] THByteVector_neg
- [x] THByteVector_normal_fill
- [ ] THCharVector_adds
- [ ] THCharVector_cadd
- [ ] THCharVector_cdiv
- [ ] THCharVector_cmul
- [ ] THCharVector_copy
- [ ] THCharVector_divs
- [ ] THCharVector_fill
- [ ] THCharVector_muls
- [ ] THCharVector_neg
- [ ] THCharVector_normal_fill
- [x] THDoubleVector_abs
- [x] THDoubleVector_acos
- [x] THDoubleVector_adds
- [ ] THDoubleVector_adds_AVX
- [x] THDoubleVector_asin
- [x] THDoubleVector_atan
- [x] THDoubleVector_cadd
- [ ] THDoubleVector_cadd_AVX
- [x] THDoubleVector_cdiv
- [ ] THDoubleVector_cdiv_AVX
- [x] THDoubleVector_ceil
- [x] THDoubleVector_cinv
- [x] THDoubleVector_cmul
- [ ] THDoubleVector_cmul_AVX
- [x] THDoubleVector_copy
- [ ] THDoubleVector_copy_AVX
- [x] THDoubleVector_cos
- [x] THDoubleVector_cosh
- [x] THDoubleVector_digamma
- [x] THDoubleVector_divs
- [ ] THDoubleVector_divs_AVX
- [x] THDoubleVector_erf
- [x] THDoubleVector_erfc
- [x] THDoubleVector_erfinv
- [x] THDoubleVector_exp
- [x] THDoubleVector_expm1
- [x] THDoubleVector_fill
- [ ] THDoubleVector_fill_AVX
- [x] THDoubleVector_floor
- [x] THDoubleVector_frac
- [x] THDoubleVector_lgamma
- [x] THDoubleVector_log
- [x] THDoubleVector_log10
- [x] THDoubleVector_log1p
- [x] THDoubleVector_log2
- [x] THDoubleVector_muls
- [ ] THDoubleVector_muls_AVX
- [x] THDoubleVector_neg
- [x] THDoubleVector_normal_fill
[...]

PR #62, opened 2018-11-09
[issue #22] Create binding for TH*Vector_* functions

motus:
This PR addresses issue #22


TorchSharp/Raw/Vector.tt
```diff
@@ -0,0 +1,153 @@
...
+using System.Runtime.InteropServices;
+
+namespace TorchSharp.Raw {
+
+    internal static class Vector {
+
+<#
+foreach (var type in TorchTypeDef.Types) {
+#>
+        // Fill n elements of vector x with the value c.
```
interesaaat: What do you mean with fill n elements? Are these the first n elements, or skip the first n elements? (I see that n is ptrdiff)

TorchSharp/Raw/Vector.tt
```diff
@@ -0,0 +1,153 @@
...
+        //
+        // Corresponds to the following TH declaration:
+        //
+        // TH_API void THVector_(fill)(scalar_t *x, const scalar_t c, const ptrdiff_t n);
+        [DllImport ("caffe2")]
+        internal extern static void TH<#=type.Name#>Vector_fill(
+            <#=type.Ptr#> x,
```
interesaaat: Can we have the parameters in one line? I see that this is what you are doing for the other functions.

TorchSharp/Raw/Vector.tt
motus: `ptrdiff_t` is used like `size_t` here and in other functions, and `x` can be an offset into the vector. That is, the function will assign `c` to `n` values starting from pointer `x`. I will update the comment to clarify.

interesaaat: ah! I see. For some reason I was assuming that `ptrdiff_t` was the index in `x`.

interesaaat:
One question, instead of having Raw with the tt files for BLAS, etc. should we be consistent with the project and having tt files and cs file named something like THBLAS and leaving them into the main directory?

motus:
@interesaaat I actually don't like cluttering the main directory with files related to the low-level API. I would rather move the `extern TH*Tensor_*` functions into their own file and namespace in `Raw/`. I also think that having `TH*` prefix for high-level classes and modules is redundant - it does not not carry any additional information.  I am OK with keeping `TH*` prefix for low-level functions, though, so that the C# names match their C counterparts exactly.

interesaaat:
Sounds good to me. Should we fill an issue to re-organize the project structure?

PR #63, opened 2018-11-09
Read and write tensors

NiklasGustafsson:
This PR adds support for reading and writing entire tensors.
Also, adding more ways to create tensors with explicitly specified storage.


TorchSharp/TypeGeneration.tt
```diff
@@ -541,21 +539,51 @@ foreach (var type in new [] {
...
+        ///   Create a new tensor with a specified storage arena.
+        /// </summary>        
+        /// <param name="storage">Storage arena.</param>
+        /// <param name="offset">Offset within the input storage the storage of the new tensor will start from.</param> 
+        /// <param name="size">Size of the first dimension.</param>     
+        /// <param name="stride">Stride of the first dimension.</param>          
+        public static <#=tname#>Tensor NewWithStorage1d(<#=tname#>Tensor.<#=tname#>Storage storage, UIntPtr offset, long size, long stride)
+        {
+            var result = new <#=tname#>Tensor(TH<#=tname#>Tensor_newWithStorage1d(storage.handle, offset, size, stride));
+            result.storage = storage;
```
interesaaat: Not sure about this. Can you please make sure that `result.storage.handle == storage.handle`. Because if they are different (which I think they are) we are making some assumptions here that may create bugs.

interesaaat: This comment applies for the following functions too.

interesaaat: I see what you mean but I don't understand the sentence "having managed state means that in the future, when APIs surface a type, we have no way of retrieving "attached" data (like this field in Storage)". As it is now storage contains no data, it is just a memoization for the call to `TH<#=tname#>Tensor_storage (handle)`. I added this because we were creating (what I thought were) unnecessary objects.

migueldeicaza: This is a very common scenario when doing native bindings like this, and while it is solvable, it introduces a tremendous amount of complexity of dealing with correctly.   It happens in almost every binding where you have pure C# objects that are proxies to unmanaged objects.

The bug happens like this.

You create a managed object that is a peer to a native object, with additional data, then you pass the instance of this to some unmanaged object, for example "fooBar.SetTensor (t)", now you let "t" go out of scope and the reference managed by managed code drops and the object is collected.   Now, at some point in the future, the type is resurfaced, for example "fooBar.GetTensor()" call.   

When you have a pure peer, it is cheap to create those objects, you merely create a new wrapper type by setting the handle, and you are off to the races.   But if you had state, the state is now gone.

Now, it might be possible that we can always recreate this particular one, but if this was out of sync, then we would have a situation where those do not match.

To solve this problem, you now need to keep a GCHandle to the object, to ensure that as long as the unmanaged reference exists, the managed object is kept alive.   This complicates things, because now only unmanaged code can properly release the data, as the GC would not be able to do its job (as far as the GC is concerned, the managed object is retained by the GCHandle).

So now we need to deal with some sort of notification system for when the unmanaged object is released.   Some frameworks have it, some do not.    This one does not.

Now, it is possible that this is not a problem for storage, but I would rather not have any managed state in the pure peers that forces us to come up to a solution for the above problem.

Btw, the above points out that we probably should seal these types as well.

TorchSharp/FileOperations.cs
```diff
@@ -116,6 +114,69 @@ public long WriteTensor(TorchSharp.ByteTensor tensor)
...
+
+        /// <summary>
+        ///   Write bytes to the file from the given byte array.
+        /// </summary>
+        /// <param name="data">An array containing data to be written to the file.</param>
+        /// <param name="n">The number of bytes to write. Pass -1 (default) to write the whole array.</param>
+        /// <returns>The number of bytes written.</returns>
+        public long WriteChars(byte[] data, int n = -1)
+        {
+            n = (n == -1) ? data.Length : Math.Min(n, data.Length);
```
interesaaat: Should we have a check that n is not < -1?

TorchSharp/FileOperations.cs
```diff
@@ -625,5 +676,25 @@ public long WriteTensor(TorchSharp.DoubleTensor tensor)
...
         }
+
+        [DllImport("caffe2")] 
+        extern static long THFile_readChar(HType self, MemoryFile.CharStorage.HType storage);
+
+        /// <summary>
+        ///   Read chars from the file into the given storage.
+        /// </summary>
+        /// <param name="storage">A storage object to read data into.</param>
+        /// <returns>The number of shorts read.</returns>
```
interesaaat: bytes?

TorchSharp/FileOperations.cs
```diff
@@ -625,5 +676,25 @@ public long WriteTensor(TorchSharp.DoubleTensor tensor)
...
+        public long ReadChars(MemoryFile.CharStorage storage) { return THFile_readChar(this.handle, storage.handle); }
+
+        [DllImport("caffe2")] 
+        extern static long THFile_writeChar(HType self, MemoryFile.CharStorage.HType storage);
+
+        /// <summary>
+        ///   Write chars to the file from the given storage.
+        /// </summary>
+        /// <param name="storage">A storage object fetch data from.</param>
+        /// <returns>The number of shorts written.</returns>
```
interesaaat: bytes?

TorchSharp/FileOperations.cs
```diff
@@ -116,6 +114,69 @@ public long WriteTensor(TorchSharp.ByteTensor tensor)
...
+        
+        /// <summary>
+        ///   Read bytes from the file into the given byte array.
+        /// </summary>
+        /// <param name="data">An array to place the data in after reading it from the file.</param>
+        /// <param name="n">The maximum number of bytes to read.</param>
+        /// <returns>The number of bytes read.</returns>
+        public long ReadChars(byte[] data, int n)
+        {
+            if (n > data.Length)
```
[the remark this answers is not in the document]
NiklasGustafsson: Good idea. I added it.

TorchSharp/FileOperations.cs
NiklasGustafsson: Good idea. I added it.

TorchSharp/FileOperations.cs
NiklasGustafsson: Fixed

TorchSharp/TypeGeneration.tt
```diff
@@ -447,9 +448,19 @@ foreach (var type in new [] {
 
         /// <summary>
         ///  Returns the associated storage for this tensor
-        /// </summary>
-        public <#=tname#>Storage Storage => new <#=tname#>Storage (TH<#=tname#>Tensor_storage (handle));
-
+        /// </summary>
```
interesaaat: I believe [we wanted](https://github.com/xamarin/TorchSharp/pull/63#discussion_r233228263) to remove `internal storage`. Please delete the modifications to storage. Probably they will be fixed if you fetch and rebase with master.

TorchSharp/TypeGeneration.tt
NiklasGustafsson: I have removed the memoization logic.

NiklasGustafsson:
All comments have been taken into account at this point.

PR #65, opened 2018-11-13
Removed storage internal managed state

interesaaat:


migueldeicaza:
Thank you for the fix!    I manually merged it due to the conflict in the PR.  THanks!

PR #93, opened 2019-06-26
Added few missing functionalities in NN and tensors.

interesaaat:
Also testing the pipelines


src/TorchSharp/NN/Linear.cs
```diff
@@ -10,47 +10,43 @@ public Linear(IntPtr handle) : base(handle)
         {
         }
 
-        [DllImport("LibTorchSharp")]
-        private static extern IntPtr THSNN_linearModule(long input_size, long output_size, bool with_bias);
+        [DllImport("libTorchSharp")]
```
artidoro: The errors you get for Linux are probably due to the lower case of libTorchSharp.
Linux is case sensitive, and I think I used LibTorchSharp everywhere else in the code.

Let me know if that does not work.

joemoorhouse:
Oh great; was going to ask if someone was working on scalar-tensor ops! Is idea to have implicit conversions to Scalar or add overloads? e.g. public static implicit operator Scalar(double value){…}

src/TorchSharp/NN/Linear.cs
interesaaat: Ah-ha! Let me fix this

interesaaat:
@joemoorhouse for the moment we you need to explicitly convert into scalars. For example we do 
```cs
 weights.Eq(1.ToScalar()).Sum(new[] { 1L }) * (-1).ToScalar() + 0.5.ToScalar();
```
which is terrible! Please open an issue and feel free to fix this if you want to!

interesaaat:
Tests pass now (thanks @artidoro). I am going to merge this.

Issue #100, opened 2019-07-11
Check "inPlace" methods


interesaaat:
Apparently they are not strictly in place and instead we need to use the returning tensor.

PR #105, opened 2019-07-18
Add C Sharp numeric type scalar ops

joemoorhouse:
Added various arithmetic and comparison operators to test. Suggested name for static Math class.


joemoorhouse:
Need to correct merge error

joemoorhouse:
Notice that I tabified a couple of C++ files which makes diff inconvenient! Will undo that (I guess convention is to use spaces in C++ code?)

interesaaat:
> Notice that I tabified a couple of C++ files which makes diff inconvenient! Will undo that (I guess convention is to use spaces in C++ code?)

Yes please! You can also change the IDE settings from tabs to 4 spaces.

src/TorchSharp/Scalar.cs
```diff
@@ -12,6 +12,11 @@ internal Scalar(IntPtr handle)
             Handle = handle;
         }
 
+        public static implicit operator Scalar(double value)
```
interesaaat: Can we add this also for other types?

src/Native/LibTorchSharp/THSTensor.h
```diff
@@ -219,25 +229,25 @@ EXPORT_API(Tensor) THSTensor_transpose(const Tensor twrapper, const int64_t dim1
 EXPORT_API(void) THSTensor_transpose_(const Tensor twrapper, const int64_t dim1, const int64_t dim2);
```
interesaaat: This should return `Tensor` (so that we can also close #100).

src/TorchSharp/Tensor/TorchTensor.cs
```diff
@@ -551,13 +509,29 @@ public TorchTensor Eq(TorchTensor target)
         }
 
         [DllImport("LibTorchSharp")]
-        extern static IntPtr THSTensor_eqS(IntPtr src, IntPtr trg);
+        private static extern void THSTensor_eq_(IntPtr src, IntPtr trg);
```
interesaaat: This should return IntPtr and the method create a new tensor.

src/TorchSharp/Tensor/TorchTensor.cs
```diff
@@ -527,19 +477,27 @@ public void DivInPlace(TorchTensor target)
...
-            return new TorchTensor(THSTensor_gtS(handle, target.Handle));
+            return new TorchTensor(THSTensor_erf(handle));
+        }
+
+        [DllImport("LibTorchSharp")]
+        private static extern IntPtr THSTensor_erf_(IntPtr src);
+
+        public void ErfInPlace()
+        {
+            THSTensor_erf_(handle);
```
interesaaat: Should we return a new tensor here?

src/TorchSharp/Tensor/TorchTensor.cs
```diff
@@ -551,13 +509,29 @@ public TorchTensor Eq(TorchTensor target)
...
+        [DllImport("LibTorchSharp")]
+        private static extern IntPtr THSTensor_eqS(IntPtr src, IntPtr trg);
 
         public TorchTensor Eq(Scalar target)
         {
             return new TorchTensor(THSTensor_eqS(handle, target.Handle));
         }
 
+        [DllImport("LibTorchSharp")]
+        private static extern void THSTensor_eqS_(IntPtr src, IntPtr trg);
```
interesaaat: IntPtr here

src/TorchSharp/Tensor/TorchTensor.cs
```diff
@@ -551,13 +509,29 @@ public TorchTensor Eq(TorchTensor target)
...
+
+        public void EqInPlace(Scalar target)
+        {
+            THSTensor_eqS_(handle, target.Handle);
```
interesaaat: new tensor here

src/TorchSharp/Tensor/TorchTensor.cs
```diff
@@ -575,13 +549,149 @@ public TorchTensor Exp()
...
-        extern static IntPtr THSTensor_geS(IntPtr src, IntPtr trg);
+        private static extern IntPtr THSTensor_ge(IntPtr src, IntPtr trg);
+
+        public TorchTensor Ge(TorchTensor target)
+        {
+            return new TorchTensor(THSTensor_ge(handle, target.Handle));
+        }
+
+        [DllImport("LibTorchSharp")]
+        private static extern void THSTensor_ge_(IntPtr src, IntPtr trg);
```
interesaaat: here as well (and for all the following "in place" methods)

src/TorchSharp/Tensor/TMath.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
```
interesaaat: Do we actually need this class?

joemoorhouse: Yes, wanted to ask about this one! In libtorch we can do things like:
return 0.5 * (1 + torch::exp(value / sqrt(2))); // libtorch
Was wondering if we need static Math-like class to do analogous thing in TorchSharp, e.g. 
return 0.5 * (1 + TMath.Exp(value / Math.Sqrt(2))); // TorchSharp?
Can add to TorchTenor, but perhaps becomes a little verbose and in C# we might expect a class with a name like 'Math' to be used? Please let me know what you prefer...

interesaaat: Ah that a good point. I believe that what we can do is detaching the current C# API into a functional (basically containing the static methods calling into C) and an object oriented API (the current C# API). In this way one can basically pick which API to use. Some of the methods of course will only live in one of the API (like exp only on the functional one). WDYT?

markusweimer: The lower level, all `static` function API would also be useful for a more idiomatic F# API, right?

interesaaat: Exactly.

joemoorhouse:
I think this fixes the in-place operations now.

interesaaat:
For the moment I am going to delete the TMath class, merge the PR, and open a new issue on the functional API.

interesaaat:
Closes #105 and #100.

src/TorchSharp/Tensor/TMath.cs
interesaaat: #110

PR #106, opened 2019-07-19
Fix libtorch packaging

artidoro:
This fixes issues that we were seeing with the Libtorch.Redist nuget.
With this PR LibTorchSharp will be included in the Libtorch.Redist nuget. I noticed that we were getting DLL errors if the native components were split into two nugets. This change will put all the native components in the same nuget and folder within that nuget.


interesaaat:
I will try it today. However I see that the CI returns an error. Can you please look at the `DownloadBuildArtifacts@0` in the azure pipeline yaml file and see whether you can pick single packages instead of downloading everything or if it will silently fail if a package is already in the dir? (It looks that is failing because we are downloading the Redist from all builds).

interesaaat:
> I will try it today. However I see that the CI returns an error. Can you please look at the `DownloadBuildArtifacts@0` in the azure pipeline yaml file and see whether you can pick single packages instead of downloading everything or if it will silently fail if a package is already in the dir? (It looks that is failing because we are downloading the Redist from all builds).

I confirm that it works :) @artidoro if you can fix the pipeline thing I will then merge it.

artidoro:
Yes I'll take a look as soon as possible!

artidoro:
I think we can prevent the test builds from building packages to fix the error we were getting.

I am not sure if it's going to work like this though... I am afraid we will get all the dlls build by all the builds which is going to duplicate some since we don't just have one build per OS. Let's see!

interesaaat:
> I think we can prevent the test builds from building packages to fix the error we were getting.
> 
> I am not sure if it's going to work like this though... I am afraid we will get all the dlls build by all the builds which is going to duplicate some since we don't just have one build per OS. Let's see!

Yes I think this is the right strategy to get the check faster. However the problem is not there I believe. I think that the problem is that we create the redist stuff for each build before the packaging step.

PR #108, opened 2019-07-25
Add GPU Support

### LINKED ISSUE #98: Add redists for the GPU versions of libtorch

artidoro:
Fixes #98.

In this PR I add GPU support through the following:
1. A new project that downloads the CUDA version of LibTorch, extracts the dlls that are needed
2. A new nuget definition for LibTorch.GPU.Redist which contains the CUDA enabled LibTorch dlls and LibTorchSharp (our native code)
3. I tested that using the LibTorch.GPU.Redist nuget a very simple program which multiplies two matrices would run the computation on GPU. I used a GPU Linux machine (with Ubuntu 19.04) and checked that `nvidia-smi` would show the GPU usage

This is not a very thorough test and it should also be tested on Windows. But it is definitely a starting point. I am not sure how we can request GPU machines on DevOps, it would definitely be useful to run tests as part of the CI.

Note that mac does not have support for GPU (as expected).


pkg/LibTorch.GPU.Redist/LibTorch.GPU.Redist.nupkgproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk" DefaultTargets="Pack">
+
+  <PropertyGroup>
+    <Authors></Authors>
+    <Copyright>Copyright (c) Microsoft Corporation</Copyright>
+    <Owners>Microsoft Corporation</Owners>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <PackageDescription>$(MSBuildProjectName) contains the PyTorch C++ frontend LibTorch library version $(LibTorchVersion) 
+    for GPU with Cuda redistributed as a NuGet package. </PackageDescription>
```
artidoro: We should probably specify that this is using LibTorch with CUDA 9.0

interesaaat: Yes. Also can we change the name of the redist? Something like LibTorch.GPU.CUDA.9.0.Redist? So that eventually we can create a CUDA 10 as well

interesaaat:
Also should we add something in the README file on how to consume TorchSharp / the redist packages? Like get the torchsharp nuget plus the redist package for your target compute (i.e., CPU or CUDA 9.0).

pkg/LibTorch.GPU.Redist/LibTorch.GPU.Redist.nupkgproj
artidoro: Yes I think it's a good idea.

artidoro:
Yes good point. Let me add that to the readme as well.

interesaaat:
Although we didn't test this thoroughly I will commit this since it is not breaking anything on the CPU path.

PR #109, opened 2019-07-25
Adding example commands to build on Ubuntu

artidoro:


interesaaat:
More than example commands this is to get all the dependencies, right? Should we move the section under requirements and say something like example commands to get the required dependencies (tested on Ubuntu 16)?

PR #114, opened 2019-07-31
avg pool 2d

sumitdvlp:
avg pool 2d is required for VGG , 
i have used maxpool2d example and https://pytorch.org/cppdocs/api/function_namespaceat_1a7341b8ebb8cfbb5cd252da45997b80b5.html


artidoro:
Here is where we have our tests:
https://github.com/xamarin/TorchSharp/tree/master/test/TorchSharpTest

test/TorchSharpTest/NN/NNModuleTest.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Linq;
+using System.Runtime.InteropServices;
+using TorchSharp.JIT;
+using TorchSharp.NN;
+using TorchSharp.Tensor;
+using Xunit;
+
+namespace TorchSharpTest.NN
+{
+    public class NNModuleTest
```
interesaaat: Can you please move this tests into TorchSharp test? Or alternatively, can you please move all NN modules tests in TorchSharp into NNModuleTest?

sumitdvlp: done moved test cases to TorchSharp test

PR #118, opened 2019-08-08
VGG example

### LINKED ISSUE #88: Add VGG to the Examples
[Paper](https://arxiv.org/abs/1409.1556)
[Example implementation](https://github.com/kuangliu/pytorch-cifar/blob/master/models/vgg.py) in PyTorch.

sumitdvlp:
VGG example with test and train methods.

Fixes #88.


src/Examples/VGG.cs
```diff
@@ -0,0 +1,211 @@
...
+
+            /// <summary>
+            /// 
+            /// </summary>
+            /// <param name="input"></param>
+            /// <returns></returns>
+            public override TorchTensor Forward(TorchTensor input)
+            {
+                TorchTensor res = this.sequential.Forward(input);
+                res = res.View(new long[] { -1, 0 });
```
interesaaat: I think this is wrong. You need to take the size of the first dimension of res.

src/Examples/VGG.cs
```diff
@@ -0,0 +1,211 @@
...
+                int in_channels = 3;
+
+                foreach (object x in cfg)
+                {
+                    if (x.ToString() == "M")
+                        layers.Add(Module.MaxPool2D(new long[2], new long[1]));
+                    else
+                    {
+                        layers.Add(Module.Conv2D(in_channels, (long)x, 3, padding: 1));
+                        //layers.Add(Module.Bat);
```
interesaaat: We have batchNorm now, you can unncomment this.

src/Examples/VGG.cs
```diff
@@ -0,0 +1,211 @@
...
+                        layers.Add(Module.Relu(inPlace: true));
+                        in_channels = (int)x;
+                    }
+                }
+                // layers.Add(Module.AdaptiveAvgPool2D();
```
interesaaat: We also have avg pool, so please add this as well.

src/Examples/VGG.cs
sumitdvlp: Yes ,
I think this should work
                res = res.View(new long[] { res.Shape[0], -1 });

src/Examples/VGG.cs
sumitdvlp: agree, I have added it, also I had to create a override function for avgpool2d which i have added with this PR

src/Examples/VGG.cs
sumitdvlp: I couldn't find it master branch, please help me to find where it is implemented

dsyme:
Closing this as it is old, we can revisit at a later point

PR #126, opened 2020-05-07
output => prediction

moloneymb:
Fixing a simple error.


interesaaat:
Oh wow thanks Matthew!

PR #127, opened 2020-05-12
Add additional operations to TorchTensor

dsyme:
This fills in some missing operations on TorchTensor.

@interesaaat Let me know what you think

### DIFF
--- a/src/TorchSharp/NN/ReLu.cs
+++ b/src/TorchSharp/NN/ReLu.cs
@@ -16,12 +16,9 @@ namespace TorchSharp.NN
             _inPlace = inPlace;
         }
 
-        [DllImport("LibTorchSharp")]
-        private static extern IntPtr THSNN_reluApply(IntPtr tensor, bool inPlace);
-
         public override TorchTensor Forward(TorchTensor tensor)
         {
-            return new TorchTensor(THSNN_reluApply(tensor.Handle, _inPlace));
+            return _inPlace ? tensor.ReluInPlace() : tensor.Relu(); 
         }
 
         public override string GetName()
--- a/src/TorchSharp/Tensor/TorchTensor.cs
+++ b/src/TorchSharp/Tensor/TorchTensor.cs
@@ -192,7 +192,7 @@ namespace TorchSharp.Tensor
         [DllImport("LibTorchSharp")]
         private static extern sbyte THSTensor_type(IntPtr handle);
 
-        public ATenScalarMapping Type => (ATenScalarMapping) THSTensor_type(handle);
+        public ATenScalarMapping Type => (ATenScalarMapping)THSTensor_type(handle);
 
         [DllImport("LibTorchSharp")]
         private static extern string THSTensor_deviceType(IntPtr handle);
@@ -214,7 +214,7 @@ namespace TorchSharp.Tensor
 
         public TorchTensor ToType(ATenScalarMapping type)
         {
-            return new TorchTensor(THSTensor_to_type(handle, (sbyte) type));
+            return new TorchTensor(THSTensor_to_type(handle, (sbyte)type));
         }
 
         [DllImport("LibTorchSharp")]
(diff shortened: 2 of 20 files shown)


src/TorchSharp/NN/ReLu.cs
```diff
@@ -16,12 +16,9 @@ internal ReLU(bool inPlace = false) : base()
             _inPlace = inPlace;
         }
 
-        [DllImport("LibTorchSharp")]
-        private static extern IntPtr THSNN_reluApply(IntPtr tensor, bool inPlace);
```
dsyme, line 20: This signature doesn't match the signature in the native DLL (which has no inPlace argument)

src/TorchSharp/Tensor/TorchTensor.cs
```diff
@@ -382,11 +407,11 @@ public TorchTensor Transpose(long dimension1, long dimension2)
         }
 
         [DllImport("LibTorchSharp")]
-        private static extern void THSTensor_transpose_(IntPtr src, long dim1, long dim2);
+        private static extern IntPtr THSTensor_transpose_(IntPtr src, long dim1, long dim2);
```
dsyme, line 410: The underlying native entrypoint returns an IntPtr, so putting that here too.

interesaaat:
Thanks @dsyme for the PR. Looks good to me, but for some reason it is failing a test for `SplitWithSizesTest` on the mac build. Do you think you can look into it? I will merge afterwards.

dsyme:
Just recording this for posterity: I had a bug when passing an optional Tensor to the C++ API.  

The Torch C++ methods often use optional arguments, but we generally need to pass them all through from C# and make them optional in C#, filling in the right default values.

When a Tensor is optional however you can't pass null into the C++ API, you have to pass a pointer to an empty tensor.

Compare
```
Tensor THSTensor_conv_transpose1d(
    const Tensor input, const Tensor weight, const Tensor bias,
    const int64_t* stride, const int strideLength,
    const int64_t* padding, const int paddingLength,
    const int64_t* outputPadding, const int outputPaddingLength,
    const int64_t* dilation, const int dilationLength,
    int64_t groups)
{
    auto res =
        torch::conv_transpose1d(*input, *weight, (bias ? *bias : at::Tensor()),
            at::IntList(stride, strideLength),
            at::IntList(padding, paddingLength),
            at::IntList(outputPadding, outputPaddingLength),
            groups,
            at::IntList(dilation, dilationLength));
    return new torch::Tensor(res);
}
```
and
```
Tensor THSTensor_conv_transpose1d(
    const Tensor input, const Tensor weight, const Tensor bias,
    const int64_t* stride, const int strideLength,
    const int64_t* padding, const int paddingLength,
    const int64_t* outputPadding, const int outputPaddingLength,
    const int64_t* dilation, const int dilationLength,
    int64_t groups)
{
    auto res =
        torch::conv_transpose1d(*input, *weight, *bias,
            at::IntList(stride, strideLength),
            at::IntList(padding, paddingLength),
            at::IntList(outputPadding, outputPaddingLength),
            groups,
            at::IntList(dilation, dilationLength));
    return new torch::Tensor(res);
}
```
Note the bias is an optional Tensor.  The first formulation is correct.

dsyme:
> Thanks @dsyme for the PR. Looks good to me, but for some reason it is failing a test for SplitWithSizesTest on the mac build. Do you think you can look into it? I will merge afterwards.

I'm looking into this - perhaps it is to do with the memory leak bug.

interesaaat:
Was this meant to be closed or merged?

dsyme:
I have updated this with 

* [x] adding exception handling for most TorchTensor tensor operations 
* [x] testing for convolutions
* [x] maxpool and conv_transpose functions
* [x] SByteTensor

dsyme:
> Was this meant to be closed or merged?

This was a mistake, GitHub automatically closed it when I accidentally pushed the "ops1" branch to this repo instead of my personal one

dsyme:
@interesaaat This is now ready - it appears the Mac bug was due to the memory corruptions when pinned objects got moved around by the GC

Issue #128, opened 2020-05-13
Remove NetFX build from this repo


dsyme:
I think we should remove the .NET Framework build from this repo, in aid of simplification, and just focus on .NET Core.

@interesaaat @migueldeicaza That ok?

dsyme:
@interesaaat Just to let you know that @migueldeicaza has granted me co-maintainer rights in this repo and for the nuget packages

My aim is limited - to get to the point where we can use the libtorch tensor implementation as the backend for DiffSharp, see https://github.com/DiffSharp/DiffSharp/pull/106.  Roughly I expect this will mean:

* [ ] Get changes such as https://github.com/xamarin/TorchSharp/pull/127
* [ ] Update the nuget packages 
* [ ] Update TorchSharp to more recent pytorch redist
* [ ] Simplify TorchSharp where it makes sense for a long term trajectory (e.g. this issue)

interesaaat:
I am all for it!

migueldeicaza:
Sure - it is just that .NET Core is still slower to build/debug than .NET Framework is on Mac, hence why I prefer those.   But I am not working on this anymore, so up to you Don.

gbaydin:
> @interesaaat Just to let you know that @migueldeicaza has granted me co-maintainer rights in this repo and for the nuget packages
> 
> My aim is limited - to get to the point where we can use the libtorch tensor implementation as the backend for DiffSharp, see [DiffSharp/DiffSharp#106](https://github.com/DiffSharp/DiffSharp/pull/106). Roughly I expect this will mean:
> 
> * [ ]  Get changes such as #127
> * [ ]  Update the nuget packages
> * [ ]  Update TorchSharp to more recent pytorch redist
> * [ ]  Simplify TorchSharp where it makes sense for a long term trajectory (e.g. this issue)

This is awesome!

Issue #67, opened 2018-11-16
Copyright notices


NiklasGustafsson:
We should add an appropriate standard copyright notice to all the files.

dsyme:
Closing as part of https://github.com/xamarin/TorchSharp/pull/129

PR #129, opened 2020-05-14
clean out Aten, SNT, netfx

dsyme:
In discussion with @interesaaat we propose the following to simplify this repo

1. Remove AtenSharp.  The PyTorch C++ API is now the correct layer to wrap

2. Remove the SNT implementation built on ATenSharp.  It is better if this repo focus on the TorchTensor layer wrapping the PyTorch 1.5 C++ API 

3. Remove .NET Framework support

4. Always build for both netstandard 2.1 and netcoreapp3.0 (i.e. with intrinsics support)

See issue #128

Also addresses #67 - copyright notices


dsyme:
Note it's possible someone might want to resurrect the System.Numerics.Tensor/AtenSharp bindings, but these should I think be in a separate repo.

dsyme:
I am archiving the AtenSharp and STN portions of this repository at https://github.com/fsprojects/AtenSharp

PR #130, opened 2020-05-15
IO operations

dsyme:
This replays the substantial part of https://github.com/xamarin/TorchSharp/pull/115 over master


dsyme:
This needs update to 1.1.0, which I'll push through separately

PR #144, opened 2020-06-01
Also test on Cuda binaries

dsyme:
Adds testing using the actual Cuda binaries (not on GPU, but using the CUDA binaries on CPU)


dsyme:
This is actually quite difficult on the CI machines we have, and doesn't gain much.  Closing for now

Issue #146, opened 2020-06-02
Resurrect JIT functionality and loading of models


dsyme:
The PyTorch C++ guide shows that loading of modules means using the JIT loading: https://pytorch.org/tutorials/advanced/cpp_export.html

The JIT functionality was removed after the extensive churn in Pytorch C++ API in v1.01 --> 1.50. 
 It needs to be resurrected.

The more direct Load/Save that was present on modules should probably not be resurrected directly.

Also need to add ONNX load/save support

dsyme:
@NiklasGustafsson What's your thinking on this for the roadmap for TorchSharp?  It feels like it should be there?

NiklasGustafsson:
I think this is one area that requires very careful consideration. I'm not convinced that TorchSharp needs to follow the PyTorch recipe for externalizing models exactly.

While we have spent a lot of effort making the eager evaluation of models as close as possible to what PyTorch developers are doing, we do not __have__ to make the externalized model experience the same. For example, there may be a role for Roslyn analyzers for the C# solution, rather than a clone of the Python solution. Likewise for F# -- there may be a F#-specific solution.

I think this should come post-v1.0, and we also need to consider how we will support ONNX export from TorchSharp. To me, they are two sides of the same coin.

GeorgeS2019:
@dsyme @NiklasGustafsson 

Currently, I could be wrong, it seems the continuity from pytorch saved model to Torchsharp is through ONNX. 

There is a need for a proof of concept that the state dictionary exported (using TorshSharp export.py script) from pyTorch can be loaded back to TorchSharp by having the requirement that both the TorchSharp and PyTorch models are compatible.

As Nikas has implied in a few occasions, one possible option is the TorchSharp support for ONNX import of PyTorch exported ONNX. 

It seems TorchSharp ONNX import is equally and perhaps more important than TorchSharp ONNX export.

Again, I hope this issue is evaluated post-v1.0

Congratulation of rapid progress towards v1.0!

dsyme:
@NiklasGustafsson I see for loading (though I'd still think we need to support loading of the pytorch format, if Java or Rust or Scala bindings do, we should check) 

What about JIT? That's really what I was asking about, I forgot this issue referred to loading too.

dsyme:
Here are relevant links from Rust PyTorch bindings

https://github.com/LaurentMazare/tch-rs#using-some-pre-trained-model

[...]

PR #147, opened 2020-06-02
start of #146 (Load/Save models, JIT)

dsyme:
This is the start of https://github.com/xamarin/TorchSharp/issues/146

* [ ] The existing Load/Save on the module type is broken because the incorrect C++ Impl object is created for the loaded model (e.g. `torch::nn::Module` not `torch::nn::ReLUImpl`).  It should be removed.

* [ ] To get loading according to the C++ PyTorch docs, the JIT functionality needs to be brought back.


dsyme:
I'm closing this as I'm not personally planning to work on it at this stage

PR #153, opened 2020-06-08
add some memory tests

dsyme:
Add some tests that we are disposing of memory


gbaydin:
Hi @dsyme how do these test check whether memory is disposed or not? I'm confused because there are no assertions or checks for memory usage in the test code.

Issue #152, opened 2020-06-08
Discussion: Disposal and TorchTensor structness


dsyme:
@interesaaat @migueldeicaza  I have a question about the decision to make TorchTensor a struct

TorchTensor captures an unmanaged resources and these resources need to be reliably freed.  The normal way to do this in .NET code is to add a finalizer on an object wrapper. 

However TorchTensor became a struct, and so this places the burden on the user of TorchTensor to hold the TorchTensor in a class and add a finalizer to that.

That is an OK and in DiffSharp we do have an object that can act as this holder.   But I just wanted to check that we are ok with this design.

Additionally, since such a holder must necessarily be a unique holding object - you can't copy the struct around feeely and dispose it multiple times.

How important is it to make TorchTensor a struct anyway?  It feels like it is optimizing the wrong thing unless tensors really are small.

Looking about how I've been coding with TorchTensors in DiffSharp, as an example, I've been freely writing code like this (F# syntax, but the same applies in C#):
```fsharp
tt.[4].DataItem<double>()
```
and that code is incorrect - it would have to be 
```fsharp
use tt2 = tt[4]
return tt2.DataItem<double>()
```
to ensure correct disposal.  There is no warning here, no guidance, you're just left exposed. That's too buggy for my liking, an optimization that might be accessible via some low level `TorchValueTensor` but not for the type exposed in the most common programming model.

dsyme:
See also https://github.com/DiffSharp/DiffSharp/pull/133

dsyme:
TBH my proposal is

1. We make TorchTensor a class (wrapping the existing ref-counted C++ `shared_ptr*`)

2. In the long term we consider adding a `TorchTensorImpl` which is a struct wrapping a non-ref-counted C++ `at::TensorImpl *` pointer (thus eliminating two allocations - one .NET TorchTensor and one C++ `at::Tensor` reference counted wrapper).  

The half-way house we have at the moment TorchTensor is a struct wrapping an inaccurately-counted `at::Tensor` just seems wrong.

migueldeicaza:
I agree that this should be a class - otherwise we end up with even more complicated scenarios later involving copying.

dsyme:
Yes, thanks.

interesaaat:
[...]

PR #154, opened 2020-06-08
make TorchTensor a class (to avoid memory leaks)

dsyme:
See #152

### DIFF
--- a/src/TorchSharp/Tensor/TorchTensor.cs
+++ b/src/TorchSharp/Tensor/TorchTensor.cs
@@ -5,9 +5,10 @@ using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Text;
 
+#nullable enable
 namespace TorchSharp.Tensor
 {
-    public struct TorchTensor : IDisposable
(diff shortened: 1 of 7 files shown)


src/TorchSharp/Tensor/TorchTensor.cs
```diff
@@ -1843,15 +1845,15 @@ public TorchTensor Slice(long dimension, long start, long finish, long step)
...
+            long[]? strides = null,
+            long[]? padding = null,
+            long[]? dilation = null,
             long groups = 1)
         {
             strides = (strides == null) ? new long[] { 1 } : strides;
             padding = (padding == null) ? new long[] { 0 } : padding;
             dilation = (dilation == null) ? new long[] { 1 } : dilation;
-            var biasHandle = (bias.HasValue ? bias.Value.Handle : IntPtr.Zero);
+            var biasHandle = (bias is null ? IntPtr.Zero : bias.Handle);
```
lostmsu, line 1856: NIT: you should be able to just write `bias ?? IntPtr.Zero`

PR #181, opened 2020-11-12
Added torch.gcd & torch.lcm

skyline75489:
Hi, this is my first PR. Just wanna to make sure I'm understanding the project correctly.


dsyme:
@skyline75489 Looks good.  Feel free to add tests if you like.  I haven't been adding tests for everything, mainly because the implementations are straight-forward and the operators are primarily being consumed by DiffSharp.  However if you think it's best to add tests into this layer for your purposes then please do.

PR #182, opened 2020-11-13
Use macros to simplify libtorch interop code

skyline75489:
Here's an idea of mine, that can simplify the code. Don't know if you'll like it or not. If you don't feel like this, feel free to ignore the PR.

### DIFF
--- a/src/Native/LibTorchSharp/THSTensor.cpp
+++ b/src/Native/LibTorchSharp/THSTensor.cpp
@@ -4,26 +4,6 @@
 #include <iostream>
 #include <fstream>
 
-Tensor THSTensor_abs(const Tensor tensor)
-{
-    CATCH_TENSOR(tensor->abs());
-}
-
-Tensor THSTensor_abs_(const Tensor tensor)
-{
-    CATCH_TENSOR(tensor->abs_());
-}
-
-Tensor THSTensor_acos(const Tensor tensor)
-{
-    CATCH_TENSOR(tensor->acos());
-}
-
-Tensor THSTensor_acos_(const Tensor tensor)
-{
-    CATCH_TENSOR(tensor->acos_());
-}
-
 Tensor THSTensor_add(const Tensor left, const Tensor right, const Scalar alpha)
 {
     CATCH_TENSOR(left->add(*right, *alpha));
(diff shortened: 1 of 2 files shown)


dsyme:
@skyline75489 It looks good!  There's some weirdness on Linux/gcc

```
/__w/1/s/src/Native/LibTorchSharp/THSTensor.h(44,1): error G8699F740: pasting formed '(abs', an invalid preprocessing token [/__w/1/s/src/Native/build.proj]
  EXPORT_THS_TENSOR_API_NO_ARG(abs);
```

skyline75489:
Sorry i only tested it on MSVC. I was worrying that it might overcomplicate things. I think the best approach would be using a tool to generated all the public API for libtorch. But i haven't got a detailed plan.

获取 Outlook for Android<https://aka.ms/ghei36>

skyline75489:
And by the way, is C++/CLI working with Linux toolchain? I once wrote a C++/CLI project and it will automatically handle the PInvoke process. Basically no C# PInvoke is needed any more

dsyme:
> Sorry i only tested it on MSVC. I was worrying that it might overcomplicate things. I think the best approach would be using a tool to generated all the public API for libtorch. But i haven't got a detailed plan.

Thanks, yes, @moloneymb did some work on code-generating

You're right, I do think the macros complicate things a little and maybe isn't worth it at this stage, since each is so easy to write.

> And by the way, is C++/CLI working with Linux toolchain? I once wrote a C++/CLI project and it will automatically handle the PInvoke process. Basically no C# PInvoke is needed any more

I'm not so keen to take a dependency on C++/CLI unless it gives concrete performance benefits.

skyline75489:
Good to know. I'm wondering what work has been done about code generation previously. I found there's a yaml that describes all the interfaces (https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/native_functions.yaml). Seems a perfect source for code generation.

dsyme:
> Good to know. I'm wondering what work has been done about code generation previously. I found there's a yaml that describes all the interfaces (https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/native_functions.yaml). Seems a perfect source for code generation.

Yes, that's what @moloneymb was using. However there were enough corner cases in the generation that it was perhsps more trouble than its worth, given that we also should be testing each function.

This may change if we can also code-generate the Layer/Modules APIs .  Because they use C++ a lot they are trickier to generate.

moloneymb:
FYI; my WIP code generation script is in this largely self contained script; https://github.com/moloneymb/TorchCodeGen/blob/master/src/GenerateCode.fsx

I think this method of code generation is flexible enough to handle the corner cases.

My planned next step was to move on from reproducing the existing interop code to pretty printing a C++ and C# AST. This would be less hacky and clean up the code a bit.

However, this work is on hold for the foreseeable future due to a change in priorities.

dsyme:
@skyline75489 I've thought this over and think we shouldn't use macros like this, and instead eventually pursue full and proper code generation of both C# and C++ based of LibTorch header files or the other inputs @moloneymb was using

PR #187, opened 2020-11-20
fix signatures for avg pool

dsyme:


src/TorchSharp/Tensor/TorchTensor.cs
```diff
@@ -3864,12 +3866,14 @@ public TorchTensor MaxUnpool3D(TorchTensor indices, long[] outputSize, long[] st
                 bool ceil_mode,
                 bool count_include_pad);
 
-        public TorchTensor AvgPool3D(long kernelSize, long? stride = null,
-            long? padding = null, bool ceil_mode = false, bool count_include_pad = true)
+        public TorchTensor AvgPool3D(long[] kernelSizes,
+            long[]? strides,
```
lostmsu, line 3870: Shouldn't `strides` have a default parameter value of `null`?

PR #210, opened 2021-03-22
Optimizers

NiklasGustafsson:
Added RMSprop and Adagrad optimizers and corresponding unit tests.

### DIFF
--- a/src/Native/LibTorchSharp/THSNN.cpp
+++ b/src/Native/LibTorchSharp/THSNN.cpp
@@ -499,19 +499,34 @@ Tensor THSNN_poisson_loss(const Tensor input, const Tensor target, const bool lo
     )
 }
 
-Optimizer THSNN_Adam_ctor(const Tensor* parameters, const int length, const double learnig_rate)
+Optimizer THSNN_Adagrad_ctor(const Tensor* parameters, const int length, const double learning_rate, const double lr_decay, const double weight_decay)
 {
-    auto  params = toTensors<at::Tensor>((torch::Tensor**)parameters, length);
+    auto params = toTensors<at::Tensor>((torch::Tensor**)parameters, length);
+    auto options = torch::optim::AdagradOptions(learning_rate).lr_decay(lr_decay).weight_decay(weight_decay);
 
-    auto optimizer = torch::optim::Adam(params, learnig_rate);
-
-    return new std::shared_ptr<torch::optim::Optimizer>(std::make_shared<torch::optim::Adam>(torch::optim::Adam(params, learnig_rate)));
+    return new std::shared_ptr<torch::optim::Optimizer>(std::make_shared<torch::optim::Adagrad>(torch::optim::Adagrad(params, options)));
 }
 
-Optimizer THSNN_SGD_ctor(const Tensor* parameters, const int length, const double learnig_rate, const double momentum)
+Optimizer THSNN_Adam_ctor(const Tensor* parameters, const int length, const double learning_rate)
 {
     auto  params = toTensors<at::Tensor>((torch::Tensor**)parameters, length);
-    auto opts = torch::optim::SGDOptions(learnig_rate).momentum(momentum);
+
+    return new std::shared_ptr<torch::optim::Optimizer>(std::make_shared<torch::optim::Adam>(torch::optim::Adam(params, learning_rate)));
+}
+
+Optimizer THSNN_RMSprop_ctor(const Tensor* parameters, const int length, const double learning_rate, const double alpha)
+{
+    auto  params = toTensors<at::Tensor>((torch::Tensor**)parameters, length);
+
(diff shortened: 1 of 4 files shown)


dsyme:
Looks great!

src/Native/LibTorchSharp/THSNN.cpp
```diff
@@ -499,19 +499,34 @@ Tensor THSNN_poisson_loss(const Tensor input, const Tensor target, const bool lo
...
+{
+    auto params = toTensors<at::Tensor>((torch::Tensor**)parameters, length);
+    auto options = torch::optim::AdagradOptions(learning_rate).lr_decay(lr_decay).weight_decay(weight_decay);
+
+    return new std::shared_ptr<torch::optim::Optimizer>(std::make_shared<torch::optim::Adagrad>(torch::optim::Adagrad(params, options)));
+}
+
+Optimizer THSNN_Adam_ctor(const Tensor* parameters, const int length, const double learning_rate)
+{
+    auto  params = toTensors<at::Tensor>((torch::Tensor**)parameters, length);
```
dsyme, line 512: There is an AdamOptions we should be using here https://pytorch.org/cppdocs/api/structtorch_1_1optim_1_1_adam_options.html?highlight=adamoptions#_CPPv4N5torch5optim11AdamOptionsE

dsyme: Also this has numerous additional parameters available on it

NiklasGustafsson: Yes, we can add that. There are a lot of options that are being skipped for the various optimizers. I can add them systematically.

src/Native/LibTorchSharp/THSNN.cpp
```diff
@@ -499,19 +499,34 @@ Tensor THSNN_poisson_loss(const Tensor input, const Tensor target, const bool lo
     )
 }
 
-Optimizer THSNN_Adam_ctor(const Tensor* parameters, const int length, const double learnig_rate)
+Optimizer THSNN_Adagrad_ctor(const Tensor* parameters, const int length, const double learning_rate, const double lr_decay, const double weight_decay)
+{
+    auto params = toTensors<at::Tensor>((torch::Tensor**)parameters, length);
+    auto options = torch::optim::AdagradOptions(learning_rate).lr_decay(lr_decay).weight_decay(weight_decay);
```
dsyme, line 505: What happens about the optionality of these?  

We don't have any consistent way of passing across optional values across the C API, just adhoc uses of NULL, min values and so on

src/TorchSharp/NN/Optimizer.cs
```diff
@@ -78,10 +78,40 @@ protected void Dispose (bool disposing)
...
+        /// <summary>
+        /// Implements RMSprop algorithm.
+        ///
+        /// Proposed by G.Hinton in his course.
+        /// </summary>
+        /// <param name="parameters">Prameters to optimize</param>
+        /// <param name="learningRate">Learning rate (default: 1e-2)</param>
+        /// <param name="alpha">Smoothing constant (default: 0.99)</param>
+        /// <returns></returns>
+        public static Optimizer RMSProp(IEnumerable<TorchTensor> parameters, double learningRate = 0.01, double alpha = 0.99)
```
dsyme, line 93: I assume you double checked the default valus with the pytorch docs?

src/Native/LibTorchSharp/THSNN.cpp
NiklasGustafsson: We should remove the defaults when crossing the interop boundary.

src/TorchSharp/NN/Optimizer.cs
NiklasGustafsson: Yes.

src/Native/LibTorchSharp/THSNN.cpp
dsyme: Thanks.  If you skip anything knowingly please make sure there's a note of some form in an open bug e.g. just add it to one of the existing bugs listing missing things

src/Native/LibTorchSharp/THSNN.cpp
NiklasGustafsson: I'll do that next.

dsyme: OK yes in this case that can be done.  

In some other cases I've so far made the optional explicit, e.g. see 
* `THSTensor_avg_pool2d_backward` `divisor_override` 
* `THSTensor_fft` 
* `THSTensor_upsample_nearest1d` - `outputSize` and `scaleFactors`

and so on. There are about 25 explicit uses of `c10::optional`

If you think it's better to simplify these out to optionality on the C# side please go ahead

PR #214, opened 2021-03-23
Convolution, 1D pooling, BatchNorm

### LINKED ISSUE #188: Implement BatchNorm2D class
Hello, Thank you very much for your great work on this project!
I am currenty working on neural network using your package, but I miss the BatchNorm2D layer.
Will it be implemented in a next release, and if yes when?
If no and if you are interested, I can contribute on this one.
Let me know, have an excellent day!

NiklasGustafsson:
Adding 1D pooling, Conv1D, and BatchNormalization modules and corresponding unit tests.
Closes #188.

### DIFF
--- a/src/TorchSharp/NN/AdaptiveAvgPool2D.cs
+++ b/src/TorchSharp/NN/AdaptiveAvgPool2D.cs
@@ -29,7 +29,29 @@ namespace TorchSharp.NN
         [DllImport ("LibTorchSharp")]
         extern static IntPtr THSNN_AdaptiveAvgPool2d_ctor (IntPtr psizes, int length, out IntPtr pBoxedModule);
 
-        static public AdaptiveAvgPool2D AdaptiveAvgPool2D (long[] kernelSize)
+        /// <summary>
+        /// Applies a 2D adaptive average pooling over an input signal composed of several input planes.
+        /// The output is of size H x W, for any input size.The number of output features is equal to the number of input planes.
+        /// </summary>
+        /// <param name="outputSize">The target output size of the image of the form H x H.</param>
+        /// <returns></returns>
+        static public AdaptiveAvgPool2D AdaptiveAvgPool2D((long, long) outputSize)
+        {
+            return AdaptiveAvgPool2D(new long[] { outputSize.Item1, outputSize.Item2 });
+        }
+
+        /// <summary>
+        /// Applies a 2D adaptive average pooling over an input signal composed of several input planes.
+        /// The output is of size H x W, for any input size.The number of output features is equal to the number of input planes.
+        /// </summary>
+        /// <param name="outputSize">The target output size (H,W) of the image of the form H x W.</param>
+        /// <returns></returns>
+        static public AdaptiveAvgPool2D AdaptiveAvgPool2D(long outputSize)
+        {
+            return AdaptiveAvgPool2D(new long[] { outputSize, outputSize });
+        }
+
+        static private AdaptiveAvgPool2D AdaptiveAvgPool2D (long[] kernelSize)
         {
             unsafe {
                 fixed (long* pkernelSize = kernelSize) {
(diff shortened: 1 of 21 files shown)


dsyme:
It's looking good so far.  Freel free to keep it open with "WIP" or mark as "draft"

NiklasGustafsson:
That's it. I also made sure to add a full set of parameters to the convolution modules/functions, which is why I held this back earlier.

src/TorchSharp/NN/AdaptiveAvgPool2D.cs
```diff
@@ -29,7 +29,29 @@ public static partial class Modules
...
         extern static IntPtr THSNN_AdaptiveAvgPool2d_ctor (IntPtr psizes, int length, out IntPtr pBoxedModule);
 
-        static public AdaptiveAvgPool2D AdaptiveAvgPool2D (long[] kernelSize)
+        /// <summary>
+        /// Applies a 2D adaptive average pooling over an input signal composed of several input planes.
+        /// The output is of size H x W, for any input size.The number of output features is equal to the number of input planes.
+        /// </summary>
+        /// <param name="outputSize">The target output size of the image of the form H x H.</param>
+        /// <returns></returns>
+        static public AdaptiveAvgPool2D AdaptiveAvgPool2D((long, long) outputSize)
```
dsyme, line 38: `(long, long)` is not idiomatic I think - we currently aren't using tuples in this way for known-length integer vector parameters, nor would it be idiomatic replacement for two parameters

So I think use this to avoid inconsistency - if we decided to adopt tuples it should be done systematically

```
static public AdaptiveAvgPool2D AdaptiveAvgPool2D(long outputLength, long outputWidth)
```

dsyme: Hmmm it's a `long[]` here: https://github.com/xamarin/TorchSharp/pull/214/files#diff-ec3f83c280eb9ad9e93409d310c45340d5650be301e30845750cb40ffc9671bcR38

So either two parameters and three parameters in each, or a `long[]` in each - and whichever it is should be consistent with other modules.

dsyme: Or just remove this overload.

dsyme:
It look great - I added one comment about the `(long, long)` overload - I think just remove it and we can look at that issue systematically as part of API ergonomics for the C# user

BTW when you get a chance please check the module names all the same as in PyTorch, e.g. with respect to capitalization and so on

src/TorchSharp/NN/AdaptiveAvgPool2D.cs
NiklasGustafsson: This was an attempt to encode the PyTorch semantics, but to do so in a statically typed way. Here's what the PT docs have to say:
```
output_size – the target output size of the image of the form H x W. Can be a tuple (H, W) or a single H for a square image H x H. H and W can be either a int, or None which means the size will be the same as that of the input.
```

So, in Python, you would call the operator either with something like '720', or '(28,28)' -- with the overloads, you get the same in C#.

NiklasGustafsson:
>> BTW when you get a chance please check the module names all the same as in PyTorch, e.g. with respect to capitalization and so on

I can do that systematically in a separate PR -- in this, I was following the existing scheme in TorchSharp, which had '2d' -> '2D' and the same for one- and three-dimensional module names. Otherwise, it should already be the same capitalizations as in PyTorch.

src/TorchSharp/NN/AdaptiveAvgPool2D.cs
NiklasGustafsson: But you are right, it's inconsistently not a tuple on the 3D module.

dsyme: Yes, please do this kind of API ergonomics as a systematic separate end-to-end activity, based off like-for-like python-C#-F# API usage examples.   C#, F# and .NET have numerous ways to get the ergonomic effects we need - some of these multiply the API in one or more dimensions (e.g. overloads), so the techniques have to be used carefully.  

Python also offers many different ergonomic options and which of these we allow needs to be considered carefully.  

From the F# perspective e.g. DiffSharp [this RFC](https://github.com/fsharp/fslang-design/blob/master/RFCs/FS-1092-anonymous-type-tagged-unions.md) is also related as a way to reduce the overload-dimensionality-expansion problem.

So in these PRs I'd recommend dealing with the core capabilities for native code first.

NiklasGustafsson: Okay, I removed the overloads on the 2D version. When the CI is done, I'll merge.

dsyme: Cool. We certainly need excellent C# and F# ergonomics - just want to separate that as an activity we can reason about separately.

dsyme (approved):

src/TorchSharp/NN/AdaptiveAvgPool2D.cs
NiklasGustafsson: That's fair.

dsyme: There may also be ergonomic technical tricks we can learn from Tensorflow.NET here

NiklasGustafsson:
Most of the work is monkey-editing to put the C interfaces in place, and then copy those to the .NET code. The C++ module implementations are already there, so it's straight-forward.

PR #217, opened 2021-03-23
Loss functions

NiklasGustafsson:
Added:
-  a number of the most common loss functions.
- one-hot encoding
- Embedding module

Moved the loss functions to the 'Functions' static class, since there's no 'nn.losses' in PyTorch.

### DIFF
--- a/src/TorchSharp/NN/Losses.cs
+++ b/src/TorchSharp/NN/Losses.cs
@@ -16,14 +16,39 @@ namespace TorchSharp.NN
     {
 
         [DllImport ("LibTorchSharp")]
-        private static extern IntPtr THSNN_binary_cross_entropy (IntPtr srct, IntPtr trgt, IntPtr wgt, long reduction);
+        private static extern IntPtr THSNN_cross_entropy(IntPtr srct, IntPtr trgt, IntPtr wgt, long ignore_index, bool hasII, long reduction);
 
-        public static Loss binary_cross_entropy (TorchTensor? weigths = null, Reduction reduction = Reduction.Mean)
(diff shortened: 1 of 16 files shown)


dsyme:
This looks good so far, I'll review again when out of draft

src/TorchSharp/NN/Losses.cs
```diff
@@ -12,21 +12,46 @@ namespace TorchSharp.NN
...
         [DllImport ("LibTorchSharp")]
-        private static extern IntPtr THSNN_binary_cross_entropy (IntPtr srct, IntPtr trgt, IntPtr wgt, long reduction);
+        private static extern IntPtr THSNN_cross_entropy(IntPtr srct, IntPtr trgt, IntPtr wgt, long ignore_index, bool hasII, long reduction);
 
-        public static Loss binary_cross_entropy (TorchTensor? weigths = null, Reduction reduction = Reduction.Mean)
+        public static Loss cross_entropy_loss(TorchTensor? weigths = null, long? ignore_index = null, Reduction reduction = Reduction.Mean)
         {
             return (TorchTensor src, TorchTensor target) => {
-                var res = THSNN_binary_cross_entropy (src.Handle, target.Handle, weigths?.Handle ?? IntPtr.Zero, (long)reduction);
+                var ii = ignore_index.HasValue ? ignore_index.Value : -100;
```
dsyme, line 24: where's the -100 come from?

dsyme (approved):

src/TorchSharp/NN/Losses.cs
NiklasGustafsson: This is the default value in the Torch C++ code for this option. Since we do all the defaults on the C# side of the interface, I thought I should use the same.

Issue #218, opened 2021-03-24
Linux native binary portability issues


dsyme:
This is a tracking issue for Linux binary comptibility issues for the native code component.

Currently we build this on Ubuntu 18.04

-----------------------

Original issue:

The `libLibTorchSharp.so` we create has a GLIBC_2.30 dependency.  I don't fully understand why - it could be as simple as the fact that we build on an Ubunutu 20.04 machine, though it would be good to control the dependency better

Ubuntu 18.04 only has GLIBC_2.27 by default or something like that.

On an Ubuntu 28.04 Colab machine:

```
ldd /root/.nuget/packages/torchsharp/0.91.52458/runtimes/linux-x64/native/libLibTorchSharp.so

gnu/libpthread.so.0: version `GLIBC_2.30' not found (required by /root/.nuget/packages/torchsharp/0.91.52458/runtimes/linux-x64/native/libLibTorchSharp.so)
/root/.nuget/packages/torchsharp/0.91.52458/runtimes/linux-x64/native/libLibTorchSharp.so: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.26' not found (required by /root/.nuget/packages/torchsharp/0.91.52458/runtimes/linux-x64/native/libLibTorchSharp.so)
	linux-vdso.so.1 (0x00007fffcdd83000)
	/usr/lib/x86_64-linux-gnu/libtcmalloc.so.4 (0x00007f80bad9d000)
	libtorch.so => not found
	libc10.so => not found
	libtorch_cpu.so => not found
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f80bab7e000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f80ba7f5000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f80ba457000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f80ba23f000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f80b9e4e000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f80bb00d000)
	libunwind.so.8 => /usr/lib/x86_64-linux-gnu/libunwind.so.8 (0x00007f80b9c33000)
[...]

PR #220, opened 2021-03-24
investigate #218 - reduce libLibTorchSharp.so dependency to GLIBCXX_3.4.21

dsyme:
See https://github.com/xamarin/TorchSharp/issues/218


dsyme:
This causes a test to fail, closing for now

dsyme:
OK, I believe this has correctly reduced the dependency to GLIBCXX_3.4.21 by using an Ubuntu 16.04 container for the Linux build

PR #223, opened 2021-03-25
CUDA

NiklasGustafsson:
Changed how tensors are moved between devices so that it looks more like the Python API.
Added 'to()' method to modules, so that weights can be moved between devices.
Added a create_module<T1,T2>() function in the C++ code, to avoid unnecessary duplication of code.


dsyme (approved):

dsyme:
Looks great

PR #224, opened 2021-03-25
Restructure

NiklasGustafsson:
This is just a restructuring of the source code, no logic changes.
I introduced folders under 'NN' to separate by functionality,
I split the unit test file 'NN.cs' into three files, also organizing by functionality.


test/TorchSharpTest/TestTraining.cs
```diff
@@ -0,0 +1,516 @@
...
+        }
+
+        [Fact]
+        public void TestTrainingRMSAlpha()
+        {
+            var lin1 = Linear(1000, 100);
+            var lin2 = Linear(100, 10);
+            var seq = Sequential(("lin1", lin1), ("relu1", ReLU()), ("lin2", lin2));
+
+            var x = Float32Tensor.randn(new long[] { 64, 1000 });
```
dsyme, line 253: Just to mention as an aside that I'm not attached to this tensor creation API

DiffSharp uses this sort of thing:

    dsharp.tensor(obj, ...)

    dsharp.randn(..., dtype=Dtype.Float32)

It's quite possible we should support identical APIs or close to

    tsharp.tensor(obj, ...)

    tsharp.randn(..., dtype=....)

where replacing "torch.*" byt "tsharp" or "dsharp" gives fairly good results.

dsyme (approved):

dsyme:
I left a side note

Also, I'm not particularly tied to `TorchSharp.Tensor` namespace.  I'd be happy with `open TorchSharp` or anything that gives the ergonomics you want

I also noticed `TorchSharp.Device` rather than `TorchSharp.TorchDevice`.  Previously I was putting `Torch` on the front to help disambiguate w.r.t. DiffSharp "Tensor".  But I understand you'll likely need to switch to plain `Tensor`.  But either do one or the other

     TorchSharp.DeviceType
     TorchSharp.Device
     TorchSharp.Tensor
     TorchSharp.Scalar

or

     TorchSharp.TorchDeviceType
     TorchSharp.TorchDevice
     TorchSharp.TorchTensor
     TorchSharp.TorchScalar

NiklasGustafsson:
I agree with the naming consistency. Which is your preference -- 'Torch*' or just '*'?

PR #229, opened 2021-03-26
More NN layers implemented

NiklasGustafsson:
I added a number of normalization layers, as well as the padding layers in torch.nn.


NiklasGustafsson:
The Ubuntu build is failing because there are a number of unit tests that rely on training resulting in monotonically decreasing losses as the training loop goes on.

dsyme (approved):

dsyme:
One test failure
```
  Failed TorchSharp.TestTraining.TestTrainingWithDropout [10 ms]
  Error Message:
   Assert.True() Failure
Expected: True
Actual:   False
  Stack Trace:
     at TorchSharp.TestTraining.TestTrainingWithDropout() in /__w/1/s/test/TorchSharpTest/TestTraining.cs:line 81
```

NiklasGustafsson:
@dsyme -- Yes, it is this piece of code, in the middle of a training loop:

```
                Assert.True(lossVal < prevLoss);
                prevLoss = lossVal;
```
While it is generally the case that the loss will decrease over time, there is no guarantee that it will decrease in every single step. Therefore, the unit test assertion is volatile and depends on the random numbers used to generate both the test data and the initial weights of the various layers.

I'm going to see if setting the seed fixes it, but I don't think it will -- there is a single default RNG, and since unit tests are run in parallel, that could lead to its own issues. What we need to do is add explicit RNGs and use one per training unit test, setting the seed to something deterministic. However, that will take some time.

The test is useful, so I don't want to disable it.

NiklasGustafsson:
Hold off merging this, I've changed the unit tests to be less volatile.

dsyme:
> I'm going to see if setting the seed fixes it, but I don't think it will -- there is a single default RNG, and since unit tests are run in parallel, that could lead to its own issues. What we need to do is add explicit RNGs and use one per training unit test, setting the seed to something deterministic. However, that will take some time.

OK cool

PR #230, opened 2021-03-26
Random

NiklasGustafsson:
Several APIs take a random number Generator object as input. In order to be able to provide them, the Generator class itself has to be implemented. This PR adds it, as 'TorchGenerator,' and allows the seed to be set, state to be snapshot and restored, etc. Currently, only CPU-based RNGs are supported.

### DIFF
--- a/src/Native/LibTorchSharp/THSTorch.cpp
+++ b/src/Native/LibTorchSharp/THSTorch.cpp
@@ -44,13 +44,13 @@ void  THSGenerator_set_rng_state(const Generator gen, const Tensor tensor)
 
 Generator THSGenerator_new(uint64_t seed, int64_t device, int64_t index)
 {
+    // TODO: Support creation of GPU RNGs. 'device' and 'index' are in the
+    //       function signature in preparation thereof.
     return new at::Generator(at::detail::createCPUGenerator(seed));
 }
 
 void THSGenerator_dispose(const Generator generator)
 {
-    auto defi = at::globalContext().defaultGenerator(at::DeviceType::CPU).getIntrusivePtr();
-    auto geni = generator->getIntrusivePtr();
     delete generator;
 }
 
--- a/src/Native/LibTorchSharp/THSTorch.cpp
+++ b/src/Native/LibTorchSharp/THSTorch.cpp
@@ -9,6 +9,35 @@ void THSTorch_manual_seed(const int64_t seed)
     torch::manual_seed(seed);
 }
 
+Generator THSGenerator_manual_seed(const int64_t seed)
+{
+    torch::manual_seed(seed);
+    return THSGenerator_default_generator();
+}
+
+Generator THSGenerator_default_generator()
+{
+    auto gen = at::globalContext().defaultGenerator(at::DeviceType::CPU);
+    return new at::Generator(gen.getIntrusivePtr());
+}
+
+int64_t THSGenerator_initial_seed(const Generator gen)
+{
+    return gen->current_seed();
+}
+
+Generator THSGenerator_new(int64_t device, int64_t index)
+{
+    return new torch::Generator(c10::make_intrusive<torch::CPUGeneratorImpl>());
+}
+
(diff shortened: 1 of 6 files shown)


src/Native/LibTorchSharp/THSTorch.cpp
```diff
@@ -9,6 +9,51 @@ void THSTorch_manual_seed(const int64_t seed)
...
+
+Generator THSGenerator_new(uint64_t seed, int64_t device, int64_t index)
+{
+    return new at::Generator(at::detail::createCPUGenerator(seed));
+}
+
+void THSGenerator_dispose(const Generator generator)
+{
+    auto defi = at::globalContext().defaultGenerator(at::DeviceType::CPU).getIntrusivePtr();
+    auto geni = generator->getIntrusivePtr();
```
dsyme, line 53: What's the purposes of `geni`, unused?

src/Native/LibTorchSharp/THSTorch.cpp
```diff
@@ -9,6 +9,51 @@ void THSTorch_manual_seed(const int64_t seed)
...
+    CATCH_TENSOR(gen->get_state());
+}
+
+void  THSGenerator_set_rng_state(const Generator gen, const Tensor tensor)
+{
+    gen->set_state(*tensor);
+}
+
+
+Generator THSGenerator_new(uint64_t seed, int64_t device, int64_t index)
```
dsyme, line 45: `index` is unused here

src/Native/LibTorchSharp/Utils.h
```diff
@@ -10,6 +10,8 @@ extern thread_local char *torch_last_err;
 
 typedef torch::Tensor *Tensor;
 typedef torch::Scalar *Scalar;
+typedef torch::Generator* Generator;
```
dsyme, line 13: So it would be good to write some notes on when we use `shared_ptr` reference counting and why.

src/TorchSharp/TorchGenerator.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        [DllImport("LibTorchSharp")]
+        extern static void THSGenerator_dispose(IntPtr handle);
+
+        #region Dispose() support
+        protected virtual void Dispose(bool disposing)
+        {
+            if (Handle != IntPtr.Zero) {
+                THSGenerator_dispose(Handle);
+                Handle = IntPtr.Zero;
```
dsyme, line 96: Set Handle to Zero first? In case `THSGenerator_dispose` raises an exception or something similar?

dsyme:
Asked a few questions - but basically the code looks fine

dsyme (approved):

src/Native/LibTorchSharp/THSTorch.cpp
NiklasGustafsson: Oops! Debugging leftovers.

src/Native/LibTorchSharp/THSTorch.cpp
NiklasGustafsson: Added comments in source file

src/Native/LibTorchSharp/Utils.h
NiklasGustafsson: That's a fair question. When I put this in place, I followed the Tensor pattern rather than the Module / Optimizer pattern. The generators are already implemented using their own proprietary reference counting scheme, so it seemed like adding a layer of reference counting on top of an already existing one. Redundant.

Happy to be convinced that was misguided.

src/TorchSharp/TorchGenerator.cs
NiklasGustafsson: Excellent point. My next push will fix this.

NiklasGustafsson:
I have addressed the PR comments in the new push.

dsyme:
> I have addressed the PR comments in the new push.


👍

PR #235, opened 2021-04-02
Examples

NiklasGustafsson:
Fixed up the MNIST and AlexNet examples, adding .NET-implemented dataset readers for MNIST (and Fashion MNIST) as well as CIFAR10.
Corrected an issue regarding the sequencing of SetSeed() and IsCudaAvailable() -- setting the seed first disabled CUDA use.
Added a couple of loss functions and unit tests.

### DIFF
--- a/src/Examples/BigEndianReader.cs
+++ b/src/Examples/BigEndianReader.cs
@@ -1,0 +1,43 @@
+using System;
+using System.IO;
+
+namespace TorchSharp.Examples.Utils
+{
+    public class BigEndianReader
+    {
+        public BigEndianReader(BinaryReader baseReader)
+        {
+            mBaseReader = baseReader;
+        }
+
+        public int ReadInt32()
+        {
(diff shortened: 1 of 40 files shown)


src/Examples/BigEndianReader.cs
```diff
@@ -0,0 +1,43 @@
+using System;
```
dsyme, line 1: Tbh we should likely move all typical datasets into a new package TorchSharp.Data, much as we do in DiffSharp

Can be done in separate PR

Would also like to see most samples done using notebooks, which are pretty great to use in VS Code

dsyme (approved):

src/Examples/BigEndianReader.cs
NiklasGustafsson: Setting up a separate TorchSharp.Data package seems like a good idea.

NiklasGustafsson:
I'll take care of merging once it's done with its checks. I just added a comment, so it will restart the tests.

PR #242, opened 2021-04-07
Adding the Vision modules and functions.

NiklasGustafsson:


src/Native/LibTorchSharp/THSNN.cpp
```diff
@@ -986,6 +1020,185 @@ Tensor THSNN_FeatureAlphaDropout_forward(const NNModule module, const Tensor ten
...
+}
+
+template<typename T>
+void ApplyGridMode(T& opts, const int8_t mode)
+{
+    if (mode == 0)
+        opts = opts.mode(torch::kNearest);
+    if (mode == 2)
+        opts = opts.mode(torch::kBilinear);
+    //if (mode == 3)
```
dsyme, line 1133: what's up here?

dsyme (approved):

src/Native/LibTorchSharp/THSNN.cpp
NiklasGustafsson: With the commented line? The PyTorch docs say that it should be supported, but the C++ variant<> type declaration doesn't allow it. I left it in for future use.

NiklasGustafsson: I'll put a comment in the code to clarify.

PR #253, opened 2021-05-06
Save / load of models, and an example using it

NiklasGustafsson:
In this PR, I added a .NET-specific way to save and restore the weights of a model. This also includes a Python script demonstrating how to save weights in a way that allows .NET to load them.

I fixed the MNIST and CIFAR readers, which both managed to skip the first batch of input data.

I also updated the document on memory management strategy with some more details on the implications of relying on GC for  e training on GPUs.

### DIFF
--- a/docfx/index.md
+++ b/docfx/index.md
@@ -11,6 +11,12 @@ This surfaces the C API as a strongly-typed C# API.
 Check the [GitHub project page](https://github.com/xamarin/TorchSharp) for
 TorchSharp.
 
+## Tips and Tricks
+
+[Some thoughts](articles/memory.md) on how to manage memory well in your TorchSharp training code.
+
+An intruction on how to [share model](articles/saveload.md) weights between applications, whether in Python or .NET.
+
 ## API documentation
 
 The [API Documentation](api/TorchSharp.html)
--- a/src/Examples/MNIST.cs
+++ b/src/Examples/MNIST.cs
@@ -50,7 +50,6 @@ namespace TorchSharp.Examples
 
             var cwd = Environment.CurrentDirectory;
 
-            //var device = Device.CPU;
             var device = Torch.IsCudaAvailable() ? Device.CUDA : Device.CPU;
             Console.WriteLine($"Running MNIST on {device.Type.ToString()}");
             Console.WriteLine($"Dataset: {dataset}");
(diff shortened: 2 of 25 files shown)


docfx/index.md
```diff
@@ -11,6 +11,12 @@ This surfaces the C API as a strongly-typed C# API.
 Check the [GitHub project page](https://github.com/xamarin/TorchSharp) for
 TorchSharp.
 
+## Tips and Tricks
+
+[Some thoughts](articles/memory.md) on how to manage memory well in your TorchSharp training code.
+
+An intruction on how to [share model](articles/saveload.md) weights between applications, whether in Python or .NET.
```
dsyme, line 18: typo

src/Examples/MNIST.cs
```diff
@@ -70,120 +65,107 @@ static void Main(string[] args)
...
         }
 
-        private class Model : CustomModule
+        internal class Model : CustomModule
         {
-            private Conv2d conv1 = Conv2d(1, 10, 5);
-            private Conv2d conv2 = Conv2d(10, 20, 5);
-            private Linear fc1 = Linear(320, 50);
-            private Linear fc2 = Linear(50, 10);
+            private Conv2d conv1 = Conv2d(1, 32, 3);
```
dsyme, line 114: I think examples could use named parameters for things like this, it reads much more clearly and teaches

test/TorchSharpTest/TestLoadSave.cs
```diff
@@ -15,26 +15,34 @@ namespace TorchSharp
...
             var loadedLinear = NN.Linear.Load(".model.ts");
+            var params1 = loadedLinear.parameters();
             File.Delete(".model.ts");
+
             Assert.NotNull(loadedLinear);
+            Assert.Equal(params0, params1);
         }
 
-        [Fact]
+        [Fact(Skip = "https://github.com/xamarin/TorchSharp/issues/251")]
```
dsyme, line 33: Should these now work, and no longer be skipped?

dsyme (approved):

test/TorchSharpTest/TestLoadSave.cs
NiklasGustafsson: So, I'll fix that.

dsyme:
I released 0.91.52573 which contains this and all other work in master

NiklasGustafsson:
@dsyme, maybe you can show me how to do a release, some time? There some features in the subsequent PR that I would like to have released, and it seems like the perfect reason for me to learn how.

dsyme:
@NiklasGustafsson Documented here, I think you have the permissions: https://github.com/xamarin/TorchSharp/blob/master/DEVGUIDE.md#doing-releases-of-the-torchsharp-package

It's pretty simple and can be done for any successful build of master

pkese:
Could it be that there's something wrong with this build or with what got published to Nuget?

I have tried to use `copy_`, which should supposedly be included in this version, but it fails saying that the method doesn't exist.

```fsharp
#r "nuget: libtorch-cpu, 1.8.0.7"
#r "nuget: TorchSharp, 0.91.52573"

open TorchSharp
open TorchSharp.Tensor

let first = Float32Tensor.rand [|1L;2L;3L|]
let second = Float32Tensor.zeros [|1L;2L;3L|]

second.copy_(first);

assert(second = first)
```

dsyme:
Release queued, it should appear in 5-10 minutes

pkese:
Now it works. Thanks.

PR #255, opened 2021-05-14
F# Examples

NiklasGustafsson:
F# examples corresponding to each of the C#-based examples.

This is a draft, submitted to solicit stylistic comments.


src/FSharp Examples/SequenceToSequence.fs
```diff
@@ -0,0 +1,273 @@
...
+        let NULL = System.Nullable<int64>()
+        pe.[TorchTensorIndex.Ellipsis, TorchTensorIndex.Slice(0L, NULL, 2L)] <- (position * divTerm).sin()
+        pe.[TorchTensorIndex.Ellipsis, TorchTensorIndex.Slice(1L, NULL, 2L)] <- (position * divTerm).cos()
+
+        pe <- pe.unsqueeze(0L).transpose(0L,1L)
+
+        this.RegisterComponents()
+
+    override _.forward(t) =
+        let NULL = System.Nullable<int64>()
```
cartermp, line 73: You can actually just use `System.Nullable()` here, as type inference will figure out the specific type. `null` cannot be passed in directly, however.

src/FSharp Examples/SequenceToSequence.fs
```diff
@@ -0,0 +1,273 @@
...
+        if (batch % logInterval = 0) && (batch > 0) then
+            let cur_loss = (total_loss / (float32 logInterval)).ToString("0.00")
+            printfn $"epoch: {epoch} | batch: {batch} / {tdlen/bptt} | loss: {cur_loss}"
+            total_loss <- 0.0f
+
+        batch <- batch + 1
+        i <- i + bptt
+
+
+let evaluate (model:TransformerModel) (evalData:TorchTensor) ntokens =
```
cartermp, line 182: The mutables + `while` loop in here might be easily translatable into a recursive inner function. That'd be more idiomatic F#, but might not seem as friendly to Python users. What do you think?

NiklasGustafsson: It would definitely not be Pythonic -- don't know whether that's what we should go for, though.

src/FSharp Examples/AdversarialExampleGeneration.fs
```diff
@@ -0,0 +1,134 @@
...
+
+let test (model:MNIST.Model) (eps:float) (dataLoader:MNISTReader) size =
+
+    let mutable correct = 0
+
+    for (input,labels) in dataLoader do
+
+        input.requires_grad <- true
+        
+        begin  // This is introduced in order to let a few tensors go out of scope before GC
```
pkese, line 74: Is this idiomatic F#?  
This is not a complaint ... I was facing a similar situation and was wondering what's an optimal syntax in such case.  
Would a `do` block work here as well?  
Maybe a question for @cartermp?

pkese:
If the assembly name is `FSharp.Examples` maybe the directory name could be the same instead of `FSharp Examples` (with a blank).
Also spaces in file and folder names are sometimes unpleasant to work with.

src/FSharp Examples/AdversarialExampleGeneration.fs
dsyme: It's not particularly idomatic but I think nothing is - it's a rare situation.  A 'do' would be ok as well, as would parentheses

src/FSharp Examples/AlexNet.fs
```diff
@@ -0,0 +1,194 @@
...
+open System.Diagnostics
+
+open TorchSharp
+open TorchSharp.Tensor
+open TorchSharp.NN
+
+open type TorchSharp.NN.Modules
+open type TorchSharp.TorchScalar
+
+/// <summary>
```
dsyme, line 14: I think remove the XML docs here and elsewhere. Just use `//` docs -  or else put the XML docs on the outer module for the whol file (the `module` currently on the first line.  

As you have it the XML docs are attaching to the first 'let' that follows the docs, which is `trainBatchSize` in this case

dsyme (approved):
I left a couple of minor comments.  On the whole it looks excellent, very regular, very pythonic, easy to read.

src/FSharp Examples/AdversarialExampleGeneration.fs
NiklasGustafsson: Yes, the need to clean up GPU memory ASAP does introduce some non-attractive code here and there. Same thing in C#, with lots of 'usings' that are there for no other reason.

src/FSharp Examples/AlexNet.fs
NiklasGustafsson: Will do.

src/FSharp Examples/AdversarialExampleGeneration.fs
cartermp: Another alternative could be the `using` function:

```fsharp
using (input --> model) (fun estimate ->
	use loss = criterion estimate labels

	model.ZeroGrad()
	loss.backward()

	use perturned = attack input (eps.ToScalar()) (input.grad())
	use final = perturbed --> model
	correct <- correct + final.argmax(1L).eq(labels).sum().ToInt32()
)
```

I believe that this can achieve the same thing without an explicitly scoped block. The downside of course is that we're mixing two kinds of "bind a disposable resource" approaches.

NiklasGustafsson: This seems like a step in the wrong direction, IMO -- greater syntactic overhead.

NiklasGustafsson:
Okay, since Don approved, and I have addressed most of the comments, I'll merge this now.

PR #258, opened 2021-05-19
Adding miscellaneous missing operators.

NiklasGustafsson:
There was a number of missing tensor factory methods, as well as some more obscure operators, such as the spectral window operators.

The set of random number distributions should be fairly complete right now.

### DIFF
--- a/src/Native/LibTorchSharp/THSTensor.cpp
+++ b/src/Native/LibTorchSharp/THSTensor.cpp
@@ -1895,9 +1895,9 @@ void THSTensor_min_along_dimension(const Tensor tensor, Tensor* (*allocator)(siz
 {
     CATCH(
         auto max = tensor->min(dim, keepdim);
-    Tensor * result = allocator(2);
-    result[0] = new torch::Tensor(std::get<0>(max));
-    result[1] = new torch::Tensor(std::get<1>(max));
+        Tensor * result = allocator(2);
+        result[0] = new torch::Tensor(std::get<0>(max));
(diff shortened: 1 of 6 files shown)


src/Native/LibTorchSharp/THSTensor.cpp
```diff
@@ -3121,6 +3359,61 @@ Tensor THSInit_eye_(Tensor tensor)
     CATCH_TENSOR(torch::nn::init::eye_(*tensor))
 }
 
+Tensor THSTensor_bernoulli(const Tensor tensor, const double p, const Generator gen)
```
dsyme, line 3362: Just to mention I'd been ordering the file end-to-end alphabetical, rather than by kind (distributions etc.)

I don't really mind, I just figured alphabetical was the only thing which didn't requrie a lot of thought :-)

NiklasGustafsson: Yeah, but that had already started to break down a bit (my fault). I see value in having things that are related to each other close, but the real way to do that would be to break it into multiple .cpp files, not have one gigantic file. Maybe I should do that next...

NiklasGustafsson: I have some changes that separates the C++ code into multiple files, but I'll put that in a different PR, since it makes it much harder to compare and review.

NiklasGustafsson:
I will go ahead and merge the PR now.

dsyme:
Yes, it's fine, I meant to say "approved"

PR #263, opened 2021-05-22
Complex number support

NiklasGustafsson:
This is a first draft PR for complex number support. At this point, you can create, do arithmetic, and convert complex tensors.

The really interesting work will be FFTs for signal processing, but I will add that in a later PR. This is just the basic stuff.


dsyme (approved):

dsyme:
This looks great

PR #271, opened 2021-06-15
[WIP] add mac build 

dsyme:
Kind of seems to build on my local machine, passes all tests first time.

No MKL downloaded needed so far


dsyme:
@NiklasGustafsson I'm pretty optimistic about this after seeing the tests pass so easily.

After we merge this we will need to check the native build and the packages. The CUDA native builds aren't done on every PR and are only done on master due to the large downloads.  The packages are only built and pushed on master.

NiklasGustafsson:
@dsyme -- let me check it out on my MacBook.

NiklasGustafsson:
Here's what I see early in the build: 

```
/Users/niklasg/Repos/dsyme/TorchSharp/src/Redist/libtorch-cuda-11.1/libtorch-cuda-11.1.proj(137,5): error MSB3923: Failed to download file "https://download.pytorch.org/libtorch/cu111/libtorch-macos-1.8.0%2Bcu111.zip".  Response status code does not indicate success: 403 (Forbidden).
```

NiklasGustafsson:
@dsyme -- the problem is that I have to say 'dotnet build /p:SkipCuda=true'

Should we make that the default on Mac, so that we don't have to provide the extra parameter?

NiklasGustafsson:
Okay, it builds and runs on my Mac. I'm going to approve and merge this PR.

NiklasGustafsson:
@dsyme -- after merging, the MacOS Release build fails during tests.

PR #278, opened 2021-06-16
Test PR

dsyme:
Ubuntu builds have randomly started failing, checking this 

```
E: Unable to locate package dotnet-sdk-5.0
E: Couldn't find any package by glob 'dotnet-sdk-5.0'
E: Couldn't find any package by regex 'dotnet-sdk-5.0'
```

https://donsyme.visualstudio.com/TorchSharp/_build/results?buildId=452&view=logs&jobId=80b813b5-9a08-5859-11a8-dc0e5b556e52&j=80b813b5-9a08-5859-11a8-dc0e5b556e52&t=99848337-6ccc-53eb-9c14-1b676ae001b9

PR #279, opened 2021-06-16
Use ubuntu-latest to hold ubuntu 16-04 containers

dsyme:
We do our linux builds on Ubuntu 16.04 containers to get Linux binaries that have very low dependency levels, hence usable in environments like Google Colab notebooks.

Ubuntu 16.04 vmImages are being retired, however the 16.04 containers we use should be able to run on `ubuntu-latest`


dsyme:
Hitting the same problem as #278

PR #280, opened 2021-06-18
allow splitting files size > int32 max

dsyme:
So LibTorch now contains single binaries with size > Int32.MaxValue.  😱 😱 😱 😱

### DIFF
--- a/src/Redist/libtorch-cuda-11.1/libtorch-cuda-11.1.proj
+++ b/src/Redist/libtorch-cuda-11.1/libtorch-cuda-11.1.proj
@@ -189,8 +189,8 @@
     <ParameterGroup>
       <SourceFile ParameterType="System.String" Required="true" />
       <DestinationFolder ParameterType="System.String" Required="true" />
-      <Start ParameterType="System.Int32" Required="true" />
-      <Size ParameterType="System.Int32" Required="true" />
+      <Start ParameterType="System.Int64" Required="true" />
+      <Size ParameterType="System.Int64" Required="true" />
       <Index ParameterType="System.String" Required="true" />
     </ParameterGroup>
     <Task>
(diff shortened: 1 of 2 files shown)


lostmsu:
I am guessing you already know that does not really work, because .NET Core and .NET Framework do not support arrays > 4GB. :(

https://github.com/dotnet/runtime/issues/12221

lostmsu:
Actually, the supported (and probably in this case better) way would be to https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.copytoasync or its synchronous variant.

src/Redist/libtorch-cuda-11.1/libtorch-cuda-11.1.proj
```diff
@@ -217,7 +217,7 @@
                     var outBytes = new byte[Size];
                     using (var accessor = mmf.CreateViewAccessor(Start, Size))
                     {
-                        accessor.ReadArray<byte>(0, outBytes, 0, Size);
+                        accessor.ReadArray<byte>(0, outBytes, 0, (int)Size);
```
lostmsu, line 220: This cast is going to silently lose data: `i64` 0x1_0000_0001 (4GB+1byte) will be turned into just 0x1 `i32`

dsyme: It's ok as size is always < 250MB, it could have been kept as int32

dsyme:
> I am guessing you already know that does not really work, because .NET Core and .NET Framework do not support arrays > 4GB. :(


For now it's ok as we need the 2GB-4GB range

Above 4GB we will indeed need to stop using arrays

dsyme:
@lostmsu Thanks for checking, and highlighting what will happen soon given the crazy rate of size increase of the libtorch binaries.  Seriously, 3GB binaries increasing about 30% every release??

Issue #225, opened 2021-03-25
Resolve some naming questions


dsyme:
1. I'm not particularly tied to `open TorchSharp.Tensor` namespace.  I'd be happy with `open TorchSharp` or anything that gives the ergonomics you want

2. When updating DiffSharp to latest I noticed `TorchSharp.Device` rather than `TorchSharp.TorchDevice`.  Previously I was putting `Torch` on the front to help disambiguate w.r.t. DiffSharp "Tensor".  But I understand you'll likely need to switch to plain `Tensor`.  But either do one or the other

        TorchSharp.DeviceType
        TorchSharp.Device
        TorchSharp.Tensor
        TorchSharp.Scalar

or

        TorchSharp.TorchDeviceType
        TorchSharp.TorchDevice
        TorchSharp.TorchTensor
        TorchSharp.TorchScalar

Right now we're a mix of these.

NiklasGustafsson:
My preference would be to remove the 'Torch' prefix -- it is essentially a redundant encoding of the namespace in type names.

NiklasGustafsson:
How about:

Torch.DeviceType
Torch.Device
Torch.Tensor
Torch.Scalar

dsyme:
This is ok by me.

NiklasGustafsson:
So, we change 'TorchSharp' to 'Torch' and remove the prefix on all types where it differs from PyTorch? 

Do we capitalize following traditional .NET conventions, or do we try to stay as close to PyTorch as possible? Tensorflow.NET has made a consistent effort to stay close to PyTorch, which I believe is a SciSharp design principle. Do we do the same here?

dsyme:
> So, we change 'TorchSharp' to 'Torch' and remove the prefix on all types where it differs from PyTorch?

Actually TBH I'd probably prefer "TorchSharp" since that's the name of the tech.  

> Do we capitalize following traditional .NET conventions, or do we try to stay as close to PyTorch as possible? Tensorflow.NET has made a consistent effort to stay close to PyTorch, which I believe is a SciSharp design principle. Do we do the same here?

We stay as close as possible to PyTorch.  But for type names we use capitalization.  See https://github.com/xamarin/TorchSharp/issues/226 for the broader question of how to get closer to PyTorch through a lower-case `torch` type.    Indeed for this reason I think it's better to keep `open TorchSharp` as the very top namespace so that approach wouldn't lead to confusion

NiklasGustafsson:
[...]

PR #302, opened 2021-07-02
MASSIVE BREAKING CHANGE: Renaming things

NiklasGustafsson:
This PR will align TorchSharp with the naming convention and API placement more closely with Pytorch. This was discussed in issue #225. 

Effectively, this means that with a 'using TorchSharp;' in C#, or a 'open TorchSharp' in F#, you should be able to find methods and classes in the same place that they are found in PT.

This is a draft PR, more work is necessary. I'm creating it in order to get feedback. Almost every file in the TorchSharp library was changed, and I don't expect a fine-tooth-comb review, just a scan of a few representative files for feedback.

This PR also includes the changes in #301.


GeorgeS2019:
@NiklasGustafsson 

I checkout this pull request and tried the ToString examples
Now the statement is more compact!

image

NiklasGustafsson:
```csharp
var t0 = torch.zeros(4);
```
is even more compact... :-)

src/Examples.Utils/AG_NEWSReader.cs
```diff
@@ -3,41 +3,41 @@
...
     {
-        public static AG_NEWSReader AG_NEWS(string split, Device device, string root = ".data")
+        public static AG_NEWSReader AG_NEWS(string split, device device, string root = ".data")
         {
             var dataPath = Path.Combine(root, $"{split}.csv");
             return new AG_NEWSReader(dataPath, device);
         }
 
-        private AG_NEWSReader(string path, Device device)
+        private AG_NEWSReader(string path, device device)
```
dsyme, line 18: Regarding the lower case `device` type - I was expecting type names would stay uppercase?

src/Examples/SequenceToSequence.cs
```diff
@@ -171,43 +167,43 @@ private static double evaluate(TorchTensor eval_data, TransformerModel model, Lo
...
-            var target = source[TorchTensorIndex.Slice(index + 1, index + 1 + len)].reshape(-1);
+            var data = source[TensorIndex.Slice(index, index + len)];
+            var target = source[TensorIndex.Slice(index + 1, index + 1 + len)].reshape(-1);
             return (data, target);
         }
 
         class TransformerModel : CustomModule
         {
-            private TransformerEncoder transformer_encoder;
+            private impl.TransformerEncoder transformer_encoder;
```
dsyme, line 200: What's `impl`?

dsyme (approved):

src/Examples.Utils/AG_NEWSReader.cs
NiklasGustafsson: Yes, because Python class names are usually Capitalized. However, this specific type is all lowercase in Pytorch, and I went with the same in TorchSharp, for inter-language consistency.

src/Examples/SequenceToSequence.cs
NiklasGustafsson: Well... This is one point that needs some discussion.

Generally, in order to mimic Python's new-free constructor syntax for creating torch modules, TorchSharp used to have a special static class 'Modules' in with nothing but factories, all having the same name as the module class:

```csharp
namespace TorchSharp.NN

  public class Conv1d { ... }

  public static partial class Modules { public Conv1d Conv1d(...) { ... } }

}
```

When attempting to move as close to Pytorch as possible, you would want the class and the factory accessible as `torch.conv1d`:

```csharp
namespace TorchSharp

  public static partial class Modules { 

    public class Conv1d { ... }

    public Conv1d Conv1d(...) { ... } 
  }
}
```

but that's not legal. So, there's a choice to be had between placing the factory method in a nested scope, or the type, or forgo the factory altogether and make users say 'new Conv1d()' when creating a module in C#.

Since you rarely __have__ to use the class name, I thought it would make sense to put that in a nested scope rather than the factory. I could certainly be persuaded that relying on C# constructor syntax is just fine, but that does take us away from the "as close to Pytorch as possible" principle.

In most situations, you can infer the type, or just rely on 'Module' as the type name, but in models where the forward method use for a module isn't the standard (and overridden) 'forward(Tensor input)' but something else, you need to capture the most-derived class name in the field declaration. That should be fairly rare.

src/Examples.Utils/AG_NEWSReader.cs
dsyme: Hmmmm.  I think for this one I'd go uppercase, with a lower-case `device` value with instance members to allow `torch.device.CPU`.   I just can't quite handle seeing `device device` in C# code :-)

dsyme:
(When reviewing I focused on the F# and C# samples, I think they are representative enough)

src/Examples/SequenceToSequence.cs
GeorgeS2019: @NiklasGustafsson @dsyme 
Another reason why I am supporting pytorch like naming for TorchSharp is to eventually use a **Python2Csharp** converter to speed up the porting of Pytorch to TorchSharp with a list of Rosyln rules to assist the conversion. 

A successful conversion is one which can be compiled and then using additional Rosyln rules to determine if the results are as expected from the pyTorch through test driven development (TDD). 

What ever final naming or layout of codes we finally reach a consensus, I hope, will make writing the Rosyln rules easier and the TDD simpler.

GeorgeS2019:
@dsyme @NiklasGustafsson 
> (When reviewing I focused on the F# and C# samples, I think they are representative enough)

I did a quick glance over the [Pytoch Recipes chapter 2 jupyter ](https://github.com/Apress/pytorch-recipes/blob/master/Torch_AI_2.ipynb), which covers:

## Chapter 2: Probability Distributions Using PyTorch
- Tensor sampling
- Tensor variables
- Tensor statistics
- Tensor gradient computation
- Tensor operations
- Tensor distributions

I personally find it not often intuitive to port a few of these codes using the latest TorchSharp breaking change renaming. 

If both of you see values in looking at these tensor topics in pytorch for renaming consistency, it would be great.

e.g. I do not see sufficient TorchSharp unit tests coverage on the topic [Tensor sampling: torch.bernoulli and torch.multinomial](https://github.com/xamarin/TorchSharp/issues/303), especially with the consistent renaming.  [Perhaps they are there, and I could have miss them ]

dsyme:
> I personally find it not often intuitive to port a few of these codes using the latest TorchSharp breaking change renaming.
> 
> If both of you see values in looking at these tensor topics in pytorch for renaming consistency, it would be great.
> 
> e.g. I do not see sufficient TorchSharp unit tests coverage on the topic Tensor sampling: torch.bernoulli and torch.multinomial, especially with the consistent renaming. [Perhaps they are there, and I could have miss them ]

Could you write out the issues you encountered with porting more systematically?   Thanks.

src/Examples/SequenceToSequence.cs
dsyme: Could you make it `Module.TransformerEncoder` or `Modules.TransformerEncoder` or `ModuleTypes.TransformEncoder`?  At least something that doesn't feel like it's leaking implementation internals (which is what `impl` says to me)

NiklasGustafsson: I like `Modules.TransformerEncoder`

dsyme: Cool

lostmsu:
@NiklasGustafsson is this by design, that `CustomModule` is a nested class in `torch.nn`, but class `Sequential` is located directly in `TorchSharp.Modules`? Seems very inconsistent.

NiklasGustafsson:
@lostmsu

Don't recall off the top of my head. I'll look at it when I'm back working tomorrow. If you find inconsistencies, which are not at all hard to imagine given the immense API restructure, please file issues. If you have the time, that is.

Having a critical eye take a look at the API is truly valued.

NiklasGustafsson:
@lostmsu 

After reminding myself about it, this inconsistency is by design. The 'Modules' is a namespace for implementation details that need to be public.

The design principle for the reworked API is to make moving code from Python to C#/F# as easy as possible. As discussed elsewhere, this is because we have reams and reams of texts on PyTorch that we simply cannot duplicate for .NET at a reasonable expense, nor can we track the new texts as they come out.

Thus, for every symbol (and we may have missed a few), the idea is to make using it the way it is used in Python code as easy as possible. For Sequential, that means having the factory method in the torch.nn "module," since that is how Sequential is most often used -- to construct a new sequential module.

For CustomModule, the most common use is to derive a new class from it, so the type name should be in the 'torch.nn' module.

That's the design rationale. It's reasonable to disagree with it, but there was thought behind it.

GeorgeS2019:
@lostmsu your critical eye and valuable feedback plus @NiklasGustafsson detailed explanation helps others like myself to appreciate further the challenge in making TorchSharp works for a bigger purpose. Hope more will join in to  keep refining the TorchSharp implementation.

lostmsu:
@NiklasGustafsson re: `CustomModule`, I just find it a bit weird considering `CustomModule` is not present in PyTorch. I also went through its code, and it seems that for the sake of simplicity we could simply merge the entire contents of `CustomModule` into `Module` class at the small expense of having unused delegate pointer for regular modules. That way you'd simply inherit from `Module` when you make your own custom one.

NiklasGustafsson:
Yes, that is worth considering.

gbaydin:
@NiklasGustafsson could you please tell me which nuget package version of TorchSharp introduces these naming changes? Is it 0.91.52719? I couldn't find this information.

GeorgeS2019:
@gbaydin the naming change started (2nd July) less than 2 months ago (0.91.52681). 0.91.52719 (15 days old) should include the naming change.

PR #313, opened 2021-07-19
More Distributions

NiklasGustafsson:
Adding the torch.distributions package, and about half of the implementations.


dsyme (approved):

dsyme:
It's a big change - I took a look through all of it and didn't spot any problems

PR #317, opened 2021-07-23
Fixes for double-freeing native tensor handles.

NiklasGustafsson:
This PR fixes three instances of native tensor handles being freed more than once, causing access violations.

It includes the fix surfaced in PR #316.


NiklasGustafsson:
@lostmsu, I tried to add you as a reviewer, but your name didn't pop up on the list.

lostmsu (approved):
Looks good to me

lostmsu:
Do your tests for the custom module without the fix? I wonder why you had to replace the original one.

NiklasGustafsson:
Yeah, I copied your test to debug it and understand exactly how it went wrong, then I copied your fix, etc.

Issue #321, opened 2021-08-09
A callback was made on a garbage collected delegate `CustomModule+ForwardFunctionC`


lostmsu:
Process terminated. A callback was made on a garbage collected delegate of type `TorchSharp!TorchSharp.torch+nn+CustomModule+ForwardFunctionC::Invoke`.

This happens, because one does not generally has a live reference to an instance of `CustomModule`. For example, when it is part of `Sequential`.

Because the class instance has to be preserved, it needs to allocate a `GCHandle`, that C++ side of the code must hold on to until torch has no more references to the module.

NiklasGustafsson:
@lostmsu -- thanks for reporting. 

If you come up with a small repro case, please let me know.

NiklasGustafsson:
@lostmsu -- I believe a fix to this bug is for the .NET Sequential implementation to hold on to the submodules list that it is given, rather than just handing the native handles to LibTorch.

lostmsu:
@NiklasGustafsson I believe this does the trick:

```csharp
class TestModule : CustomModule {
    public TestModule() : base(nameof(TestModule)) {}

    public override torch.Tensor forward(torch.Tensor t) => t;

    static void Main() {
        var seq = Make();
        GC.Collect();
        GC.WaitForPendingFinalizers();
        GC.Collect();
        seq.forward(torch.zeros(10));
    }

    [MethodImpl(MethodImplOptions.NoInlining)]
    static Module Make() => Sequential(("t", new TestModule()), ("d", Linear(10, 1)));
}
```

Though this just generates `System.ExecutionEngineException`. So might be an entirely separate issue :-D

lostmsu:
@NiklasGustafsson I think trying to fix `Sequential` is going to be error prone: if there are more composite modules in the future, their developers won't know about the need to keep module instances alive in .NET heap.

`GCHandle` is the correct solution here.

NiklasGustafsson:
Ok. I'll take a look at this tomorrow.

NiklasGustafsson:
@lostmsu -- the repro allowed me to successfully reproduce the bug. Thank you, that made my life much easier. I have incorporated the code into the unit tests.

I see the ExecutionEngineException, but I'm convinced that it's the same problem, since the root cause is that the .NET module reference is GC'd while the native handle still lives on and gets used. The delegate is also GC'd, like you say in the issue title, since the only reference to it is in the module instance. 

[...]

Issue #318, opened 2021-07-25
A few inconsistencies with the new naming


lostmsu:
1. I believe the entire contents of `torch.functional` are misplaced.
2. `CustomModule` is a nested class inside `torch.nn`, but `Sequential` is directly in `TorchSharp.Modules` (personally prefer the second variant).

NiklasGustafsson:
@lostmsu -- thanks! We've discussed #2 elsewhere, but #1 seems like something that needs investigation. I'll look into it today.

lostmsu:
I guess n1 was not too detailed, but the problem is existence of both `torch.functional` and `torch.nn.functional`.

NiklasGustafsson:
@lostmsu -- yup, that whole namespace shouldn't exist. It was created by sloppy placement of '}' characters in a number of files. PR #319 addresses it.

NiklasGustafsson:
@lostmsu: I will address the CustomModule issue in an upcoming PR -- the class does not exist in Pytorch, as you have pointed out, and I will remove it.

PR #322, opened 2021-08-11
Removing CustomModule; bugfixes

NiklasGustafsson:
CustomModule was introduced as a base class for all non-framework modules, but had no corresponding class in the Pytorch API. This PR removes it from the API.

In addition, a couple of bug fixes and corrected APIs were added, and I added some doc comments that were missing.

This PR addresses issues #321 and the remaining problem in #318.

### DIFF
--- a/src/TorchSharp/NN/Bilinear.cs
+++ b/src/TorchSharp/NN/Bilinear.cs
@@ -26,7 +26,7 @@ namespace TorchSharp
             [DllImport("LibTorchSharp")]
             extern static IntPtr THSNN_Bilinear_forward(torch.nn.Module.HType module, IntPtr input1, IntPtr input2);
 
-            public Tensor forward(Tensor input1, Tensor input2)
+            public override Tensor forward(Tensor input1, Tensor input2)
             {
                 var res = THSNN_Bilinear_forward(handle, input1.Handle, input2.Handle);
                 if (res == IntPtr.Zero) { torch.CheckForErrors(); }
--- a/src/TorchSharp/NN/Bilinear.cs
+++ b/src/TorchSharp/NN/Bilinear.cs
@@ -74,6 +74,14 @@ namespace TorchSharp
             [DllImport("LibTorchSharp")]
             private static extern IntPtr THSNN_Bilinear_ctor(long in1_features, long in2_features, long output_size, bool bias, out IntPtr pBoxedModule);
 
+            /// <summary>
+            /// Applies a bilinear transformation to the incoming data
+            /// </summary>
+            /// <param name="in1Features">size of each first input sample</param>
+            /// <param name="in2Features">size of each second input sample</param>
+            /// <param name="outputSize">size of each output sample</param>
+            /// <param name="hasBias">If set to false, the layer will not learn an additive bias</param>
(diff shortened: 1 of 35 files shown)


src/TorchSharp/NN/Bilinear.cs
```diff
@@ -83,10 +91,22 @@ static public Bilinear Bilinear(long in1Features, long in2Features, long outputS
...
+                /// <summary>
+                /// Applies a bilinear transformation to the incoming data
+                /// </summary>
+                /// <returns></returns>
+                static public Tensor bilinear(Tensor input1, Tensor input2, Tensor weight, Tensor? bias = null)
                 {
-                    using (var d = nn.Bilinear(in1Features, in2Features, outputSize, hasBias)) {
-                        return d.forward(x1, x2);
+                    var in1Features = input1.shape[input1.shape.Length - 1];
+                    var in2Features = input2.shape[input2.shape.Length - 1];
```
lostmsu, line 101: NIT: inputX.shape[^1]

src/TorchSharp/NN/Module.cs
```diff
@@ -415,7 +452,9 @@ public virtual string GetName()
                     return res;
                 }
 
-                public virtual Tensor forward(Tensor t) => throw new NotImplementedException("forward");
+                public virtual Tensor forward(Tensor t) => throw new NotImplementedException("forward(t)");
+
+                public virtual Tensor forward(Tensor x, Tensor y) => throw new NotImplementedException("forward(x,y)");
```
lostmsu, line 457: It might be confusing, that overriding this function from C# on a custom module has no effect.

src/TorchSharp/NN/Bilinear.cs
NiklasGustafsson: Good catch. Still getting used to the idea that C# supports this kind of thing...

src/TorchSharp/NN/Module.cs
NiklasGustafsson: @lostmsu -- please elaborate. I'm not sure why it won't have any effect, I'm missing something.

dsyme (approved):

src/TorchSharp/NN/Bilinear.cs
lostmsu: Whenever I use that feature, I have to look up if I need to put `^0` or `^1` to get the last element. I wish it would be clearer.

src/TorchSharp/NN/Module.cs
lostmsu: I was referring to the 2 parameter overload. It is not used when the `THSNN_custom_module` is called.

NiklasGustafsson: Yes, that is a complex problem -- the fact that the Pytorch forward() really takes a variable-length argument list, so currently Sequential will only forward to the single-argument forward(), you have to build your own custom forward() to do anything else, like a couple of the examples do. Unfortunately, it also interacts with default arguments declared in Python, etc., so it's far from straight-forward to fix.

I'll punt on that in this PR, and open a specific issue to address the problem.

PR #352, opened 2021-09-20
Fixing #353

NiklasGustafsson:
Also, updating the release notes and taking explicit control over the patch version number.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -66,17 +66,19 @@
   <PropertyGroup>
 
     <VersionPrefix >$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
+    <!--
     <VersionSuffix>local-$(Configuration)-$([System.DateTime]::Now.ToString(yyyyMMdd))</VersionSuffix>
 
     <VersionPrefix Condition="'$(TF_BUILD)' != ''">$(MajorVersion).$(MinorVersion).$(MyRunNumber)</VersionPrefix>
     <VersionSuffix Condition="'$(TF_BUILD)' != ''  AND '$(BUILD_SOURCEBRANCHNAME)' == 'main'"></VersionSuffix>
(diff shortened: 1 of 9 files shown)


dsyme:
> taking explicit control over the patch version number.

This might cause problems, I'm not sure.  Is it needed?   The nugets from the builds go on the DevOps feed and I'm not sure what happens if two of them have the same version.  Guess we can try and see, I just hate debugging this stuff though :)

NiklasGustafsson:
We had the issue a couple of weeks ago that the build number didn't increase, so it got buried down the list. By taking explicit control, we can ensure that the numbers increase monotonically.

dsyme:
> We had the issue a couple of weeks ago that the build number didn't increase, so it got buried down the list. By taking explicit control, we can ensure that the numbers increase monotonically.

That was just because of the repo move. They do auto-increase

NiklasGustafsson:
Okay. Still, I do prefer having subsequent numbers starting at 0. That's what just about every other library does, across ecosystems.

dsyme:
> Okay. Still, I do prefer having subsequent numbers starting at 0. That's what just about every other library does, across ecosystems.

Yes that's ok.

dsyme:
I guess I just like being able to release any main branch build at the moment, without needing to think about it.  But I understand we need to be more normal

Directory.Build.props
```diff
@@ -66,17 +66,19 @@
   <PropertyGroup>
 
     <VersionPrefix >$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
+    <!--
     <VersionSuffix>local-$(Configuration)-$([System.DateTime]::Now.ToString(yyyyMMdd))</VersionSuffix>
```
dsyme, line 70: This line was very, very handy for `TF_BUILD` == `''` cases, i.e. local machine builds.  

This allowed me to get reliably named packages on the local machine that were distinct from anything else on the nuget feeds. I found this critical when debugging across DiffSharp --> TorchSharp builds (one thing .NET PackageReference doesn't do well is a really easy idiot-proof way to pick up local builds of things).

Anyway it's ok, I haven't needed to debug like that in a long time

dsyme (approved):

PR #354, opened 2021-09-21
Removing the '{Float|Int|Complex|Bool}Tensor' static tensor factory classes

NiklasGustafsson:
All factory methods are in 'torch' now.


dsyme (approved):

dsyme:
Cc @gbaydin  - we will need to update DiffSharp after this

gbaydin:
Thanks for the notice. One thing I'm confused about is how long after a merge do these changes show up in torchsharp packages and which version number will have the change.

NiklasGustafsson:
We're waiting for a signing certificate, so that we can start signing binaries now that TorchSharp is under .NET Foundation.

NiklasGustafsson:
@gbaydin -- we just released 0.93.0 to the NuGet gallery.

gbaydin:
Awesome, thank you for letting me know.

PR #359, opened 2021-09-23
Modules -- a more complete implementation.

NiklasGustafsson:
Adding {Module|Parameter}{List|Dict}

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -12,10 +12,15 @@ Fixed incorrectly implemented Module APIs related to parameter / module registra
 Changed Module.state_dict() and Module.load() to 'virtual,' so that saving and restoring state may be customized.
 #353 Missing torch.minimum (with an alternative raising exception)
 #327 Tensor.Data<T> should do a type check
+#358 Implement ModuleList / ModuleDict / Parameter / ParameterList / ParameterDict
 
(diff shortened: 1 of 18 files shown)


dsyme:
I've only looked through on my phone but it all looks good. I'll mark as approved - if you'd like me to do a line by line review I could do that tomorrow or next week

NiklasGustafsson:
> I've only looked through on my phone but it all looks good. I'll mark as approved - if you'd like me to do a line by line review I could do that tomorrow or next week

No need. It's pretty straight-forward.

dsyme (approved):

pkese:
@NiklasGustafsson I went through the text and it is wonderful.

The only thing I had trouble with is in the first example,
```csharp
public TestModule1(Tensor tensor, bool withGrad)
```
... where the `tensor` parameter is never used so I spent some time figuring out what it is doing there.

I assume it is a static parameter which becomes useful in a Sequential definition, but in this particular case it's just a distraction to the reader since it is unused. Maybe it would be a better idea to explain the tensor-as-a-module-parameter concept separately and omit it in the first example.

NiklasGustafsson:
@pkese -- thanks for pointing that out. That argument should not be there at all. I'll remove it.

PR #366, opened 2021-09-24
Implemented missing optimizers.

NiklasGustafsson:
Fixed the 'step()' function interface. It should return a Tensor if there's a closure passed in.

### DIFF
--- a/src/TorchSharp/Tensor/Tensor.cs
+++ b/src/TorchSharp/Tensor/Tensor.cs
@@ -465,7 +465,7 @@ namespace TorchSharp
             /// <summary>
             /// Is the tensor a sparse tensor?
             /// </summary>
-            public bool IsSparse {
+            public bool is_sparse {
                 get {
                     var res = THSTensor_is_sparse(handle);
                     torch.CheckForErrors();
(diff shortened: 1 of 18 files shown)


src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -888,15 +888,18 @@ public Tensor contiguous()
             static extern IntPtr THSTensor_grad(IntPtr handle);
 
             /// <summary>
-            /// This attribute is None by default and becomes a Tensor the first time a call to backward() computes gradients for the tensor.
+            /// This attribute is null by default and becomes a Tensor the first time a call to backward() computes gradients for the tensor.
             /// The attribute will then contain the gradients computed and future calls to backward() will accumulate (add) gradients into it.
             /// </summary>
             /// <returns></returns>
-            public Tensor grad()
+            public Tensor? grad()
```
dsyme, line 895: Any ideas when/why may this may be null?

test/TorchSharpTest/TestTraining.cs
```diff
@@ -99,8 +101,10 @@ public void TestTrainingWithDropout()
                 using (torch.no_grad()) {
                     foreach (var param in seq.parameters()) {
                         var grad = param.grad();
-                        var update = grad.mul(learning_rate);
-                        param.sub_(update);
+                        if (grad is not null) {
```
dsyme, line 104: I wonder if simply raising a NullReferenceException may be better than requiring the nullness checks, in the balance of things. It's in the nature of tensor programming to have failure points and it's not necessarily what we're trying to solve here...

dsyme (approved):

test/TorchSharpTest/TestTraining.cs
NiklasGustafsson: Except that grad() being null is not a failure -- it is expected to be null before backward() is called the first time, as well as when you are writing optimizers (that's how this came up), when you often set 'no_grad()' for a while. In PyTorch, there's also an option to set grad to null rather than zeroing when you call zero_grad() (which TorchSharp doesn't implement, but may in the future). Anyway, an exception would be bad, IMO.

PR #381, opened 2021-10-05
Improve Transform Docs

ekagra-ranjan:
* Add params to Transform docs
* rename `mode` param in RandomRotation to `interpolation` similar to [Pytorch](https://github.com/pytorch/vision/blob/b1ef19639ee845902ecf60930757d028f1e1491c/torchvision/transforms/transforms.py#L1267)


ekagra-ranjan:
Will reopen this later after

PR #385, opened 2021-10-06
Improve Transform Docs

ekagra-ranjan:
* Add params to Transform docs
* rename `mode` param in RandomRotation to `interpolation` similar to [Pytorch](https://github.com/pytorch/vision/blob/b1ef19639ee845902ecf60930757d028f1e1491c/torchvision/transforms/transforms.py#L1267)


NiklasGustafsson:
> I have updated the transform param comments with descriptions starting with a capital letter.

I saw that. I just  pushed a fix to the breaking Linux builds, so your branch was out-of-date, and I pushed an update to it. It's building again, and as soon as the checks are done everywhere, I'll merge.

NiklasGustafsson (approved):

Issue #348, opened 2021-09-13
Create combo packages


dsyme:
Part of the problem underlying stuff like #347 is the need for this "loading logic" in TorchSharp at all, it's a mess especially the cases for F# Interactive. 

On talking it through with @gbaydin, one suggestion was that we should add "runtime ready" combo packages that combine TorchSharp with the appropriate runtime, and also have these ensure that on package restore the native binaries are put together in a place loadable by F# Interactive and .NET Interactive, without any hassle or initialization code.

These would be the three combo packages:

* TorchSharp-cpu
  * references matching TorchSharp package
  * references matching libtorch-cpu package
  * on restore, ensures that native binaries are all available in one known directory

* TorchSharp-cuda-11.1-linux-x64
  * references matching TorchSharp package
  * references matching matching libtorch-cuda-11.1-linux-x64 package
  * on restore, ensures that native binaries are all available in one known directory

* TorchSharp-cuda-11.1-windows-x64
  * references matching TorchSharp package
  * references matching libtorch-cuda-11.1-windows package
  * on restore, ensures that native binaries are all available in one known directory

It's important to still have the plain "TorchSharp" package because this is what things like DiffSharp.Backend.Torch reference - it's the programming model with (most) of its implementation, but short of the selected libtorch native binaries. 

The above are in addition to the existing packages:

* libtorch-cpu

* libtorch-cuda-11.1-linux-x64
  * plus various "fragment" packages for delivery
  * on restore, restitches fragments 

* libtorch-cuda-11.1-windows-x64
  * plus various "fragment" packages for delivery
  * on restore, restitches fragments

dsyme:
Note we use combo packages in DiffSharp to give a single package to reference - the above would also simplify DiffSharp and correctly put the sanity checking for an overall valid set of references in TorchSharp

gbaydin:
Sounds great!

gbaydin:
I think libtorch-cpu covers all of linux, macos, windows and there is no libtorch-cuda because libtorch doesn't support cuda on macos. See below from https://pytorch.org/get-started/locally/

![Screenshot_2021-09-21_18-30-08](https://user-images.githubusercontent.com/5365982/134210127-8649d480-1502-4127-8425-12e158ffde3d.png)

dsyme:
This is done, but we need to check pacakges are building in main and then republish

dsyme:
[...]

PR #387, opened 2021-10-06
Add combo packages

dsyme:
This adds one-stop packages that combine TorchSharp with an appropriate backend

* `TorchSharp-cpu` (CPU)

* `TorchSharp-cuda-linux` (CPU)

* `TorchSharp-cuda-windows` (CPU)

This lets people reference just `TorchSharp-cpu` for example and everything should work with right versions - they don't need to know which version of CUDA or libtorch is being used, nor see "x64", nor even know the libtorch packages exist. The CUDA version and architecture is left implied ro "latest supported by TorchSharp".   This is good for longevity of samples


NiklasGustafsson:
@dsyme -- This is in addition to the existing packages, right?

dsyme:
> @dsyme -- This is in addition to the existing packages, right?



Yes

NiklasGustafsson:
Does the Azure-pipelines YAML file need any updating for these packages?

dsyme:
@NiklasGustafsson Yes it will, I'll think about that

See https://github.com/dotnet/TorchSharp/issues/348 also

NiklasGustafsson:
@dsyme I think your changes to the README file are a little confusing. 

Should I use TorchSharp-XXX, or TorchSharp + libtorch-XXX? Is that my choice, or does this new package structure replace the old manual combos? I'm not sure the text makes that clear.

dsyme:
@NiklasGustafsson On the naming, I think we should go for the simpler names:

* `TorchSharp-cpu`

* `TorchSharp-cuda-linux`

* `TorchSharp-cuda-windows`

That is a typical notebook will just have

```fsharp
#r "nuget: TorchSharp-cuda-linux"
```

and you'll get 64-bit Linux CUDA for whatever cuda version is the latest supported by TorchSharp.

NiklasGustafsson:
Not supporting CUDA 10.2 will make some people unhappy, so we need to cover that in the README.

gbaydin:
Do I understand it correctly that `TorchSharp`, `TorchSharp-cpu`, `TorchSharp-cuda-linux`, `TorchSharp-cuda-windows` are guaranteed to share the same version number in each new release?

NiklasGustafsson:
Yes. The build script builds all of them, and the script to push to NuGet pushes them all, so they should be in sync. The latter three are actually empty, they just contain the signature and references to TorchSharp and the right libtorch-xxx backend library. Right now, that is 0.93.5, which contains an important bug fix to how Data<T>() (also renamed 'data<T>()' in the same release) was implemented.

The libtorch-xxx packages, on the other hand, will continue to have version numbers based on the release from Pytorch.org that they wrap.

PR #388, opened 2021-10-06
fix typo

Leo-LiHao:
fix typo for MNIST example


ekagra-ranjan:
Should we keep only one Relu layer instead of three as this activation function does not have any trainable parameter?

PR #392, opened 2021-10-06
Minor Optimizations for Device type checks

ekagra-ranjan:
* Reduce the number of device checks in Examples
* Prevent tensors in Normalize() transform from being sent to `DeviceType.CPU` if the device not set to some other non-null value

### DIFF
--- a/src/Examples/VGG.cs
+++ b/src/Examples/VGG.cs
@@ -54,7 +54,7 @@ namespace TorchSharp.Examples
 
             RegisterComponents();
 
-            if (device != null && device.type == DeviceType.CUDA)
+            if (device.type == DeviceType.CUDA)
                 this.to(device);
         }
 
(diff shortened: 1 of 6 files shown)


src/Examples/VGG.cs
```diff
@@ -54,7 +54,7 @@ public VGG(string name, int numClasses, Device device = null) : base(name)
 
             RegisterComponents();
 
-            if (device != null && device.type == DeviceType.CUDA)
```
NiklasGustafsson, line 57: You can't remove the null check as long as 'null' is a possible value, which it is given the default value in the signature. This goes for all these changes.

NiklasGustafsson (changes requested):

src/TorchSharp/TorchVision/Normalize.cs
```diff
@@ -21,7 +21,8 @@ internal Normalize(double[] means, double[] stdevs, ScalarType dtype = ScalarTyp
                 this.means = this.means.to_type(dtype);
                 this.stdevs = this.stdevs.to_type(dtype);
             }
-            if (device != null) {
+
+            if (device != null && device.type != DeviceType.CPU) {
```
ekagra-ranjan, line 25: @NiklasGustafsson Is this change fine? Here null check is not being removed.

NiklasGustafsson: Yes, here is it is a (slight) performance benefit.

Issue #397, opened 2021-10-07
.Data<Type>() --> .data() ?


ekagra-ranjan:
Pytorch has `.data()` and `.item()` functions for Tensors which is represented in TorchSharp as `.Data<Type>()` and `.DataItem<Type>()` resp. 

Following the Pytorch syntax, should we **add** these function in TorchSharp? To remove the need of passing `<Type>`, we can use the `dtype` member in TorchSharp Tensor. 

For backward compatibility we can add these functions rather than replacing the older ones.

NiklasGustafsson:
The <Type> argument is important since it returns a statically typed span or scalar -- it tells the API what the user is expecting to come out of the tensor, and we can then raise an exception if that's not what's in it.

This is one of those areas where Python and .NET diverge and sticking with the .NET semantics is essential, since it's bridging the torch type system and the .NET type system.

That said, I can easily be persuaded that we should use exactly the same names that PyTorch does, and also not capitalize. The type argument is essential, though.

PR #398, opened 2021-10-07
Make Data and DataItem similar to Pytorch

ekagra-ranjan:
Addresses #397


NiklasGustafsson:
> oh, okay. I have the changes ready offline so do let me know whenever this thing is ready to go.

If it's just the name change, how about I just do it as part of the reimplementation of Data<T>()?

Issue #399, opened 2021-10-07
Data<T>() returns span that must be indexed using strides.


NiklasGustafsson:
As reported by @shonk-msft:

This shows the issue:

```C#
using System;
using System.Text;
using TorchSharp;

class Program
{
    static void Main(string[] args)
    {
        // Create a contiguous 3x4 matrix and fill with auto-inc values starting at zero.
        // 0 1 2 3
        // 4 5 6 7
        // 8 9 10 11
        using var contig = torch.zeros(3, 4, torch.ScalarType.Int32);
        var data = contig.Data<int>();
        for (int i = 0; i < data.Length; i++)
            data[i] = i;
        Console.WriteLine(DumpData(contig));

        // Create a 3x2 slice of this, which should contain:
        // 1 2
        // 5 6
        // 9 10
        // Dumping values using Data<int>() shows incorrect values.
        using var sub = contig.slice(1, 1, 3, 1);
        Console.WriteLine(DumpData(sub));

        // Dumping sub.contiguous() shows the correct values.
        using var subContig = sub.contiguous();
        Console.WriteLine(DumpData(subContig));
    }

    static string DumpData(torch.Tensor ten)
    {
        var sb = new StringBuilder();
        sb.AppendFormat("Rank: {0}", ten.ndim).AppendLine();

        sb.Append("Shape: (");
        string pre = "";
        foreach (var dim in ten.shape)
        {
            sb.Append(pre).Append(dim);
            pre = ", ";
        }
         sb.AppendLine(")");

         sb.Append("Stride: (");
         pre = "";
         foreach (var dim in ten.stride())
         {
             sb.Append(pre).Append(dim);
             pre = ", ";
         }
         sb.AppendLine(")");

         var data = ten.Data<int>();
         sb.Append("Length: ").Append(data.Length).AppendLine();
         pre = "";
         for (int i = 0; i < data.Length; i++)
         {
             sb.Append(pre).Append(data[i]);
             pre = ", ";
         }
         sb.AppendLine();

         return sb.ToString();
    }
}
```
Running this produces:

```
Rank: 2
Shape: (3, 4)
Stride: (4, 1)
Length: 12
0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11

Rank: 2
Shape: (3, 2)
Stride: (4, 1)
Length: 6
1, 2, 3, 4, 5, 6

Rank: 2
Shape: (3, 2)
Stride: (2, 1)
Length: 6
[...]

PR #400, opened 2021-10-07
Implement Data<T> in a way that supports non-contiguous tensors.

NiklasGustafsson:
Addresses bug #399 by supporting non-contiguous tensors, and #398 by renaming Data<T> and DataItem<T>


NiklasGustafsson:
This is an important one to review -- there are actual algorithmic additions, not just new interop with C++.

PR #407, opened 2021-10-08
rand() and randn() must check that the dtype is floating-point.

NiklasGustafsson:


NiklasGustafsson:
It gives an error that isn't very helpful.

PR #424, opened 2021-11-01
Update to LibTorch 1.10.0

dsyme:


dsyme:
Cool looks like one consistent failure across all platforms

```
  Failed TorchSharp.TestTorchTensorBugs.ValidateIssue399_4 [4 ms]
```

NiklasGustafsson:
Yes, that test was added to catch the point where negatives strides may be returned by a view. Apparently, 1.10.0 did so.

PR #433, opened 2021-11-08
Make able to declare sequential without name of module

dayo05:
In pytorch, there are two ways for declare sequential.

```py
model = nn.Sequential(
          nn.Conv2d(1,20,5),
          nn.ReLU(),
          nn.Conv2d(20,64,5),
          nn.ReLU()
        )

# Using Sequential with OrderedDict. This is functionally the
# same as the above code
model = nn.Sequential(OrderedDict([
          ('conv1', nn.Conv2d(1,20,5)),
          ('relu1', nn.ReLU()),
          ('conv2', nn.Conv2d(20,64,5)),
          ('relu2', nn.ReLU())
        ]))
```
But torchsharp only has second way. So I added first way.

In my commit, when using first way, It will set name as random. And I add test for new sequential.

### DIFF
--- a/src/TorchSharp/NN/Sequential.cs
+++ b/src/TorchSharp/NN/Sequential.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Runtime.InteropServices;
+using TorchSharp.torchvision;
 using static TorchSharp.torch;
 
 namespace TorchSharp
@@ -35,6 +36,9 @@ namespace TorchSharp
                 _modules.Add(submodule);
                 _names.Add(name);
             }
+
+            public void Add(torch.nn.Module module)
+                => Add(new Random().NextDouble().ToString()[2..], module);
 
             internal Sequential(IntPtr handle) : base(handle, IntPtr.Zero)
             {
@@ -86,6 +90,18 @@ namespace TorchSharp
             extern static IntPtr THSNN_Sequential_ctor();
 
             /// <summary>
+            /// Get empty sequential
+            /// </summary>
+            /// <returns></returns>
+            static public Sequential Sequential()
+            {
+                var handle = THSNN_Sequential_ctor();
+                if(handle == IntPtr.Zero) {torch.CheckForErrors();}
+
+                return new Sequential(handle);
+            }
+
+            /// <summary>
             /// A sequential container. Modules will be added to it in the order they are passed in the constructor.
             /// Alternatively, an OrderedDict of modules can be passed in. The forward() method of Sequential accepts any input and forwards it to the first module it contains.
             /// It then “chains” outputs to inputs sequentially for each subsequent module, finally returning the output of the last module.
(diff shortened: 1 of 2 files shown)


dayo05:
I found that F# compile failed but I don't know about F#. Is anyone can advice code for me...?

src/TorchSharp/NN/Sequential.cs
```diff
@@ -96,14 +112,28 @@ public static partial class nn
...
+            /// The value a Sequential provides over manually calling a sequence of modules is that it allows treating the whole container as a single module,
+            /// such that performing a transformation on the Sequential applies to each of the modules it stores (which are each a registered submodule of the Sequential).
+            /// </summary>
+            /// <param name="modules">An ordered list of the contained modules.</param>
+            /// <returns></returns>
+            static public Sequential Sequential(params torch.nn.Module[] modules)
+            {
+                var res = Sequential();
+                foreach(var m in modules)
+                    res.Add(new Random().NextDouble().ToString()[2..], m);
```
NiklasGustafsson, line 133: I would just keep a counter and use it with the first N characters from the modules class name, i.e. something that is deterministic. With random numbers in the name, it will be impossible to save and load the state of a module.

src/TorchSharp/NN/Sequential.cs
```diff
@@ -5,6 +5,7 @@
 using System.Diagnostics;
 using System.Linq;
 using System.Runtime.InteropServices;
+using TorchSharp.torchvision;
```
NiklasGustafsson, line 8: Why? I would like to avoid circular dependencies between torch.nn and torchvision. The latter should, technically, be separable into its own package some time in the future.

src/TorchSharp/NN/Sequential.cs
```diff
@@ -36,6 +37,9 @@ public void Add(string name, torch.nn.Module submodule)
                 _names.Add(name);
             }
 
+            public void Add(torch.nn.Module module)
+                => Add(new Random().NextDouble().ToString()[2..], module);
```
NiklasGustafsson, line 41: See my comment on using random numbers in the module names later on in the file.

test/TorchSharpTest/NN.cs
```diff
@@ -444,6 +444,58 @@ public void EvalLossSequence()
 
             var result = output.ToSingle();
         }
+
```
NiklasGustafsson, line 447: Given my comments on loading and saving earlier, please add a unit test to TestLoadSave that saves and restores the state. Something like 'TestSaveLoadConv2D' except using a Sequential instead of Conv2D as the test.

NiklasGustafsson (changes requested):

NiklasGustafsson:
@dayo05 -- feel free to introduce yourself in https://github.com/dotnet/TorchSharp/discussions/333. It's great if all contributors tell the community of users a little bit about themselves.

dayo05:
Thanks for comments!

PR #434, opened 2021-11-08
Fixed missing temporaries Dispose in Sequential.forward()

NiklasGustafsson:
Also updated README and azure-pipelines.yaml.

### DIFF
--- a/docfx/articles/memory.md
+++ b/docfx/articles/memory.md
@@ -10,27 +10,22 @@ Note DiffSharp (which uses TorchSharp) relies on techniques 1.
 
 > Most of the examples included will use technique #1, doing frequent explicit calls to GC.Collect() in the training code -- if not after each batch in the training loop, at least after each epoch.
 
-## Technique 1. Implicit disposal using finalizers
+## Technique 1. Automatic disposal via Garbage Collection
 
-In this technique all tensors (CPU and GPU) are implicitly disposed via .NET finalizers.
-
-When allocation of tensors via `Float32Tensor.*` (likewise `Int64Tensor.*` etc.) fails (whether on GPU or CPU),
-then TorchSharp forces a .NET garbage collection and execution of all pending finalizers.
-
-This is not yet done when using general tensor operations.  It is possible a more general retry-after-GC-on-out-of-memory will be added at some point.
+In this technique all tensors (CPU and GPU) are implicitly disposed via .NET finalizers. Just allocate new tensors to your heart's content, and let GC take care of them. It will only work for small models that do not require a lot of memory. If you do use this approach, you may want to place a call to `GC.Collect()` after each mini-batch of data. It is generally not sufficient to do it at the end of each epoch.
 
 👍 Simple
 
-👎 The .NET GC doesn't know of the memory pressure from CPU tensors, so failure may happen if large tensors can't be allocated
+👎 The .NET GC doesn't know about the memory pressure from CPU tensors, so failure may happen if large tensors can't be allocated
 
-👎 The .NET GC doesn't know of GPU resources.
+👎 The .NET GC doesn't know about GPU resources.
 
 👎 Native operations that allocate temporaries, whether on CPU or GPU, may fail -- the GC scheme implemented by TorchSharp only works when the allocation is initiated by .NET code.
 
+
 ## Technique 2. Explicit disposal
 
-In this technique specific tensors (CPU and GPU) are explicitly disposed
-using `using` in C# or explicit calls to `System.IDisposable.Dispose()`.
+This technique is more cumbersome, but will result in better performance and have a higher memory ceiling. For many non-trivial models, it is more or less required in order to train on a GPU.
 
 👍 Specific lifetime management of all resources.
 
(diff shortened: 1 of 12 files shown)


dsyme:
Some notes for @NiklasGustafsson :)

```
## Technique 2. Explicit disposal

In this technique specific tensors (CPU and GPU) are explicitly disposed
using `using` in C# or explicit calls to `System.IDisposable.Dispose()`.

👍 Specific lifetime management of all resources.

👎 Cumbersome, requiring lots of using statements in your code.

👎 You must know when to dispose.

👎 To maximize the benefit, you may have to store all temporaries to variables and dispose. 

In TorchSharp tensor programming, the .NET Tensor objects are "tensor handles" that are like shared pointers to an underlying C++ tensor. Reference counting is used to the underlying tensor. Each "tensor handle" can only be disposed once. For this reason, to write .NET code that accurately disposes of tensors requires some care. These are the rules you need to know:

* All Tensor functions and "forward" functions in TorchSharp itself always return fresh tensor handles that you may dispose. This applies even to in-place mutating operations such as `left.add_(right, alpha)`.  
* Our intention is that exceptions to this rule will always noted in the documentation - one example is `model.parameters()`.  Furthermore
* If writing your own tensor functions, and you need to return one of the inputs, you can use `t.alias()` (TBD)
* Disposing a tensor only releases the underlying storage if this is the last live TorchTensor which has a view on that tensor -- the native runtime does reference counting of tensors.

Examples:

```fsharp

let myTensorFunction0(input: Tensor) =
    input.alias()

let myTensorFunction1() =
    if today then 
       table[4].alias()  
    else
       table[5].alias()  

let myTensorFunction2(input: Tensor) =
    input.add(tensor(1))

let myTensorFunction3(input: Tensor) =
    use tmp = input.add(tensor(1))
    tmp.add(tensor(1))

let myTensorFunction4(input: Tensor) =
    use tmp1 = input.add(tensor(1))
    use tmp2 = input.add(tensor(1))
    tmp2.add(tensor(1))

let myTensorFunction5(go: bool, input: Tensor) =
    if go then
        use tmp1 = input.add(tensor(1))
        use tmp2 = input.add(tensor(1))
        tmp2.add(tensor(1))
    else
        input.alias()
    
let myTensorFunction5(go: bool, input: Tensor) =
    if go then 
        use tmp1 = input.add_(tensor(1))  // NOTE: even for in-place mutations
        use tmp2 = input.add_(tensor(1))  // NOTE: even for in-place mutations
        tmp2.add(tensor(1))
    else
        input.alias()
```

docfx/articles/memory.md
```diff
@@ -40,14 +35,67 @@ using `using` in C# or explicit calls to `System.IDisposable.Dispose()`.
...
-> NOTE: Disposing a tensor only releases the underlying storage if this is the last
-> live TorchTensor which has a view on that tensor -- the native runtime does reference counting of tensors.
+__Note__: Even with this approach, it is a good idea to place a call to `GC.Collect()` after each mini-batch of data. There may be temporaries that were overlooked, or inconvenient to pull out, or ones where the lifetime was unclear; calling `GC.Collect()` will catch them.
+
+
+### Returning Fresh References
+
+It is important to understand that all TorchSharp tensors are simple wrappers around a C++ tensor. The 'Handle' property on `Tensor` is a pointer to a C++ `shared_ptr<at::Tensor>`. When a Tensor is created and returned to .NET, the reference count is incremented. When you call `Dispose()` on the tensor, it is decremented.
+
+To enable this technique, all Tensor operations should return a fresh reference to a Tensor, if not a fresh Tensor. This is true even for in-place, destructive operations like `add_()`, which overwrites the underlying native tensor with data, but still returns a Tensor with a handle that is different from the one going in.
```
dsyme, line 45: Suggestion:

```
It is important to understand that all TorchSharp "tensors" (type Tensor) are actually "tensor references", referring to a C++ tensor. When a C++ tensor is created and returned to .NET as a tensor reference, the reference count on the C++ tensor is incremented. When you call `Dispose()` on the TorchSharp tensor reference (that is, type Tensor), it is decremented. If the tensor reference is disposed, the decrement happens implicitly.

To enable this technique, all operations that return one or more TorchSharp `Tensor`s should return "fresh" Tensor objects (though that doesn't necessarily mean freshly copied C++ Tensors). This is true even for in-place, destructive operations like `add_()`, which overwrites the underlying native tensor with data, but still returns a fresh Tensor reference to that same tensor.  

Thus, when you write methods and functions that take and produce Tensor, for example in the `forward()` method of a model, you should always make sure to return a fresh reference. Most of the time, this happens automatically, because the last action of your code will normally be to call another tensor function, but there are cases when it's not, especially when returning input tensors or tensors stored in some lookaside table.
```

NOTE: should the getters of properties of models return fresh tensor handles?  I suppose not. But then we should explicitly call out when this is not done, e.g. `model.weights` and `model.parameters()` etc.

src/TorchSharp/NN/Sequential.cs
```diff
@@ -45,17 +45,29 @@ internal Sequential(IntPtr handle) : base(handle, IntPtr.Zero)
 
             public override Tensor forward(Tensor tensor)
             {
+                // If there are no modules, just return a clone of the input.
                 if (_modules.Count == 0) return tensor.clone();
```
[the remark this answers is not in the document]
dsyme: I think this should be `tensor.alias()`

docfx/articles/memory.md
dsyme: We could perhaps systematically call them "Tensor aliases".  The use of "Tensor reference" is really confusing given the presence of "reference" and "by-ref(erence)" in .NET languages

src/TorchSharp/NN/Sequential.cs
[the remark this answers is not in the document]
NiklasGustafsson: Yes. It says 'alias' when I look at the file diff.

docfx/articles/memory.md
QingtaoLi1: Sorry if I bother you. For in-place functions, I guess we can dispose the original handle and assign the new handle to the C# Tensor, so that this behavior will be transparent to TorchSharp users and "x.add_().mul_().add_()" is still good practice.

PR #435, opened 2021-11-09
Make able to declare sequential without name of module

dayo05:
In pytorch, there are two ways for declare sequential.
```
model = nn.Sequential(
          nn.Conv2d(1,20,5),
          nn.ReLU(),
          nn.Conv2d(20,64,5),
          nn.ReLU()
        )

# Using Sequential with OrderedDict. This is functionally the same as the above code
model = nn.Sequential(OrderedDict([
          ('conv1', nn.Conv2d(1,20,5)),
          ('relu1', nn.ReLU()),
          ('conv2', nn.Conv2d(20,64,5)),
          ('relu2', nn.ReLU())
        ]))
```
But torchsharp only has second way. So I added first way.

```cs
var model = nn.Sequential(
    nn.Conv2d(1, 20, 5),
    nn.ReLU(),
    nn.Conv2d(20, 64, 5),
    nn.ReLU()
);
```

This works same with first python code.

In my commit, when using first way, It will set name by type name of module. And I add test for new sequential(Save and loading, nn in sequential)

I think `using torch.torchvision` in #433 was just mistake :D

### DIFF
--- a/src/TorchSharp/NN/Sequential.cs
+++ b/src/TorchSharp/NN/Sequential.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Runtime.InteropServices;
+using TorchSharp.torchvision;
 using static TorchSharp.torch;
 
 namespace TorchSharp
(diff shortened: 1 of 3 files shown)


NiklasGustafsson:
Since this is failing on F#, I'll merge into my own forked repo and see if I can address the problem.

NiklasGustafsson:
@dayo05 I just sent a PR to your repo.

PR #443, opened 2021-11-10
Adding Tensor.TotalCount and Tensor.PeakCount

NiklasGustafsson:
The purpose of these static properties is to help diagnosing memory management issues. The TotalCount property can be used to monitor the stability of the number of allocated tensors, which should remain steady across training epochs.


NiklasGustafsson:
Another trivial PR that I want to merge in order to do a release. The fix to Sequential is essential in order to support explicit memory management.

dsyme:
Cool!

PR #447, opened 2021-11-12
Torchvision memory management

NiklasGustafsson:
I found a whole slew of torchvision functions that did not properly dispose of temporaries. The same problem was found in a number of the optimizers whose step() functions are written in .NET rather than native code.

In the process of fixing that issue, I also moved a number of the algorithms to native code, implementing them in C++ to avoid the interop overhead and the overhead of creating and disposing of short-lived .NET tensor aliases.

Also -- addressed an infinite recursion problem in a couple of the randint() overloads.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -2,7 +2,7 @@
 
 Releases, starting with 9/2/2021, are listed with the most recent release at the top.
 
-### NuGet Version 0.95.2
+### NuGet Version 0.95.3
 
 __API Changes:__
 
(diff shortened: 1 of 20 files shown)


NiklasGustafsson:
@dsyme -- I requested a review. This is not a trivial PR.

There are a number of things that were just moved to C++, and a plethora of 'using' clauses adding in a lot of places. Those areas don't generally require a lot of review.

However, there is now a validation check in Tensor.Handle, and a function 'DecoupleFromNativeHandle' which is meant to clear the internal Tensor handle without decrementing the count for the native handle. Its introduction and use requires some scrutiny.

dsyme (approved):

NiklasGustafsson:
Merging

PR #454, opened 2021-11-19
Dispose Scope

mfagerlund:
Dispose Scopes simplifies tensor lifetime management


NiklasGustafsson:
I'm seeing this error during tests:

   System.InvalidOperationException : Collection was modified; enumeration operation may not execute.

The fact that it doesn't happen on all platforms suggests there's a multi-threading issue, that will cause intermittent results.

mfagerlund:
> I'm seeing this error during tests:
> 
> ```
>    System.InvalidOperationException : Collection was modified; enumeration operation may not execute.
> ```
> 
> The fact that it doesn't happen on all platforms suggests there's a multi-threading issue, that will cause intermittent results.

That's alarming. I'll investigate. 

/m

NiklasGustafsson:
See, for example:

https://dev.azure.com/dotnet/TorchSharp/_build/results?buildId=61218&view=logs&j=660e8eb5-d5ee-5653-778d-3490b8f06993&t=a2324835-2540-5fc0-fd49-20284380c634

mfagerlund:
Right, it's easy enough to add locking or use thread safe classes - but I'm confused as to how a thread static class can be accessed from multiple threads - but with that error report I might be able to pinpoint it.

NiklasGustafsson:
Your calling Disposables.Remove() from within a loop over Disposables:

```C#
            private void ReleaseUnmanagedResources()
            {
                lock (Singleton) {
                    foreach (var disposable in Disposables) {
                        DoDispose(disposable);
                    }

                    Singleton.RemoveDisposeScope(this);
                }
            }

            private void DoDispose(IDisposable disposable)
            {
                _disposeScopeManager.CurrentlyDisposing = disposable;
                try {
                    if (disposable is torch.Tensor tensor) {
                        if (!tensor.IsDisposed) {
                            tensor.Dispose();
                        }
                    } else {
                        disposable.Dispose();
                    }

                    Interlocked.Increment(ref _disposeScopeManager._disposedCount);
                    Disposables.Remove(disposable);
                } finally {
                    _disposeScopeManager.CurrentlyDisposing = null;
                }
            }
```

I think that should be failing all the time, not just every now and then.

NiklasGustafsson:
Thread-static classes can be accessed from the GC thread, I presume?

mfagerlund:
> Your calling Disposables.Remove() from within a loop over Disposables:

Well spotted, that's rather stupid of me.

> I think that should be failing all the time, not just every now and then.

Yes, that is odd.
 
> Instead of individually removing elements, why not clear the collection after the loop?

It's called from two places, one where it doesn't iterate over the entire list. I didn't want to repeat the code.

mfagerlund:
> Thread-static classes can be accessed from the GC thread, I presume?

Yeah, you're right, I've fixed  the bug you mention but I'll also add locking everywhere because if something's disposed by the GC for whatever reason (though they shouldn't since I have a reference  to them) we'll want protection.

src/TorchSharp/DisposeScopeManager.cs
```diff
@@ -0,0 +1,306 @@
...
+            /// clear up some memory without creating a new scope. Note that this doesn't permanently exclude the
+            /// tensors from disposing, use Exclude for that. Also, excluded tensors don't need to be included
+            /// here.
+            /// </summary>
+            public IDisposable[] DisposeEverythingBut(params IDisposable[] keep)
+            {
+                lock (this) {
+                    foreach (var disposable in keep) {
+                        if (!Disposables.Contains(disposable)) {
+                            throw new InvalidOperationException("The disposable does not belong to this scope!");
```
NiklasGustafsson, line 231: It's potentially quite a benign error to exclude something that isn't actually in the scope. It feels good to check for this, but I don't think it's disastrous to ignore it.

NiklasGustafsson:
>> It's called from two places, one where it doesn't iterate over the entire list. I didn't want to repeat the code.

Assuming that the 'keep' list is typically shorter than the non-keep list, you could save the keeps, call DoDispose(), clear the list (or just discard it, letting GC pick it up later), and replace Disposables with the keep list. That **should** mean fewer calls to 'Remove()' and still allow DoDispose() to be called in both places.

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: Agreed, sometimes I add that kind of tests  to make sure stuff's actually working the way I intended, if I just ignored that kind of "error" when coding could be hiding a bigger problem. But once it works as intended, if someone wanted to Exclude something that wasn't actually ever Included, then that should probably just be ignored.

mfagerlund: Oh, also, should it still be moved to the outer scope? What if it never belonged to the outer scope, and I ask for it to be ignored, where does it go? I'm thinking we just ignore it and move on, without moving it up?

NiklasGustafsson: I don't think something that was never in the inner scope should automatically be moved to the outer scope. If, for example, it was excluded globally, it could create some issues to do so. For example, I imagine that the tensors created in the input data preprocessing pipeline will be excluded globally, and you really do not want those accidentally transferred to a parent scope.

src/TorchSharp/DisposeScopeManager.cs
```diff
@@ -0,0 +1,306 @@
...
+
+        static DisposeScopeManager()
+        {
+            torch.Tensor.OnTensorCreateCallback.Add(TryRegisterInDisposeScope);
+            torch.Tensor.OnTensorDisposeCallback.Add(TryUnregisterInDisposeScope);
+        }
+
+        [ThreadStatic] private static DisposeScopeManager _singleton;
+
+        public static DisposeScopeManager Singleton = (_singleton ??= new DisposeScopeManager());
```
NiklasGustafsson, line 30: Why is this public?

src/TorchSharp/DisposeScopeManager.cs
```diff
@@ -0,0 +1,306 @@
...
+using System.Threading;
+
+namespace TorchSharp
+{
+    /// <summary>
+    /// Manages dispose scopes, that can make automatic tensor disposal easier. Note that the
+    /// DisposeManager is thread local. The DisposeScopeManager can also manage other disposables, such as training
+    /// batches and the like.
+    /// </summary>
+    public class DisposeScopeManager
```
NiklasGustafsson, line 17: Is there some way to avoid having this as a public class? Adding some static methods to torch.?? Not sure, but it feels like it should be possible to avoid it being public, and to only have DisposeScope exposed.

src/TorchSharp/DisposeScopeManager.cs
```diff
@@ -0,0 +1,315 @@
...
+            /// </summary>
+            public void DisposeEverything() => DisposeEverythingBut();
+
+            /// <summary>
+            /// As an intermediate step, you can dispose all the tensors/disposables currently scheduled for dispose, to
+            /// clear up some memory without creating a new scope. Note that this doesn't permanently exclude the
+            /// tensors from disposing, use Exclude for that. Also, excluded tensors don't need to be included
+            /// here.
+            /// </summary>
+            public IDisposable[] DisposeEverythingBut(params IDisposable[] keep)
```
NiklasGustafsson, line 233: I've found params arrays super-useful in TorchSharp, but I've also found it super-useful to have an IEnumerable<T> version and put the implementation inside there. Sometimes, it's just so useful to build up a data structure dynamically and pass it in, and to being able to avoid calling 'ToList()' and make a copy would be ideal.

src/TorchSharp/DisposeScopeManager.cs
```diff
@@ -0,0 +1,315 @@
...
+        [ThreadStatic] private static DisposeScopeManager _singleton;
+
+        public static DisposeScopeManager Singleton = (_singleton ??= new DisposeScopeManager());
+
+        internal Stack<DisposeScope> DisposeScopeStack { get; } = new();
+        internal IDisposable CurrentlyDisposing { get; set; }
+
+        public int CreatedCount => _createdCount;
+        public int DisposedCount => _disposedCount;
+        public int TotalCount => _createdCount - _disposedCount;
```
lostmsu, line 37: NIT: LiveCount ?

src/TorchSharp/DisposeScopeManager.cs
```diff
@@ -0,0 +1,315 @@
...
+
+        public static long TotalAllocatedSize => Singleton.InnerTotalAllocatedSize;
```
lostmsu, line 39: This is misleading, because it does not return total allocated for all threads, but the name does not have any indication of that. Why not just keep it an instance property?

src/TorchSharp/DisposeScopeManager.cs
```diff
@@ -0,0 +1,315 @@
...
+        internal static void TryUnregisterInDisposeScope(IDisposable disposable) =>
+            Singleton.InnerTryUnregisterInDisposeScope(disposable);
+
+        internal static void TryRegisterInDisposeScope(IDisposable disposable) =>
+            Singleton.InnerTryRegisterInDisposeScope(disposable);
+
+        private long InnerTotalAllocatedSize {
+            get {
+                long size = 0;
+                lock (this) {
```
lostmsu, line 50: Why do you need to lock if all instances are thread-local?

src/TorchSharp/DisposeScopeManager.cs
```diff
@@ -0,0 +1,315 @@
...
+                return size;
+            }
+        }
+
+        private void InnerTryRegisterInDisposeScope(IDisposable disposable)
+        {
+            Interlocked.Increment(ref _createdCount);
+
+            lock (this) {
+                if (DisposeScopeStack.Count > 0) {
```
lostmsu, line 65: Not much point to use `Interlocked` if you lock below.

src/TorchSharp/DisposeScopeManager.cs
```diff
@@ -0,0 +1,315 @@
...
+
+                return size;
+            }
+        }
+
+        private void InnerTryRegisterInDisposeScope(IDisposable disposable)
+        {
+            Interlocked.Increment(ref _createdCount);
+
+            lock (this) {
```
lostmsu, line 64: [Avoid locking on `this`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/lock-statement#guidelines)

src/TorchSharp/DisposeScopeManager.cs
```diff
@@ -0,0 +1,315 @@
...
+                if (DisposeScopeStack.Count > 0) {
+                    DisposeScopeStack.Peek().Include(disposable);
+                }
+            }
+        }
+
+        private void InnerTryUnregisterInDisposeScope(IDisposable disposable)
+        {
+            // It's us! We're disposing of it.
+            if (disposable == CurrentlyDisposing) {
```
lostmsu, line 74: Can you describe a scenario when this could happen?

src/TorchSharp/DisposeScopeManager.cs
```diff
@@ -0,0 +1,315 @@
...
+                return;
+            }
+
+            Interlocked.Increment(ref _disposedCount);
+            lock (this) {
+                foreach (var disposeScope in DisposeScopeStack) {
```
lostmsu, line 80: It might make more sense to have Tensor reference its dispose scope, and remove directly from it, instead of looping every time.

src/TorchSharp/DisposeScopeManager.cs
```diff
@@ -0,0 +1,315 @@
...
+                DisposeScopeStack.Pop();
+            }
+        }
+
+        internal void MoveToOuterScope(DisposeScope disposeScope, IDisposable disposable)
+        {
+            lock (this) {
+                var array = DisposeScopeStack.ToArray();
+                for (var i = array.Length - 1; i >= 1; i--) {
+                    if (array[i] == disposeScope) {
```
lostmsu, line 116: Aren't you keeping parent scope in `DisposeScope`? If not, it might be a good idea to do so.

src/TorchSharp/DisposeScopeManager.cs
```diff
@@ -0,0 +1,315 @@
...
+                        array[i - 1].Include(disposable);
+                        return;
+                    }
+                }
+            }
+        }
+
+        public class DisposeScope : IDisposable
```
lostmsu, line 124: Each class should be in its own file.

Also, is this class supposed to be `public`?

src/TorchSharp/DisposeScopeManager.cs
```diff
@@ -0,0 +1,315 @@
...
+
+            public DisposeScope(DisposeScopeManager disposeScopeManager)
+            {
+                _disposeScopeManager = disposeScopeManager;
+            }
+
+            /// <summary>
+            /// The disposables that are scheduled for disposing.
+            /// </summary>
+            public HashSet<IDisposable> Disposables { get; } = new();
```
lostmsu, line 136: I think this set should be using [`ReferenceEqualityComparer`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.referenceequalitycomparer?view=net-6.0)

src/TorchSharp/DisposeScopeManager.cs
```diff
@@ -0,0 +1,315 @@
...
+
+            public long TotalAllocatedSize =>
+                Disposables.OfType<torch.Tensor>().Sum(x => x.NumberOfElements * x.ElementSize);
```
lostmsu, line 139: This is misleading, as multiple tensors can share the same memory.

src/TorchSharp/DisposeScopeManager.cs
```diff
@@ -0,0 +1,315 @@
...
+
+            /// <summary>
+            /// Includes a disposable in the scope - for tensors this is done automatically once the scope has been
+            /// created. Use this method to add additional disposables that should be disposed, but you typically
+            /// don't need to call this method.
+            /// </summary>
+            /// <param name="disposable">The disposable to keep in the scope</param>
+            /// <returns></returns>
+            public T Include<T>(T disposable) where T : IDisposable
```
lostmsu, line 148: Why is this method generic? Why not just accept `IDisposable`?

src/TorchSharp/DisposeScopeManager.cs
```diff
@@ -0,0 +1,315 @@
...
+                Exclude(new IDisposable[] { first, second, third }, false);
+                return (first, second, third);
+            }
+
+            public IDisposable[] Exclude(IDisposable[] exclude, bool excludeGlobally)
+            {
+                lock (this) {
+                    foreach (var disposable in exclude) {
+                        if (Disposables.Contains(disposable)) {
+                            Disposables.Remove(disposable);
```
lostmsu, line 187: `HashSet.Remove` returns `true`/`false` depending if the item was found and removed.
Therefore you don't need tot call `Contains` first.

src/TorchSharp/DisposeScopeManager.cs
```diff
@@ -0,0 +1,315 @@
...
+                        Disposables.Remove(disposable);
+                    } finally {
+                        _disposeScopeManager.CurrentlyDisposing = null;
+                    }
+                }
+            }
+
+            ~DisposeScope()
+            {
+                ReleaseUnmanagedResources();
```
lostmsu, line 311: I think it should be an error to leave a dangling instance of `DisposeScope`. Either you need it and you should eventually call `Dispose` on it, or you don't need it.

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -27,11 +27,22 @@ public partial class Tensor : IDisposable
             static long _totalCount = 0;
             static long _peakCount = 0;
 
+
+            public static List<Action<Tensor>> OnTensorCreateCallback { get; } =
+                new List<Action<Tensor>>();
+
+            public static List<Action<Tensor>> OnTensorDisposeCallback { get; } =
+                new List<Action<Tensor>>();
```
lostmsu, line 35: These should not be public, or they should be events.

Also, name for the events should be `OnTensorCreated` and `BeforeTensorDisposed` correspondingly.

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -2840,6 +2860,14 @@ public Tensor eq_(Scalar target)
...
             public bool Equals(Tensor target)
             {
+                if (handle == IntPtr.Zero || target.handle == IntPtr.Zero) {
+                    return (object)this == (object)target;
+                }
+
+                if (dtype != target.dtype) {
+                    return false;
+                }
+
```
lostmsu, line 2870: As I mentioned above, you should not be hacking `Equals` for your `HashSet`. Instead, use `ReferenceEqualityComparer`.

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -5577,5 +5605,12 @@ public static bool is_complex(ScalarType type)
 
         public static ScalarType cfloat = ScalarType.ComplexFloat32;
         public static ScalarType cdouble = ScalarType.ComplexFloat64;
+
+        /// <summary>
+        /// Creates a new dispose scope for the current thread. Any tensor created within the dispose scope will
+        /// be automatically disposed once the dispose scope is disposed.
+        /// </summary>
+        /// <returns></returns>
```
lostmsu, line 5613: Remove empty doc node

test/TorchSharpTest/TestTorchTensor.cs
```diff
@@ -16,6 +16,12 @@ namespace TorchSharp
 {
     public class TestTensor
     {
+        public TestTensor()
+        {
+            // Avoid comma issues on international machines (Swedish, for instance...)
+            CultureInfo.DefaultThreadCurrentCulture = (CultureInfo)CultureInfo.InvariantCulture.Clone();
+        }
+
```
lostmsu, line 24: This should never be needed. You should fix test code instead.

lostmsu (changes requested):

src/TorchSharp/DisposeScopeManager.cs
lostmsu: If it has to do with finalizers, something needs to be reworked, as finalizers should never call into neither `DisposeScope` nor `DisposeScopeManager`. Perhaps `Tensor` does not implement [the dispose pattern](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#dispose-and-disposebool) correctly.

lostmsu:
General question, what is the scenario for a tensor to belong to multiple dispose scopes?

NiklasGustafsson:
> General question, what is the scenario for a tensor to belong to multiple dispose scopes?

There shouldn't be any __dual__ ownerships. 

Ownership may be transferred to a parent scope, though -- for example, when a function returns a final computed value, disposing of all temporaries (anonymous and those captured in variables), returning only a few (usually one) to the surrounding scope.

test/TorchSharpTest/TestTorchTensor.cs
NiklasGustafsson: The test code, for expediency, expects US numbers formatting in all the ToString() tests.

lostmsu:
> There shouldn't be any dual ownerships.

That's what I thought. Any code, that assumes multiple scopes can have the same tensor should be modified no to do so.

src/TorchSharp/Tensor/Tensor.cs
NiklasGustafsson: The lists shouldn't be publicly accessible -- that would mean that you can remove hooks, or manipulate things in weird ways. There should be functions to register hooks.

test/TorchSharpTest/TestTorchTensor.cs
lostmsu: The culture needs to be passed to `ToString` then.

src/TorchSharp/DisposeScopeManager.cs
NiklasGustafsson: I agree. The finalizer should simply refuse to release any resources. The only reasonable use of DisposeScopes is to put then in a using() {} construct.

NiklasGustafsson:
> > There shouldn't be any dual ownerships.
> 
> That's what I thought. Any code, that assumes multiple scopes can have the same tensor should be modified no to do so.

Is there such code? The only "dual" ownership that I can think of is that a tensor may be controlled by a 'using' statement. That won't prevent it from being registered with a DisposeScope. When the scope is disposed, the tensor is likely to have already been disposed.

Is that what you're thinking about as dual ownership? I don't see a way to avoid it -- we can't tell whether a tensor was created in the context of a 'using'.

lostmsu:
> The only "dual" ownership that I can think of is that a tensor may be controlled by a 'using' statement.

I am referring to implementation in this PR looping through all dispose scopes in the stack to remove the tensor, for example. It should really stop at the first encounter in most cases. Also it might make sense to redesign `Tensor` to reference its `DisposeScope` if only 1 is ever allowed.

NiklasGustafsson:
Got it.

NiklasGustafsson:
BTW, I'm taking time off until November 29th.

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: The only reason is to have a thread safe way of finding out how many tensors have been created on the thread. But that could definitely be moved somewhere else, the tensor class could keep track of it with some thread static magic. 

        [Fact]
        public void DisposeScopesCanBeNestled()
        {
            torch.Tensor data = torch.rand(10, 10);
            var preTotalCount = DisposeScopeManager.Singleton.TotalCount;

            using (torch.NewDisposeScope()) {
                var t1 = data * data + data;

                var innerCount = DisposeScopeManager.Singleton.TotalCount;
                using (torch.NewDisposeScope()) {
                    var t2 = data + data - t1;
                }

                // Inner scope was disposed
                Assert.Equal(DisposeScopeManager.Singleton.TotalCount, innerCount);
            }

            // It was all disposed
            Assert.Equal(DisposeScopeManager.Singleton.TotalCount, preTotalCount);
        }

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: It's because of the params parameter, you can't use IEnumberable<IDisposable> as a params parameter. I'm not sure when  someone would use that fallback, though, so an IEnumerable version might suffice 99.9% of the time.

mfagerlund: I changed both 

public void DisposeEverythingBut(IEnumerable<IDisposable> keep)

and

  public void Exclude(IEnumerable<IDisposable> exclude, bool excludeGlobally)

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: It's because of this naming on Tensor;

     public static long TotalCount => _totalCount;

I figured I'd use the same name?

mfagerlund: I could call it a Threadaton? But you're probably right, I'll call  it ThreadSingleton so that it becomes apparant that it's not really a singleton.

mfagerlund:
> > > It's called from two places, one where it doesn't iterate over the entire list. I didn't want to repeat the code.
> 
> Assuming that the 'keep' list is typically shorter than the non-keep list, you could save the keeps, call DoDispose(), clear the list (or just discard it, letting GC pick it up later), and replace Disposables with the keep list. That **should** mean fewer calls to 'Remove()' and still allow DoDispose() to be called in both places.
> 
> Something like:
> 
> ```cs
>                     var toKeep = keep.ToHashSet();
>                     foreach (var disposable in Disposables) {
>                         if (toKeep.Contains(disposable)) continue;
>                         DoDispose(disposable);
>                     }
>                     Disposables = toKeep;
> ```
> 
> assuming that you add a private setter to 'Disposables'

I did this;

                    var oldList = Disposables;
                    Disposables = keep.ToHashSet();
                    foreach (var disposable in oldList) {
                        if (!Disposables.Contains(disposable)) {
                            DoDispose(disposable, false);
                        }
                    }

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: I removed allocated size, @NiklasGustafsson has mentioned that it's too approximative anyway.

I made the thread "singleton" private, so I had to keep the counts public, but I did this;

        /// <summary>
        /// The number of disposables that are currently live on the current thread. It's aproximate, see
        /// Tensor.TotalCount.
        /// </summary>
        public static longThreadTotalLiveCount => ThreadSingleton._createdCount - ThreadSingleton._disposedCount;

Lemme know  what you think

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: We had  an issue where we suspected the GC thread of doing stuff, but I think that was a mistake. If we're safe to assume that no external thread will every dispose of any of our tensors, the locking is not needed. I'll remove it for now.

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: True!

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: Ok, I'm confused as to whether I need a finalizer or not. I decided that I don't, so in dispose I dispose of all contained disposables and am done with it;

            public void Dispose()
            {
                foreach (var disposable in Disposables)
                {
                    DoDispose(disposable, false);
                }
                Disposables.Clear();
                _disposeScopeManager.RemoveDisposeScope(this);
            }

No suppressing anything. Makes sense?

src/TorchSharp/Tensor/Tensor.cs
mfagerlund: Changed to this;

            internal delegate void TensorEvent(Tensor tensor);
            internal static event TensorEvent? OnTensorCreated;
            internal static event TensorEvent? BeforeTensorDisposed;

            internal Tensor(IntPtr handle)
            {
                this.handle = handle;
                System.Threading.Interlocked.Increment(ref _totalCount);
                _peakCount = Math.Max(_totalCount, _peakCount);
                OnTensorCreated?.Invoke(this);
            }

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: Removed all locks, so now the interlocked makes sense again

src/TorchSharp/Tensor/Tensor.cs
mfagerlund: ok

test/TorchSharpTest/TestTorchTensor.cs
mfagerlund: It's because of tests like below, they  expect there to be a "." as a decimal separator, but I have a "," as a decimal separator.

Did you mean I should flip the culture in every one of the tests that fail (seven) and flip it back after? Or did you have some other idea?

                Assert.Equal($"[2x2x2x4], type = Float32, device = cpu{_sep}{_sep}[0,0,:,:] ={_sep}        0   3.141 6.2834 3.1415{_sep}" +
                             $" 6.28e-06 -13.142   0.01 4713.1{_sep}{_sep}[0,1,:,:] ={_sep} 0.01 0 0 0{_sep}    0 0 0 0{_sep}{_sep}" +
                             $"[1,0,:,:] ={_sep}        0   3.141 6.2834 3.1415{_sep} 6.28e-06 -13.142   0.01 4713.1{_sep}{_sep}" +
                             $"[1,1,:,:] ={_sep} 0.01 0 0 0{_sep}    0 0 0 0{_sep}", str);

mfagerlund:
> > The only "dual" ownership that I can think of is that a tensor may be controlled by a 'using' statement.
> 
> I am referring to implementation in this PR looping through all dispose scopes in the stack to remove the tensor, for example. It should really stop at the first encounter in most cases. Also it might make sense to redesign the PR to have `Tensor` reference its `DisposeScope` if only 1 is ever allowed.

Are you referring to the foreach loop below? Ok, I'll change it to stop on the first scope that contains it, but we can't be sure it's the top scope, so we'll need to loop that far. And most appropriate would be to loop backwards, because upper stacks are more likely to contain the tensor/disposable.

        private void InnerTryUnregisterInDisposeScope(IDisposable disposable)
        {
            // It's us! We're disposing of it.
            if (disposable == CurrentlyDisposing) {
                return;
            }

            Interlocked.Increment(ref _disposedCount);
            foreach (var disposeScope in DisposeScopeStack) {
                disposeScope.Disposables.Remove(disposable);
            }
        }

mfagerlund:
> > The only "dual" ownership that I can think of is that a tensor may be controlled by a 'using' statement.
> 
> I am referring to implementation in this PR looping through all dispose scopes in the stack to remove the tensor, for example. It should really stop at the first encounter in most cases. Also it might make sense to redesign the PR to have `Tensor` reference its `DisposeScope` if only 1 is ever allowed.

If you think it appropriate, I can update the PR to have a reference to the DisposeScope it's currently in scope of. I'll design a suggestion. I wanted to make as few changes as possible in the Tensor class...

src/TorchSharp/Tensor/Tensor.cs
mfagerlund: I reverted this change in the latest version

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: Agreed

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: I've made the singleton internal and moved the count to DisposeScopeManager.ThreadTotalLiveCount

src/TorchSharp/Tensor/Tensor.cs
mfagerlund: I'm no longer using these events, every tensor knows what scope it belongs to and communicates directly with that scope as it's created and disposed.

However, there may be a need for the events for other reasons, should I remove them or leave them in?

mfagerlund:
> I am referring to implementation in this PR looping through all dispose scopes in the stack to remove the tensor, for example. It should really stop at the first encounter in most cases. Also it might make sense to redesign the PR to have `Tensor` reference its `DisposeScope` if only 1 is ever allowed.

I've reworked he system;

Every tensor knows what scope it belongs to, if it belongs to any scope. No more looking for the right scope.

I added methods to the tensor to move it up or detach it;

            public torch.Tensor MoveToOuterDisposeScope()
            public torch.Tensor DetatchFromDisposeScope()

Maybe shorter names would be nice?

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: It was to prevent a loop where us disposing a tensor made the tensor contact us to let us know it was being disposed. Since we already knew we were disposing it, we could save a roundtrip. I've solved it a different way now!

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: You're absolutely right. I was trying to minimize the impact on the Tensor object, now they're more tightly bound but it saves on the looping and the looking around.

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: I am now! It was a great idea

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: fixed

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: That doesn't seem to exist in the version of C# I'm using, it seems like it's a .NET 6. I'll try to use something similar until we've migrated to .NET 6?

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: I've removed it, though I wish there was a way to monitor memory usage.

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: Fixed, where I need to know if it actually existed, I use the return value of "Remove" as to not ask twice.

src/TorchSharp/Tensor/Tensor.cs
mfagerlund: I'm removing them for now, if they're needed/wanted, they should be added in a separate PR. 

I for one would like to have them, though.

test/TorchSharpTest/TestTorchTensor.cs
mfagerlund: I missed the conversation above, I'll fix this by passing in to string

mfagerlund:
Trying again with a new PR, thanks for your patience

src/TorchSharp/DisposeScopeManager.cs
lostmsu: Maybe `ForCurrentThread`

src/TorchSharp/DisposeScopeManager.cs
lostmsu: @mfagerlund it should be trivial to reimplement using [`RuntimeHelpers.GetHashCode`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.runtimehelpers.gethashcode) and `object.ReferenceEquals`

See (actual implementation in .NET 5.0+)[https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/libraries/Common/src/System/Collections/Generic/ReferenceEqualityComparer.cs]

src/TorchSharp/DisposeScopeManager.cs
lostmsu: I think it won't hurt to have a finalizer that would throw `InvalidOperationException` unconditionally for debugging purposes.

lostmsu:
> I did this; `Disposables = keep.ToHashSet(); ...`

I believe in any reasonable scenario the keep list will be short enough for `Contains` method to be faster than building a hashset.

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: Fixed with

    /// <summary>
    /// Keeps track of statistics for a dispose scope. Can be queried to figure out performance/memory issues.
    /// </summary>
    public class ThreadDisposeScopeStatistics
    {
        /// <summary>
        /// The number of disposables that were created on this thread, but weren't captured by a DisposeScope.
        /// </summary>
        public long CreatedOutsideScopeCount { get; internal set; }

        /// <summary>
        /// The number of disposables that were created on this thread and were captured by a DisposeScope.
        /// </summary>
        public long CreatedInScopeCount { get; internal set; }

        /// <summary>
        /// The number of disposables that were disposed on this thread and were disposed while in a DisposeScope.
        /// </summary>
        public long DisposedInScopeCount { get; internal set; }

        /// <summary>
        /// Number of disposables that were once included in the scope, but were subsequently detached.
        /// </summary>
        public long DetachedFromScopeCount { get; internal set; }

        /// <summary>
        /// The number of disposables that are currently live on the current thread. If a It's aproximate, see
        /// Tensor.TotalCount. Disposables that weren't created within a DisposeScope, or detached from the dispose
        /// scope, will not be counted as alive.
        /// </summary>
        public long ThreadTotalLiveCount => CreatedInScopeCount - DisposedInScopeCount - DetachedFromScopeCount;

        /// <summary>
        /// Resets the counts for the current thread. See ThreadTotalLiveCount etc. Mainly used in tests to make sure
        /// we get a clean slate on the thread.
        /// </summary>
        public void Reset()
        {
            CreatedOutsideScopeCount = 0;
            CreatedInScopeCount = 0;
            DisposedInScopeCount = 0;
            DetachedFromScopeCount = 0;
        }
    }

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: ok

src/TorchSharp/DisposeScopeManager.cs
mfagerlund: Yep, I found a similar one that I'm using now.

PR #463, opened 2021-11-27
Add pytorch style dataloader

dayo05:
In pytorch there are DataLoader on torch.utils.data, Torchsharp has DataIterator on TorchSharp.Data.

In pytorch, Creating Dataset is easy.
```python
class CustomImageDataset(Dataset):
    def __init__(self, annotations_file, img_dir, transform=None, target_transform=None):
        self.img_labels = pd.read_csv(annotations_file)
        self.img_dir = img_dir
        self.transform = transform
        self.target_transform = target_transform

    def __len__(self):
        return len(self.img_labels)

    def __getitem__(self, idx):
        img_path = os.path.join(self.img_dir, self.img_labels.iloc[idx, 0])
        image = read_image(img_path)
        label = self.img_labels.iloc[idx, 1]
        if self.transform:
            image = self.transform(image)
        if self.target_transform:
            label = self.target_transform(label)
        return image, label
```
This is sample code that I found on pytorch official tutorial.
But in DataIterator, I don't know how to use that, It is not similar as pytorch style, I cannot found tutorial about it.

So I make dataloader using like this.
This style is very similar as what pytorch does.

This will make iterating data more easier.


nietras:
This is "hard-coded" to only two tensors which doesn't fit general case scenarios like what we have. In python this is handled by the dynamic type system and being able to handle tuples etc. This cannot be mirrored with a good experience in C# in my view. The best generalization is to base this on `Dictionary<string, Tensor>` (or `Dictionary<object, Tensor>`) if generalizing key.

Additionally, `Dataset` alone is not in itself useful, it needs to be designed together with `DataLoader`, which is then responsible for batching, randomization etc. And since we need to batch we need to be able to iterator over `Tensor`s and produce the same structured output, which is why `Dictionary<string, Tensor>` seems apt.

src/TorchSharp/Data/Dataset.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using static TorchSharp.torch;
+
+namespace TorchSharp.Data
+{
+    public interface Dataset: IDisposable
+    {
+        public long Count();
+        public (Tensor, Tensor) GetTensor(object index);
+        public IEnumerable<object> GetDataEnumerable();
```
nietras, line 12: remove "public IEnumerable<object> GetDataEnumerable();" this should not be part of interface

src/TorchSharp/Data/Dataset.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+using System.Collections.Generic;
+
+using static TorchSharp.torch;
+
+namespace TorchSharp.Data
+{
+    public interface Dataset: IDisposable
+    {
+        public long Count();
```
nietras, line 10: should be property `int Count {get; }`

src/TorchSharp/Data/Dataset.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+using System.Collections.Generic;
+
+using static TorchSharp.torch;
+
+namespace TorchSharp.Data
+{
+    public interface Dataset: IDisposable
```
nietras, line 8: `IDataset` maybe?

src/TorchSharp/Data/Dataset.cs
```diff
@@ -0,0 +1,14 @@
...
+    {
+        public long Count();
+        public (Tensor, Tensor) GetTensor(object index);
```
nietras, line 11: `Dictionary<string, Tensor> this[int index]`

src/TorchSharp/Data/DataLoader.cs
```diff
@@ -0,0 +1,96 @@
...
+            private Device device;
+            public DataLoaderEnumerator(Dataset dataset, int batchSize, bool shuffle, Device device)
+            {
+                this.dataset = dataset;
+                this.batchSize = batchSize;
+                this.device = device;
+
+                if (shuffle)
+                {
+                    var r = new Random();
```
nietras, line 50: move to ctor and create seeded from ctor parameter

src/TorchSharp/Data/DataLoader.cs
```diff
@@ -0,0 +1,96 @@
...
+                    load = dataset.GetDataEnumerable()
+                        .Select(x => new {Number = r.Next(), Item = x})
+                        .OrderBy(x => x.Number)
+                        .Select(x => x.Item)
+                        .GetEnumerator();
```
nietras, line 55: create an array of indexes from data set e.g. `Enumerable.Range(0, dataset.Count).ToArray()` then fischer yates shuffle that and keep this array of shuffled indexes as what will be iterated over, order by random is not a great way of doing shuffling. 

Fischer-Yates shuffle code is simple:
```csharp
       // Fisher-Yates shuffling
        internal static void Shuffle<T>(T[] values, Random random)
        {
            int n = values.Count;
            while (n > 1)
            {
                int k = random.Next(n);
                --n;
                T temp = values[n];
                values[n] = values[k];
                values[k] = temp;
            }
        }
```

nietras: If not shuffling simply use indexes without shuffling, this means you can remove the GetDataEnumerable, it also fits better in potential multi-threading and pre-loading.

src/TorchSharp/Data/DataLoader.cs
```diff
@@ -0,0 +1,96 @@
...
+                }
+                else load = dataset.GetDataEnumerable().GetEnumerator();
+            }
+
+            private Tensor dataTensor;
+            private Tensor labelTensor;
+            private (Tensor, Tensor) tmp;
```
nietras, line 62: remove these members, there are not necessary

nietras (commented):
I've made some review comments too. :) I have more but just some quick remarks.

src/TorchSharp/Data/DataLoader.cs
mfagerlund: There are ways of visiting every index in "random" order, without enumerating the list up front;

https://lemire.me/blog/2017/09/18/visiting-all-values-in-an-array-exactly-once-in-random-order/
https://lemire.me/blog/2017/09/26/benchmarking-algorithms-to-visit-all-values-in-an-array-in-random-order/

Something like that would be better than the Fisher-Yates shuffle, especially if you're processing very large numbers of samples.

src/TorchSharp/Data/Dataset.cs
dayo05: I did like that because pytorch does.

src/TorchSharp/Data/DataLoader.cs
dayo05: dataTensor, labelTensor is required for return current value.

dayo05: I change that to directly change Current value.

src/TorchSharp/Data/DataLoader.cs
dayo05: I make private reset method and call that on ctor.

dayo05:
TODO: Write test about it

src/TorchSharp/Data/Dataset.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
+using System;
+using System.Collections.Generic;
+using static TorchSharp.torch;
+
+namespace TorchSharp.Data
```
NiklasGustafsson, line 6: This should be in 'TorchSharp.torch.utils.data'

src/TorchSharp/Data/Dataset.cs
NiklasGustafsson: IDataset is right for a .NET interface. 

The 'Dataset' class is abstract in PyTorch, so please change to an abstract class, or change the name.

NiklasGustafsson (changes requested):

NiklasGustafsson:
This is such an important addition, and we must make sure to get it right. I would like to see some more extensive code examples on how to use this -- the user experience is essential.

dayo05:
I'm making a new example for reflection about review

NiklasGustafsson:
I'm going to add a few things to the code comments in a review, but I have some questions:

1. This will always shuffle the data set. Are there situations when you do not want to shuffle?
2. Should we allow data loaders to accept custom shufflers? That is, if I think (I'd probably be wrong) that I have a better shuffle algorithm, shouldn't I be allowed to use that, instead?
3. There's no way to specify the device on which the 'Current' tensors end up.
4. Why does 'Current' return a dictionary? Is it input vs. labels?
5. How do I split a data set into train, test, and validation data subsets?
6. How do we add data augmentation support?

Also, it would be great to demonstrate how to use this by converting the Examples to use this instead of the ad hoc loading I added earlier.

src/TorchSharp/DataLoader.cs
```diff
@@ -0,0 +1,114 @@
...
+
+namespace TorchSharp
+{
+    public static partial class torch
+    {
+        public static partial class utils
+        {
+            public static partial class data
+            {
+                public class DataLoader : IEnumerable<Dictionary<string, Tensor>>, IDisposable
```
NiklasGustafsson, line 17: Please add doc comments to all classes and (at least) all non-private members. You can copy information from the Pytorch docs, if there is any.

src/TorchSharp/DataLoader.cs
```diff
@@ -0,0 +1,114 @@
...
+                {
+                    private Dataset dataset;
+                    private int batchSize;
+                    private bool shuffle;
+                    private Device device;
+
+                    /// <summary>
+                    /// Create pytorch style dataloader
+                    /// </summary>
+                    /// <param name="dataset"></param>
```
NiklasGustafsson, line 27: Fill in the param elements.

src/TorchSharp/DataLoader.cs
```diff
@@ -0,0 +1,114 @@
...
+                            Current = dataset.GetTensor(getNextValue());
+                            var currentKeys = Current.Keys;
+                            foreach (var x in currentKeys)
+                                Current[x].unsqueeze_(0);
+                            Dictionary<string, Tensor> dic;
+                            for (var i = 1; i < batchSize; i++) {
+                                if (isFinished())
+                                    break;
+                                dic = dataset.GetTensor(getNextValue());
+                                foreach (var x in currentKeys)
```
NiklasGustafsson, line 87: Wouldn't it be better to gather all the unsqueezed tensors in a list (or array) and then catenate them all together at the end. It would seem a lot more efficient, and will create fewer temporaries.

src/TorchSharp/DataLoader.cs
```diff
@@ -0,0 +1,114 @@
...
+
+                        private int getNextValue() => shuffle ? shuffleGenerator.next() : currentVal++;
+
+                        private void reset()
+                        {
+                            shuffleGenerator = new ShuffleGenerator(dataset.Count);
+                            currentVal = 0;
+                        }
+
+                        public bool MoveNext()
```
NiklasGustafsson, line 75: This will create a lot of temporaries. Where are they disposed?

src/TorchSharp/DataLoader.cs
```diff
@@ -0,0 +1,114 @@
...
+                        {
+                            if (isFinished()) return false;
+                            Current = dataset.GetTensor(getNextValue());
+                            var currentKeys = Current.Keys;
+                            foreach (var x in currentKeys)
+                                Current[x].unsqueeze_(0);
```
NiklasGustafsson, line 81: Couldn't this be done with a list or a tuple instead of a dictionary? What are the keys?

src/TorchSharp/Dataset.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace TorchSharp
+{
+    public static partial class torch
+    {
+        public static partial class utils
+        {
+            public static partial class data
+            {
+                public abstract class Dataset : IDisposable
```
NiklasGustafsson, line 13: Doc comments, please.

src/TorchSharp/Utils/ShuffleGenerator.cs
```diff
@@ -0,0 +1,89 @@
+using System;
```
NiklasGustafsson, line 1: Copyright notice, please.

src/TorchSharp/Utils/ShuffleGenerator.cs
```diff
@@ -0,0 +1,89 @@
...
+{
+    public class ShuffleGenerator
+    {
+        int maxrange;
+        int prime;
+        int index;
+        int offset;
+        int runningvalue;
+
+        public ShuffleGenerator(int size)
```
NiklasGustafsson, line 13: Please add a reference to where the algorithm comes from, even if it's just a Pytorch source file. A paper describing it would be best, but some attribution is important.

src/TorchSharp/Utils/ShuffleGenerator.cs
```diff
@@ -0,0 +1,89 @@
+using System;
+
+namespace TorchSharp.Utils
+{
+    public class ShuffleGenerator
```
NiklasGustafsson, line 5: Should this really be public? I'm assuming that it's not really intended to be used outside of TorchSharp, but I could be wrong.

src/TorchSharp/Utils/ShuffleGenerator.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            var min = size / 2;
+            maxrange = size;
+            prime = selectCoPrimeResev(min, size);
+            offset = new Random().Next(size);
+            index = 0;
+            runningvalue = offset;
+        }
+
+        private int getCurrentValue()
```
NiklasGustafsson, line 23: This seems like Java-style naming. If this is a public Pytorch-based API, then use Python names. If it is not, then please use .NET naming conventions.

src/TorchSharp/Utils/ShuffleGenerator.cs
```diff
@@ -0,0 +1,89 @@
...
+
+                if (count == MAX_COUNT) return val;
+            }
+
+            return selected;
+        }
+
+        static bool coprime(int u, int v) => gcd(u, v) == 1;
+
+        static int gcd(int u, int v)
```
NiklasGustafsson, line 64: Is this Greatest Common Divisor? I don't know much about which is the most efficient way to compute it, but this looks a lot more complex than some of the algorithms I found on stackoverflow:

https://stackoverflow.com/questions/18541832/c-sharp-find-the-greatest-common-divisor

src/TorchSharp/Utils/ShuffleGenerator.cs
```diff
@@ -0,0 +1,89 @@
...
+        int index;
+        int offset;
+        int runningvalue;
+
+        public ShuffleGenerator(int size)
+        {
+            var min = size / 2;
+            maxrange = size;
+            prime = selectCoPrimeResev(min, size);
+            offset = new Random().Next(size);
```
NiklasGustafsson, line 18: We should probably allow a user to pass in a custom seed for Random().

NiklasGustafsson, line 18: Some of my torchvision code also relies on .NET Random() for RNG, but I'm wondering whether that is right. Maybe using Pytorch's RNG is the right thing? Not sure.

src/TorchSharp/DataLoader.cs
```diff
@@ -0,0 +1,114 @@
...
+
+                        public DataLoaderEnumerator(Dataset dataset, int batchSize, bool shuffle, Device device)
+                        {
+                            this.dataset = dataset;
+                            this.batchSize = batchSize;
+                            this.device = device;
+                            this.shuffle = shuffle;
+                            reset();
+                        }
+
```
NiklasGustafsson, line 63: Java-style names don't make sense here.

src/TorchSharp/Dataset.cs
```diff
@@ -0,0 +1,26 @@
...
+                {
+                    public virtual void Dispose()
+                    {
+                    }
+
+                    public abstract int Count { get; }
```
NiklasGustafsson, line 19: Since data sets can be enormous, Count and GetTensor (and anyone calling them) should probably use 'long' instead of 'int'. It's not often that you need more than 32-bit int for indices, but I think this is one of those rare situations.

The same goes for all other places in the code where the integer is an index into the data set.

src/TorchSharp/DataLoader.cs
```diff
@@ -0,0 +1,114 @@
...
+                    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+
+                    public long Count => (dataset.Count - 1) / batchSize + 1;
+
+                    private class DataLoaderEnumerator : IEnumerator<Dictionary<string, Tensor>>
+                    {
+                        private Dataset dataset;
+                        private int batchSize;
+                        private Device device;
```
NiklasGustafsson, line 50: 'device' is never used.

NiklasGustafsson (changes requested):

src/TorchSharp/DataLoader.cs
dayo05: Keys are defined by user, I use dictionary to get ambiguous amount of tensor.

NiklasGustafsson:
@dayo05 -- thanks for all the work so far. I'll be taking some time off in December, but I'm looking forward to integrating this when I come back.

Thanks,

Niklas

dayo05:
I'm busy now, I'll going to work on two weeks later :(

src/TorchSharp/Utils/ShuffleGenerator.cs
dayo05: I saw pytorch code and they are using numpy random, python random, std(in C++) random, etc. So I think .NET random will doesn't matter

dayo05:
@NiklasGustafsson I finished all my job, review please.

dayo05:
I found shuffler is not correctly working. I'll fix ASAP.

dayo05:
This is example for classifying fruits360 dataset
```cs
public class Fruits360: Dataset
{
    private List<string> Labels = new();
    private List<string> images = new();
    public Fruits360(bool isTrain, Device device)
    {
        var root = "/home/dayo/datasets/fruits-360/" + (isTrain ? "Training" : "Test");
        //Labels.AddRange(Directory.GetDirectories(root));
        foreach(var x in Directory.GetDirectories(root))
            images.AddRange(Directory.GetFiles(x));
        Labels.AddRange(Directory.GetDirectories(root).Select(x => x.Split('/')[^1]));
    }

    public override long Count => images.Count;

    public override Dictionary<string, torch.Tensor> GetTensor(long index)
    {
        var image = Image.Load<Rgb24>(images[(int) index], new JpegDecoder());
        using var r = tensor(image.GetPixelMemoryGroup()[0].Span.ToArray().Select(x => x.R / 255.0f).ToList(),
            new long[] {1, 100, 100});
        using var g = tensor(image.GetPixelMemoryGroup()[0].Span.ToArray().Select(x => x.G / 255.0f).ToList(),
            new long[] {1, 100, 100});
        using var b = tensor(image.GetPixelMemoryGroup()[0].Span.ToArray().Select(x => x.B / 255.0f).ToList(),
            new long[] {1, 100, 100});
        return new()
        {
            {"image", cat(new List<Tensor> {r, g, b}, 0)},
            {"label", tensor(Labels.IndexOf(images[(int)index].Split('/')[^2]), ScalarType.Int64)}
        };
    }
}
```
```cs
using var trainDataset = new Fruits360(true, CUDA);
using var testDataset = new Fruits360(false, CUDA);
using var train = new DataLoader(trainDataset, 256, true, CUDA);
using var test = new DataLoader(testDataset, 512, false, CUDA);

var model = new Fruits360Model(CUDA);
var optimizer = optim.Adam(model.parameters(), learningRate: 0.01);

foreach (var epoch in Range(1, 1000))
{
    model.Train();
    var batchId = 1;
    Console.WriteLine($"Epoch{epoch} running");
    foreach (var x in train)
    {
        optimizer.zero_grad();

        var prediction = model.forward(x["image"]);
        var output = functional.nll_loss(reduction: Reduction.Mean)(prediction, x["label"]);
        
        output.backward();
        optimizer.step();
        
        Console.Write($"\rTrain: epoch {epoch} {batchId * 1.0 / train.Count:P2} [{batchId} / {train.Count}] Loss: {output.ToSingle():F9}");
        batchId++;
        
        prediction.Dispose();
        output.Dispose();
        GC.Collect();
    }

    using (no_grad())
    {
        model.Eval();
        
        var testLoss = 0.0;
        var correct = 0;
        var total = 0L;
        var idx = 0;
        foreach (var x in test)
        {
            idx++;
            Console.Write($"\rTest running: {idx * 1.0 / test.Count:P2}");
            var prediction = model.forward(x["image"]);
            var output = functional.nll_loss(reduction: Reduction.Sum)(prediction, x["label"]);
            testLoss += output.ToSingle();

            var pred = prediction.argmax(1);
            total += pred.size()[0];
            correct += pred.eq(x["label"]).sum().ToInt32();
            pred.Dispose();
            prediction.Dispose();
            output.Dispose();
            GC.Collect();
        }
        Console.WriteLine(
            $"\rTest set: Average loss {(testLoss / testDataset.Count):F9} | Accuracy {((double) correct / testDataset.Count):P2}");
    }
}


class Fruits360Model : Module
{
    private Module layer1 = Sequential(
        Conv2d(3, 32, 3),
        ReLU(),
        MaxPool2d(2, 2));

    private Module layer2 = Sequential(
        Conv2d(32, 64, 3),
        ReLU(),
        MaxPool2d(2, 2));

    private Module layer3 = Sequential(
        Conv2d(64, 64, 3),
        ReLU(),
        MaxPool2d(2, 2));

    private Module fc = Sequential(
        Flatten(),
        Linear(6400, 1024),
        ReLU(),
        Dropout(),
        Linear(1024, 625),
        ReLU(),
        Linear(625, 131));
    public Fruits360Model(Device? device) : base("fruits360")
    {
        RegisterComponents();
        to(device ?? CPU);
    }

    public override Tensor forward(Tensor t)
    {
        t = layer1.forward(t);
        t = layer2.forward(t);
        t = layer3.forward(t);
        t = fc.forward(t);
        return LogSoftmax(0).forward(t);
    }
}
```

dayo05:
I found that this shuffler is depend on seed. So, I'll change to other shuffler and enable to use custom shuffler with implementation of IEnumerable.

NiklasGustafsson:
@dayo05 -- A couple of comments:

1. This looks really good now. I will be happy to approve this later.
2. There seems to be some problem with the Azure pipelines that are doing the builds for MacOS. We'll have to wait for that to go away before merging.
3. Before we merge, I would like to see at least one of the examples in this repo modified to use this API instead of the version I put together for temporary purposes.

dayo05:
I make fisher yates shuffler as default because I thought that default value is friendly for beginner and previous shuffler is not good for beginner who has less amount of dataset. But, I left previous shuffler and allow to call like this
```cs
var data = new DataLoader(dataset, batchSize, new FastShuffler(dataset.Count), torch.CUDA);
```

dayo05:
https://github.com/dayo05/DataLoaderExample/tree/master
I write example usage for this commit

NiklasGustafsson:
> https://github.com/dayo05/DataLoaderExample/tree/master I write example usage for this commit See Fruits360.cs, Program.cs

That's great. Could you take the MNIST example in this repo and convert it to use your API?

dayo05:
@NiklasGustafsson  I updated here https://github.com/dayo05/DataLoaderExample/tree/master
Exist on MNIST.cs, MNISTReader.cs
I edited from torchsharp main repo source

dayo05:
@NiklasGustafsson I fix it

NiklasGustafsson:
Okay, I'm going to merge this now. Let's still get some of the examples in this repository using the DataLoader API.

PR #466, opened 2021-11-28
Tensor : SafeHandle

nietras:
DRAFT changes to discuss whether Tensor (and all other native handles) should not be based on SafeHandle. See link for why.
https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.safehandle?view=net-6.0

### DIFF
--- a/src/TorchSharp/Tensor/Tensor.cs
+++ b/src/TorchSharp/Tensor/Tensor.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Text;
+using Microsoft.Win32.SafeHandles;
 
 #nullable enable
 namespace TorchSharp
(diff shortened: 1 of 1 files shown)


NiklasGustafsson:
Fails on all builds with: 

```
src/TorchSharp/NN/Module.cs(428,74): error CS0122: 'SafeHandle.handle' is inaccessible due to its protection level [/__w/1/s/src/TorchSharp/TorchSharp.csproj]
```

This is just one of several locations with the same error.

lostmsu:
@nietras I am sorry, where's the "why" part in that link?

The rest of .NET classlib does not have common classes inherit from `SafeHandle` directly or indirectly. Instead, an instance of handle is usually internal to implementation (e.g. `FileStream` is not a `SafeHandle`, but contains one instead).

This makes sense also as a separation between public interface and internal implementation details.

NiklasGustafsson:
I'm not at all sold on this. Why is this better than what we already have in place?

NiklasGustafsson:
@nietras -- are you still of the opinion that this change should be made?

PR #467, opened 2021-11-30
Use the new DisposeScope pattern in optimizers and examples.

NiklasGustafsson:


docfx/articles/memory.md
```diff
@@ -140,6 +140,69 @@ Rather than passing tensor arguments between neural network layers inside a cust
...
+👎 You don't have fine-grained control over when each tensor is reclaimed. All tensors created while a scope is in effect are disposed at once.
+
+👎 It's a new code pattern, not widely used with other libraries.
+
+Let's look at an example:
+
+```C#
+using (var d = torch.NewDisposeScope()) {
+
+    total_acc += (predicted_labels.argmax(1) == labels).sum().cpu().item<long>();
```
lostmsu, line 160: I am curious about this example. In C# `total_acc += x` is the same as `total_acc = total_acc + x`, which in TorchSharp will:
- create a new instance of `Tensor` for `total_acc + x`
- assign it to `total_acc` rendering the old `Tensor` previously stored in the var unreachable
- from my understanding of `NewDisposeScope` leaving the scope of `using` will dispose the new `Tensor` that is now stored in `total_acc`

lostmsu (changes requested):

docfx/articles/memory.md
NiklasGustafsson: Good catch. Yes, it has to be excluded from disposal. The example is too simplistic.

NiklasGustafsson: Of course, in this particular example, total_acc is an integer, not a tensor...

lostmsu: Ooops, did not notice the last op was data access.

PR #480, opened 2021-12-01
Adding OneCycleLR and CyclicLR schedulers

NiklasGustafsson:


NiklasGustafsson:
@nietras, since we discussed OneCycleLR offline: if you have a couple of minutes to look over the OneCycleLR implementation, that would be great. If not, then don't worry about it.

PR #484, opened 2021-12-02
Modifying Module.save to copy instead of moving the tensor.

NiklasGustafsson:


src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -724,6 +725,22 @@ public Tensor to_type(ScalarType type)
             /// </summary>
             public Tensor type_as(Tensor tensor) => to_type(tensor.dtype);
 
+            [DllImport("LibTorchSharp")]
+            static extern IntPtr THSTensor_set_(IntPtr tensor, IntPtr source);
+
+            /// <summary>
+            /// Overwrite an existing tensor with the concents of another tensor.
```
lostmsu, line 732: typo: contents

src/TorchSharp/Tensor/TensorExtensionMethods.cs
```diff
@@ -105,13 +105,22 @@ internal static bool IsFloatingPoint(this ScalarType type)
         /// <param name="writer"></param>
         public static void Save(this Tensor tensor, System.IO.BinaryWriter writer)
         {
+            bool copied = false;
+
+            if (tensor.device_type != DeviceType.CPU) {
+                tensor = tensor.to(torch.CPU);
```
lostmsu, line 111: Looking at the docs [`tensor->to`](https://pytorch.org/cppdocs/api/classat_1_1_tensor.html#_CPPv4NK2at6Tensor2toEN3c108optionalIN2at10ScalarTypeEEEN3c108optionalIN2at6LayoutEEEN3c108optionalIN2at6DeviceEEEN3c108optionalIbEEbbN3c108optionalIN2at12MemoryFormatEEE) might might be moving tensor by default.

Can you add a test to verify GPU tensor remains on GPU after this call?

lostmsu: BTW, might make sense to expose that `copy` parameter to C# too to make its default value clear.

NiklasGustafsson: The copy parameter only affects the semantics when the device and type is already the same as the intended target, so in this case, it should not affect anything.

That said, it might be useful to expose the parameter in the .NET API for 'to'.

NiklasGustafsson: > Can you add a test to verify GPU tensor remains on GPU after this call?

TestSaveLoadGruOnCUDA() should already be doing that.

lostmsu (approved):

PR #498, opened 2022-01-25
Fixed yield statement in Data interface of CIFARReader

Leo-LiHao:
The example of using CIFARReader does not use any `transforms`, so the returned `IEnumerable<(Tensor, Tensor)>` by `Data()` interface will only return the raw `data` and `labels`. However, with any given transforms, `Data()` could return raw data + each transformed data. Not sure whether I dispose the original data correctly.

### DIFF
--- a/src/Examples.Utils/CIFARReader .cs
+++ b/src/Examples.Utils/CIFARReader .cs
@@ -96,7 +96,7 @@ namespace TorchSharp.Examples
         }
 
         public int Size { get {
-                return _size * (_transforms.Count + 1);
+                return _size;
             } }
         private int _size = 0;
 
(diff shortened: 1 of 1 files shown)


src/Examples.Utils/CIFARReader .cs
```diff
@@ -108,11 +108,11 @@ private int ReadSingleFile(string path, int batch_size, bool shuffle, Device dev
         public IEnumerable<(Tensor, Tensor)> Data()
         {
             for (var i = 0; i < data.Count; i++) {
-                yield return (data[i], labels[i]);
-
+                var tfdata = data[i];
                 foreach (var tfrm in _transforms) {
-                    yield return (tfrm.forward(data[i]), labels[i]);
```
NiklasGustafsson, line 114: The idea of this example code was actually to separately apply all transforms that are passed in, not pipeline them together like the modified code does.

NiklasGustafsson:
This is going to be obsoleted when CIFAR10 is added to the torchvision data sets, which is coming soon.

Issue #499, opened 2022-01-26
Setting Linear.weight is not reflected in 'parameters()'


NiklasGustafsson:
When the 'weight' property of a linear layer (and probably others, too) is set, the new tensor is not reflected in the parameters list.

NiklasGustafsson:
Debugging it, it seems like setting the 'weight' field in the native code implementation of Linear does not alter what its 'parameters()' function returns.

This suggests that we will have to re-implement the Module.parameters() function in C# to not rely on native code, instead assembling the parameters from each module in .NET code. I need to think a bit about that.

Issue #500, opened 2022-01-26
BatchNorm1d throws exception during eval with batch size of 1


FusionCarcass:
I have a model that uses BatchNorm1d after a Linear layer that results in the exception below when using a batch size of 1 during eval. The same operation works fine in Python with a batch size of 1 during eval. I believe the correct behavior here should be to allow batch sizes of 1 during eval mode.

```
Unhandled exception: System.Runtime.InteropServices.ExternalException (0x80004005): Expected more than 1 value per channel when training, got input size [1, 1024]
Exception raised from batch_norm at ..\..\torch\csrc\api\include\torch/nn/functional/batchnorm.h:27 (most recent call first):
00007FFB94A5A29200007FFB94A5A230 c10.dll!c10::Error::Error [<unknown file> @ <unknown line number>]
00007FFB94A59D1E00007FFB94A59CD0 c10.dll!c10::detail::torchCheckFail [<unknown file> @ <unknown line number>]
00000296EC14AE2D00000296EC14AD40 torch_cpu.dll!torch::nn::RNNOptions::batch_first [<unknown file> @ <unknown line number>]
00000296EC19EE2A00000296EC19EC60 torch_cpu.dll!torch::nn::BatchNormImplBase<2,torch::nn::BatchNorm2dImpl>::forward [<unknown file> @ <unknown line number>]
00007FFB903790C400007FFB90379030 LibTorchSharp.DLL!THSNN_BatchNorm1d_forward [<unknown file> @ <unknown line number>]
00007FFAFBADD15D <unknown symbol address> !<unknown symbol> [<unknown file> @ <unknown line number>]

   at TorchSharp.torch.CheckForErrors()
   at TorchSharp.Modules.BatchNorm1d.forward(Tensor tensor)
   at TorchSharp.Modules.Sequential.forward(Tensor tensor)
```

**Troubleshooting**

I have verified that all BatchNorm1d modules are set to eval prior to prediction by running the following command:

```csharp
InceptionResNetV4 module = new InceptionResNetV4();
module.load(fileName);
module.Eval();
module.apply((Module child) => { child.Eval(); });
```

**Theories**

I am training in python, then exporting the model to a .dat file and loading with the module.load method. I verified that the mean and var parameters of the BatchNorm1d are successfully loaded. I thought maybe the mean or var would be null, which might trigger a batching requirement during eval.

I think the issue is either related to loaded from a .dat file OR something not getting passed down to the native torch libraries correctly.

```
running_mean
[...]

PR #501, opened 2022-01-31
Fixed bugs #499 and #500

NiklasGustafsson:
This extensive change to the code addresses a couple of bugs. #499 was the bigger one, requiring the code to move maintenance of module parameters, buffers, and submodules into managed code.

Basically, when setting a parameter property such as Linear.weight, the named_parameters collection was not updated. This meant that an optimizer would not pick up the new tensor, leading to errors.

A breaking change in this PR is that named_parameters() and parameters() both return IEnumerable instances instead of arrays. This is more flexible, and also more in line with what PyTorch does.

Further, BatchNorm{123]d did not include running_mean or running_var in the state_dict() results, nor did it allow those buffers to be accessed or set.

For #500, the problem was that eval() and train() were not properly propagated to all submodules. This manifested itself as a bug to BatchNorm1d, which does not allow batch_size=1 during training, but does during evaluation. Here, too, the solution was to do the propagation in managed code rather than in the underlying native code.

** DO NOT MERGE: once the PR is accepted, I will bump the version number.

### DIFF
--- a/src/TorchSharp/NN/Module.cs
+++ b/src/TorchSharp/NN/Module.cs
@@ -260,7 +260,7 @@ namespace TorchSharp
                 /// <returns></returns>
                 public virtual Module apply(Action<Module> fn)
                 {
-                    foreach (var m in GetModulesInternal()) m.apply(fn);
+                    foreach (var (_,m) in _internal_submodules) m.apply(fn);
                     fn(this);
                     return this;
                 }
@@ -632,29 +632,6 @@ namespace TorchSharp
                             _internal_buffers.Add(name, value);
                         }
                     }
-                }
-
-                [DllImport("LibTorchSharp")]
-                private static extern long THSNN_Module_children_size(HType module);
-
-                [DllImport("LibTorchSharp")]
-                private static extern IntPtr THSNN_Module_child(HType module, int index);
-
-                /// Get the sub-modules of this module. The Module objects won't have the correct .NET types
-                /// so this is not made public.
-                internal virtual Module[] GetModulesInternal()
-                {
-                    var numModules = THSNN_Module_children_size(handle);
-                    torch.CheckForErrors();
-                    Module[] result = new Module[numModules];
-
-                    for (int i = 0; i < numModules; i++) {
-                        var childHandle = THSNN_Module_child(handle, i);
-                        torch.CheckForErrors();
-                        result[i] = new Module(childHandle, null, ownsHandle: false);
-                    }
-
-                    return result;
                 }
 
                 [DllImport("LibTorchSharp")]
--- a/src/TorchSharp/NN/Module.cs
+++ b/src/TorchSharp/NN/Module.cs
@@ -26,7 +26,7 @@ namespace TorchSharp
             /// to fields will be registered, and will have their parameters converted and moved when you call to(),
             /// and saved to disk when calling save().
             /// </remarks>
-            public class Module : DynamicObject, IDisposable
+            public class Module : IDisposable
             {
                 /// <summary>
                 ///    Class wrapping PyTorch's module object reference.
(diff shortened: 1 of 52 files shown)


src/TorchSharp/NN/Module.cs
```diff
@@ -129,6 +139,54 @@ public Module to(DeviceType deviceType, int deviceIndex = -1)
                         THSNN_Module_to_device(handle, (int)deviceType, deviceIndex);
                         torch.CheckForErrors();
 
+                        foreach (var (_, sm) in named_children()) sm.to(deviceType, deviceIndex);
+
+                        foreach (var field in this.GetType().GetFields(BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Instance).Where(f => !f.Name.Contains("__BackingField"))) {
```
dsyme, line 144: This is very uncomfortable, and will perhaps only work for C#.  I need to understand better why we'd be resorting to such things.

dsyme: So whatever is going on here needs to be documented properly, that is, what does it actually mean for the programming model for Modules?

NiklasGustafsson: Thanks!

Specifically, what are you uncomfortable with? The "BackingField" string comparison filter, or something else?

dsyme:
> requiring the code to move maintenance of module parameters, buffers, and submodules into managed code.
> 
> Basically, when setting a parameter property such as Linear.weight, the named_parameters collection was not updated. This meant that an optimizer would not pick up the new tensor, leading to errors.

Are you absolutely sure this is necessary?   It is not at all ideal - and must surely represent a bug in LibTorch itself, mustn't it?  Is it really the case that C++ Modules have this problem?

> Further, BatchNorm{123]d did not include running_mean or running_var in the state_dict() results, nor did it allow those buffers to be accessed or set.

Hmmm.... Those two buffers are not parameters, my loose understanding is that they meant to be part of `state_dict`.  In DiffSharp, for example, they are buffers but not parameters https://github.com/DiffSharp/DiffSharp/blob/dev/src/DiffSharp.Core/Model.BatchNorm.fs#L38

src/TorchSharp/NN/Module.cs
NiklasGustafsson: The logic was introduced when I added properties as valid holders of parameters and buffers. In order to avoid double-counting, I have to filter out the backing field for such properties. Now that you bring it up, I realize that it may not work the same in F#. The F# examples run to completion, but it may be double-counting (wasting time).

NiklasGustafsson:
> Hmmm.... Those two buffers are not parameters, my loose understanding is that they meant to be part of `state_dict`. In DiffSharp, for example, they are buffers but not parameters

That is correct. Before the PR, they did not appear in state_dict(), and now they do. As buffers.

src/TorchSharp/NN/Module.cs
NiklasGustafsson: I guess we could avoid this mess by only allowing buffers and parameters to be held in properties, not in fields.

NiklasGustafsson:
> Are you absolutely sure this is necessary? It is not at all ideal - and must surely represent a bug in LibTorch itself, mustn't it? Is it really the case that C++ Modules have this problem?

I don't know what to make of it, but it certainly is what's happening. Setting the 'weight' field in native code does not affect the parameters or named_parameters collections. I also note that in PyTorch, the parameters and buffers are maintained in Python code.

src/TorchSharp/NN/Module.cs
NiklasGustafsson: @dsyme -- I pulled the support for properties holding parameters and buffers out of this PR. We can resurface it later, with proper cross-language support.

dsyme:
> I don't know what to make of it, but it certainly is what's happening. Setting the 'weight' field in native code does not affect the parameters or named_parameters collections. I also note that in PyTorch, the parameters and buffers are maintained in Python code.

Hmmm I see.  It's really unfortunate as I'd always hoped TorchSharp could avoid this sort of code.  I ripped out a bunch of it in the revamp, since the information seemed to be available in the C++.  We should probably dig into this a little more.

NiklasGustafsson:
@dsyme -- we can always rip it out again later, if we discover the secret to getting parameters() to be updated. The implementation does not impact the API surface area. The breaking changes to APIs were made independently of the internals.

PR #506, opened 2022-02-04
Move MNISTReader to use new api

dayo05:


NiklasGustafsson:
@dayo05 -- I'll go through this PR and try it out today. Looking at it, it seems like a good idea to accept it.

I would also like to draw your attention to our examples repo: https://github.com/dotnet/TorchSharpExamples, and suggest that you take a look at the C# and F# examples there. If you want to add this API to those data loaders, that would be great, and perhaps also add a tutorial on how to use the data reader APIs. If you're not comfortable with F#, I can translate from C# later.

src/TorchSharp/TorchVision/dsets/MNISTReader.cs
```diff
@@ -0,0 +1,105 @@
...
+    public sealed class MNISTReader : Dataset
+    {
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="path">Path to the folder containing the image files.</param>
+        /// <param name="prefix">The file name prefix, either 'train' or 't10k' (the latter being the test data set).</param>
+        /// <param name="device">The device, i.e. CPU or GPU to place the output tensors on.</param>
+        /// <param name="transform"></param>
+        public MNISTReader(string path, string prefix, Device device = null, ITransform transform = null)
```
NiklasGustafsson, line 28: 'device' is not used in this constructor -- it needs to be used to indicate where to place the tensors that are created.

src/TorchSharp/TorchVision/dsets/MNISTReader.cs
```diff
@@ -0,0 +1,105 @@
...
+        private List<Tensor> data = new();
+        private List<Tensor> labels = new();
+
+        public override void Dispose()
+        {
+            data.ForEach(d => d.Dispose());
+            labels.ForEach(d => d.Dispose());
+        }
+
+        public override Dictionary<string, Tensor> GetTensor(long index)
```
NiklasGustafsson, line 95: It's always helpful to copy the doc strings from the base class or interface where a method is introduced to all locations where it is overridden / implemented

src/TorchSharp/TorchVision/dsets/BigEndianReader.cs
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
+using System;
+using System.IO;
+
+namespace TorchSharp.torchvision.dsets
+{
+    public class BigEndianReader
```
NiklasGustafsson, line 7: Please add comments.

NiklasGustafsson (changes requested):

src/TorchSharp/TorchVision/dsets/BigEndianReader.cs
NiklasGustafsson: @dayo05 -- ping! :-)

dayo05: I was focusing to other project, I;ve fixed that now

NiklasGustafsson:
@dayo05 -- this is a long weekend in the US, so I'll take a look at the PR changes on Tuesday.

src/Examples/AdversarialExampleGeneration.cs
```diff
@@ -76,20 +79,22 @@ internal static void Main(string[] args)
 
             MNIST.Model model = null;
 
-            var normImage = torchvision.transforms.Normalize(new double[] { 0.1307 }, new double[] { 0.3081 }, device: (Device)device);
+            var normImage = torchvision.transforms.Normalize(new double[] {0.1307}, new double[] {0.3081},
+                device: (Device) device);
 
-            using (var test = new MNISTReader(targetDir, "t10k", _testBatchSize, device: device, transform: normImage)) {
```
NiklasGustafsson, line 81: In the previous iteration, MNISTReader ignored the device, and here it isn't even being passed in. Somewhere, the code needs to exist to place the data tensors (and the model) on the GPU, if that is the desired device. When MNISTReader was only an example, it was obeying the device it was passed.

NiklasGustafsson (changes requested):

NiklasGustafsson:
@dayo05 -- would you be okay with me making some changes to your PR and then sending them your way?

NiklasGustafsson:
> @dayo05 -- would you be okay with me making some changes to your PR and then sending them your way?

I just sent you a PR in your fork. One thing I'm wondering about, is that the DataLoader always throws away the tensors that it has built up. That doesn't seem like it's necessary when 'shuffle=false'.

dayo05:
> > @dayo05 -- would you be okay with me making some changes to your PR and then sending them your way?
> 
> I just sent you a PR in your fork. One thing I'm wondering about, is that the DataLoader always throws away the tensors that it has built up. That doesn't seem like it's necessary when 'shuffle=false'.

But I think this makes same between shuffled datasets and non-shuffled datasets

NiklasGustafsson:
> 
> But I think this makes same between shuffled datasets and non-shuffled datasets

No, I don't think so. If shuffle=true, then the batches will be different in each epoch, so it makes sense to dispose each batch as soon as possible, to relieve memory pressure. On the other hand, when shuffling is turned off, then the batches are the same every epoch, and there's no need to reconstruct them. It is more efficient (but use more memory) to keep the batched data around for the next epoch.

NiklasGustafsson (changes requested):

dayo05:
@NiklasGustafsson bump

NiklasGustafsson (approved):

PR #517, opened 2022-02-16
Adding support for Tensor.conj() and related functions.

NiklasGustafsson:
This is a straight-forward edit: just adding a few Tensor methods related to conjugates.

### DIFF
--- a/src/TorchSharp/Tensor/Tensor.Math.cs
+++ b/src/TorchSharp/Tensor/Tensor.Math.cs
@@ -549,6 +549,81 @@ namespace TorchSharp
             }
 
             [DllImport("LibTorchSharp")]
+            static extern IntPtr THSTensor_conj(IntPtr tensor);
+
+            /// <summary>
+            /// Returns a view of input with a flipped conjugate bit. If input has a non-complex dtype, this function just returns input.
+            /// </summary>
+            /// <returns></returns>
+            public Tensor conj()
+            {
+                var res = THSTensor_conj(Handle);
+                if (res == IntPtr.Zero)
+                    torch.CheckForErrors();
+                return new Tensor(res);
+            }
+
+            [DllImport("LibTorchSharp")]
+            static extern IntPtr THSTensor_conj_physical(IntPtr tensor);
+
+            /// <summary>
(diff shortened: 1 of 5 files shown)


src/TorchSharp/Tensor/Tensor.Math.cs
```diff
@@ -548,6 +548,81 @@ public Tensor ceil_()
                 return new Tensor(res);
             }
 
+            [DllImport("LibTorchSharp")]
+            static extern IntPtr THSTensor_conj(IntPtr tensor);
+
+            /// <summary>
+            /// Returns a view of input with a flipped conjugate bit. If input has a non-complex dtype, this function just returns input.
+            /// </summary>
+            /// <returns></returns>
```
lostmsu, line 557: NIT: skip empty `<returns>`

src/TorchSharp/Tensor/Tensor.Math.cs
```diff
@@ -548,6 +548,81 @@ public Tensor ceil_()
...
+            public Tensor conj()
+            {
+                var res = THSTensor_conj(Handle);
+                if (res == IntPtr.Zero)
+                    torch.CheckForErrors();
+                return new Tensor(res);
```
lostmsu, line 563: Not for this PR, but would not it be nice to have a function, that combines result check with `Tensor` creation?

Then most of these `Tensor` functions would be one-liners. E.g. `=> Tensor.FromHandleOrError(THSTensor_conj(Handle))`

lostmsu (dismissed):

lostmsu:
@NiklasGustafsson I'd love to see the cause for the test failure. I see nothing wrong with the code. 😲

NiklasGustafsson:
> @NiklasGustafsson I'd love to see the cause for the test failure. I see nothing wrong with the code. 😲

So would I, and I will. :-)

NiklasGustafsson:
Huh!

The fix is to avoid 'bool' as the return type of the native function:

```C++
int64_t THSTensor_is_conj(const Tensor tensor)
{
    CATCH_RETURN_RES(int64_t, 0, res = tensor->is_conj();)
}
```

Issue #516, opened 2022-02-14
AdamW bug on v0.96.0


QingtaoLi1:
When I update TorchSharp from 0.95.4 to 0.96.0, some exception occurs on `torch.optim.AdamW()`:

> System.Runtime.InteropServices.SEHException: 'External component has thrown an exception.'

The only change I commit is lower-casing `train()` and `eval()`. Notice that the first argument of `torch.optim.AdamW()` (and other optimizers) changes from `IEnumerable<torch.Tensor>` to `IEnumerable<TorchSharp.Modules.Parameter>`, so I guess the bug is related to library changes to solve #501.

NiklasGustafsson:
@QingtaoLi1 -- the renaming and the type change should not affect anything. Down in native code, neither matters.

Could you provide the whole stack trace? The one line you have is not enough to give me an idea of what is going on.

NiklasGustafsson:
Also, in the next release, AdamW will be implemented in managed code, so things are likely to change.

QingtaoLi1:
> @QingtaoLi1 -- the renaming and the type change should not affect anything. Down in native code, neither matters.
> 
> Could you provide the whole stack trace? The one line you have is not enough to give me an idea of what is going on.

Stack trace here. It points to the constructor of Adam and nothing else. The only error message is "external component..." I showed above.
![image](https://user-images.githubusercontent.com/89620019/153972842-06ff4afb-f0e1-4abd-8e42-77c5583dbd18.png)
![image](https://user-images.githubusercontent.com/89620019/153972923-ab67342a-5695-4571-9dc6-3e55b5d3fd8a.png)

NiklasGustafsson:
Okay, thanks!

QingtaoLi1:
@NiklasGustafsson Oh I didn't notice I was captioning the error of `Adam`. But the exception is the same for both optimizers.

NiklasGustafsson:
Let's see how this changes with the new implementation in 0.96.1

QingtaoLi1:
> Let's see how this changes with the new implementation in 0.96.1

Great! I tried `SGD` just now and it also fails.

NiklasGustafsson:
Hmmm. There's something fundamentally wrong. Is there any possibility of any mixup of DLLs or packages?

QingtaoLi1:
I'm not sure. Just try deleting `.nuget\\packages\\torchsharp\\` and re-installing it, but nothing changes.

[...]

PR #523, opened 2022-02-18
Optimizers support for parameter groups

NiklasGustafsson:
This is a massive PR.

The main thrust is to support parameter groups in the optimizers and LR schedulers. This is a massive and breaking change.

The change also involves moving all the optimizer implementations to managed code, except for LBFGS, which does not allow more than a single parameter group, and therefore does not require the move.

On the way to implementing this support, a few things were discovered missing and/or incorrect, and were therefore fixed.

@lostmsu, @oluwandabira -- if you have a chance to take a look, too, that would be wonderful. I still haven't heard about the right way to include individuals in the 'Request Review' UI.


NiklasGustafsson:
@lostmsu:

Yes, excellent question. 

I completely agree that the fact that we're just interfacing with native code is attractive. Without a question, it is why we have come as far as we have as quickly as we have. There are a couple of areas, like determining whether types are float, complex, etc. doesn't seem worth going to C++ for, but otherwise, TorchSharp is and should remain, a .NET wrapper of LibTorch.

There are only (even before this PR) three areas where there is significant managed code:

1. TorchVision (but I moved code to C++ to avoid the dispose problem)
2. Optimizers (half of them were already in managed code)
3. LR Schedulers.

The situation with the optimizers is that only about half of the ones available in 1.10 are available in native code, so there was already a number of managed code implementations. In fact, I dragged my feet for a long time, some months back, because I didn't want to tackle the managed code optimizers -- this was before the DisposeScope addition, when expressing tensor computations in .NET was nasty.

So, faced with a disruptive change, i.e. adding parameter groups, and the challenge of representing / encoding parameter groups and their options in two different ways, and having to continue to maintain two very different approaches to implementation, I felt that it would be better to have just one, like Pytorch does.

So, that's the rationale, for better or worse.

src/TorchSharp/NN/Module.cs
```diff
@@ -151,7 +153,9 @@ public virtual Module to(DeviceType deviceType, int deviceIndex = -1)
 
                             if (param is not null) {  // This test must come before the Tensor test
                                 if (deviceType != param.device_type || deviceIndex != param.device_index) {
-                                    var p = new Modules.Parameter(param.to(deviceType, deviceIndex), param.requires_grad);
+                                    var t = param.to(deviceType, deviceIndex);
+                                    t.retain_grad();
```
lostmsu, line 157: Can you comment on the need for `retain_grad` call here? Is there a bug this fixes?

src/TorchSharp/NN/LRScheduler.cs
```diff
@@ -26,16 +26,18 @@ protected LRScheduler()
 
                     }
 
-                    protected LRScheduler(ILearningRateController optimizer, int last_epoch = -1, bool verbose = false)
+                    protected LRScheduler(Optimizer optimizer, int last_epoch = -1, bool verbose = false)
```
lostmsu, line 29: It would seem more natural to have `LRScheduler` work per parameter group, not per optimizer. Same for all other schedulers.

src/TorchSharp/NN/Optimizer.cs
```diff
@@ -293,563 +452,1901 @@ public static AdadeltaOptimizer Adadelta(IEnumerable<(string name, Modules.Param
...
+                        foreach (var group in _parameter_groups) {
 
-                    using var grad = p.grad();
+                            body(group as T);
 
-                    if (grad is null) continue;
+                        }
 
-                    grad.zero_().Dispose();
+                        d.DisposeEverything();
```
lostmsu: Why do you have to call this manually? I thought you did not need to, it would be called by the implicit `d.Dispose()`?

src/TorchSharp/NN/Optimizer.cs
```diff
@@ -293,563 +452,1901 @@ public static AdadeltaOptimizer Adadelta(IEnumerable<(string name, Modules.Param
...
+            {
+                base.Parameters = parameters;
+                base.Options = options;
+            }
+
+            public new TOptions Options { get => (TOptions)base.Options; set => base.Options = value; }
+
+        }
+
+        public class SGD : OptimizerHelper, IMomentum
```
lostmsu: I would suggest to move individual optimizers into their own files.

src/TorchSharp/NN/Optimizer.cs
```diff
@@ -293,563 +452,1901 @@ public static AdadeltaOptimizer Adadelta(IEnumerable<(string name, Modules.Param
...
+                        param.add_(grad, alpha: alpha);
+
+                    }
+                }, closure);
+            }
+
+            protected override void Dispose(bool disposing)
+            {
+                base.Dispose(disposing);
+                foreach (var (name, state) in _state) {
```
lostmsu: NIT: you can do `_state.Values` instead of unpacking `KeyValuePair` here.

Also, where is `_state` defined? When reading this change sequentially, I encountered it a few times by now, but without seeing its definition before the first use I am not certain about its contents.

src/TorchSharp/NN/Optimizer.cs
```diff
@@ -293,563 +452,1901 @@ public static AdadeltaOptimizer Adadelta(IEnumerable<(string name, Modules.Param
...
+                if (!opt.maximize.HasValue) opt.maximize = def.maximize;
+
+                opt.InitialLearningRate = opt.LearningRate.Value;
+
+                _parameter_groups.Add(param_group);
+
+                foreach (var p in param_group.Parameters) {
+                    var state = new State();
+                    _state[p.Handle] = state;
+                    state.momentum_buffer = null;
```
lostmsu: NIT: why is this needed?

src/TorchSharp/NN/Optimizer.cs
```diff
@@ -293,563 +452,1901 @@ public static AdadeltaOptimizer Adadelta(IEnumerable<(string name, Modules.Param
...
+                _parameter_groups = new List<Modules.ParamGroup>();
+
+                foreach (var g in parameters) {
+                    add_param_group(g);
+                }
+            }
+
+            public override Tensor step(Func<Tensor> closure = null)
+            {
+                return _step<ParamGroup>(group => {
```
lostmsu: I can't really check the correctness of the optimizer implementation.

A few things:

1. It needs a link to either the paper or to the PyTorch implementation
2. It needs a reproducible test for every valid options combination.

lostmsu (changes requested):

src/TorchSharp/NN/Module.cs
NiklasGustafsson: Yes, this is the fix to #516

src/TorchSharp/NN/LRScheduler.cs
NiklasGustafsson: Following the Pytorch API.

https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.StepLR.html#steplr

 The schedulers get the parameter groups from the optimizer.

lostmsu: Is there a reason for the scheduler to store a link to the optimizer? If it only needs to store parameter groups internally, this it what is should it. You can have a constructor overload taking optimizer.

src/TorchSharp/NN/Optimizer.cs
```diff
@@ -293,563 +452,1901 @@ public static AdadeltaOptimizer Adadelta(IEnumerable<(string name, Modules.Param
...
+                            : grad.alias();
+
+                        square_avg.mul_(rho).addcmul_(grad, grad, 1 - rho);
+
+                        var std = square_avg.add(eps).sqrt_();
+                        var delta = acc_delta.add(eps).sqrt_().div_(std).mul_(grad);
+
+                        param.add_(delta, alpha: -lr);
+                        acc_delta.mul_(rho).addcmul_(delta, delta, 1 - rho);
+                    }
```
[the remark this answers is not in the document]
NiklasGustafsson: Yes, that's my conviction. the algorithms of optimizers are very stable.

src/TorchSharp/NN/Optimizer.cs
lostmsu: Writing tests could be pretty easy with `TestOptimizer(Model, Optimizer, float targetLoss)` helper function that would do a reproducible run.

The problem would be to know what the target loss should be.

src/TorchSharp/NN/Optimizer.cs
```diff
@@ -293,563 +452,1901 @@ public static AdadeltaOptimizer Adadelta(IEnumerable<(string name, Modules.Param
...
+                }, closure);
+            }
+
+            protected override void Dispose(bool disposing)
+            {
+                base.Dispose(disposing);
+                foreach (var (name, state) in _state) {
+                    if (state.momentum_buffer is not null) {
+                        state.momentum_buffer.Dispose();
+                    }
```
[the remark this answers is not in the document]
NiklasGustafsson: I agree -- that's cleaner. Thx!

src/TorchSharp/NN/Optimizer.cs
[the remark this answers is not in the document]
lostmsu: @NiklasGustafsson frankly, this alone looks like at least 2x in size and probably 10-20x in complexity over mirroring structs in C++.

NiklasGustafsson:
@lostmsu -- your review and comments are much appreciated!

test/TorchSharpTest/TestTraining.cs
```diff
@@ -183,6 +185,43 @@ public void TestTrainingAdamDefaults()
...
+
+                finalLoss = lossVal;
+
+                optimizer.zero_grad();
+
+                output.backward();
+
+                optimizer.step();
+            }
+            Assert.True(finalLoss < initialLoss);
```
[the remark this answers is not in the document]
NiklasGustafsson: I have to think about that. In the past, we've struggled with reproducible RNG sequences in the unit tests, since the tests are run in parallel. As long as there aren't any hidden random number API calls that use the global RNG, it should be doable by using Generator objects.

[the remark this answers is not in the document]
NiklasGustafsson: For example, I don't know how Dropout layers generate the mask, but we can just avoid Dropout in unit tests.

src/TorchSharp/NN/LRScheduler.cs
lostmsu: Sounds like a pretty good explanation to me.

But if user can remove a parameter group, the corresponding state should be removed as well.

src/TorchSharp/NN/Optimizer.cs
NiklasGustafsson: >> The problem would be to know what the target loss should be.

Yes. I can think of doing it two ways:

1. Write all the unit tests in both Python and .NET and use the Python values as 'expected.' Tons of work.
2. Run the unit tests once, look at the loss and use that as the target loss.

GeorgeS2019:
> @NiklasGustafsson before jumping into this, what is the motivation to move optimizer implementations into C# code? Every implementation copied from/modeled after the original PyTorch code is a maintenance burden, and a huge potential for future divergence from PyTorch in subtle but potentially very important details. Also error prone. That approach and its consequences are one of the major reasons I am not using TensorFlow.NET project.

@lostmsu You made a right observation. One critical difference between TorchSharp and Tensorflow.NET is that the **% of Unit Test code coverage** of LibTorch API and TorchSharp (in my impression) is significantly much higher in TorchSharp compared the situation in Tensorflow.NET.  

The transparency of code coverage and regular alignment with the latest pytorch in Torchsharp is also a key advantage.

src/TorchSharp/NN/Optimizer.cs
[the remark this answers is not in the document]
NiklasGustafsson: Here's a new issue to track: 

https://github.com/dotnet/TorchSharp/issues/531

NiklasGustafsson:
The PR comments, except for the request to revert optimizers to native code, have been taken into account. The optimizer implementation request will be tracked in its own issue (#531) and addressed as long as the long-term cost is not found to be prohibitive.

PR #527, opened 2022-02-19
Correct brackets

oluwandabira:


oluwandabira:
I don't think that error is from my end.

NiklasGustafsson:
@oluwandabira -- the bracket was actually meant to be turned that way -- [0,1[ is an alternative notation for a half-open interval (see [https://en.wikipedia.org/wiki/Interval_(mathematics)], and the one that was used where I went to school. If it's confusing to most people, changing it to [0,1) would be the correct change.

oluwandabira:
Oh, I didn't realize it was an interval. I'm used to [0, 1) but I don't know which is more popular so I'll just leave it alone.

PR #530, opened 2022-02-22
fix package loading of TorchSharp-cpu from F#/.NET Interactive

dsyme:
This should fix https://github.com/DiffSharp/DiffSharp/issues/387


dsyme:
This stuff is annoying and hard to test - it requires running an F# script against actual nuget package resolutions.  We could possibly do it after our package build by adding a local source, the running a small F# Interactive script using `dotnet fsi`.

PR #533, opened 2022-02-23
Bug525

### LINKED ISSUE #525: LSTM's forward function not work with null hidden and cell state
Hi, I'm implementing an example of using LSTM, however, I found that if I pass null `h0_c0` to 'LSTM.forward' there will raise an exception like:

> Unhandled exception. System.Runtime.InteropServices.ExternalException (0x80004005): Expected hidden[0] size 1 128 512, got [0]

NiklasGustafsson:
Fixes #525 for LSTM, GRU, and RNN modules.


test/TorchSharpTest/NN.cs
```diff
@@ -3008,6 +3008,19 @@ public void TestRNN2()
 
         }
 
+        [Fact]
+        public void TestRNN3()
```
lostmsu, line 3012: NIT: TestRnnDefaultH0?

lostmsu (dismissed):

PR #534, opened 2022-02-23
Fixed double-counting of parameters when a module is referenced in more than one way.

NiklasGustafsson:


src/TorchSharp/NN/Module.cs
```diff
@@ -461,13 +461,19 @@ public virtual IEnumerable<(string name, Module module)> named_modules()
...
+                        seen.Add(nsm.Item2.Handle);
                         yield return nsm;
                     }
 
                     if (recurse) {
                         foreach (var (name, sm) in _internal_submodules) {
                             foreach (var (n, p) in sm.named_parameters(true)) {
+                                if (seen.Contains(p.Handle)) continue;
+                                seen.Add(p.Handle);
                                 yield return ($"{name}.{n}", p);
```
lostmsu, line 477: seen should be passed recursively

NiklasGustafsson: I guess that would be more efficient, but it shouldn't affect the result. I'll have to create a non-public member for it.

lostmsu: You are right, I did not notice it was not actually a counting function, but a parameter enumerator.

lostmsu (approved):

src/TorchSharp/NN/Module.cs
NiklasGustafsson: It's still a good idea to pass it down for efficiency.

NiklasGustafsson:
@Metritutus -- thanks for the issue, it allowed me to find and fix a significant error in the implementation of Module.parameters()!

Issue #519, opened 2022-02-17
Multiprocessing dataloader support


peteryang1:
Hello,

In Pytorch, a 'torch.utils.data.DataLoader' is often used to serve the data preprocessing in multiprocessing mode because data processing is very time costy and Python cannot start multithread in main thread. 
However, TorchSharp is based on C# which supports multi thread while Python should use multi process to make the best use of the CPU.

Is there a plan of implementing a multi thread DataLoader in TorchSharp or the users are intended to write their own multi-thread dataloader to support a faster and parallel data preprocess?

Thanks!

NiklasGustafsson:
@peteryang1:

Yes and no. The DataLoader class that @dayo05 recently implemented and introduced, implements IEnumerable, and is therefore sequential in its API. It defers to 'Dataset' to implement the retrieval of the data. Dataset is an abstract class, and the concrete implementations could very well be implemented using concurrency.

Adding @oluwandabira, who has been thinking about this stuff related to TorchVision.

peteryang1:
Thanks @NiklasGustafsson for the reply.

As we can find from Pytorch API Doc, the 'dataloader' handles the multi process procedures while the 'dataset' class implemented by the user only considers one sample strategy.
![image](https://user-images.githubusercontent.com/25981102/154891355-83fa6812-d96f-4f07-93f5-8dc9888ebacc.png)

It seems the MNIST sample codes provided only supports sequential data process strategy which makes training slow down when data augmentation is very time costy.

NiklasGustafsson:
Adding @dayo05 and @oluwandabira to the discussion. This is a very important feature, so it would be great to get input from all who have been thinking about data load.

oluwandabira:
Looking through the current `TorchSharp.Data`, it appears that everything there is a wrapper around `LibTorch` methods, which makes sense as TorchSharp is to be just a wrapper but if any async/parallel Loaders are to be implemented then we would have to add pure C# implementations of those classes (and maybe alter Dataset to have aync methods?).

That is my current understanding of the situation and I wouldn't mind the pure C# code if it allows faster loading times.

NiklasGustafsson:
@peteryang1:

Maybe I'm misunderstanding what the multi-process support is about, but looking at the PyTorch documentation for Dataset, it seems like it's about splitting the data set among multiple processes. I thought you were asking about concurrency in loading the data.

PR #539, opened 2022-02-28
Support multiprocessing dataloader

dayo05:


src/TorchSharp/DataLoader.cs
```diff
@@ -112,16 +123,51 @@ public bool MoveNext()
...
+                                while (TaskBatch()) {
+                                    var cv = 0L;
+                                    lock (moveNextLock) {
+                                        if (!MoveNextValue()) {
+                                            break;
+                                        }
+                                        cv = currentVal;
+                                    }
+
+                                    dic.Add(dataset.GetTensor(cv));
```
lostmsu, line 157: `List<T>` is not thread safe

src/TorchSharp/DataLoader.cs
```diff
@@ -112,16 +123,51 @@ public bool MoveNext()
...
                                 Current[x] = cat(dic.Select(k => k[x].unsqueeze(0)).ToArray(), 0).to(device);
                             return true;
+
+                            void CreateBatch(object _)
+                            {
+                                while (TaskBatch()) {
+                                    var cv = 0L;
+                                    lock (moveNextLock) {
+                                        if (!MoveNextValue()) {
+                                            break;
```
lostmsu, line 152: Without building the mental model it is unclear to me if this would work when number of workers does not evenly divide the number of elements in the dataset.

src/TorchSharp/DataLoader.cs
```diff
@@ -112,16 +123,51 @@ public bool MoveNext()
...
+                            var t = currentVal;
+
+                            //Run Async
+                            foreach(var _ in Enumerable.Range(1, num_worker - 1))
+                                ThreadPool.QueueUserWorkItem(CreateBatch);
+
+                            dic.Add(dataset.GetTensor(t));
+                            CreateBatch(null);
+
+                            while (dic.Count < taskedBatchCount) { } //Wait for every task finished
```
lostmsu, line 139: If I see this correctly, this loop's behavior is barely predictable.

To see the point, imagine by the time this thread reaches this line, none of the `QueueUserWorkItem` has started their execution.

I believe in this case the while loop's condition will be `false`, so it won't wait for anything asynchronous.

lostmsu (changes requested):

NiklasGustafsson:
@dayo05 -- adding this multi-threading capability is a big deal, especially since concurrency bugs are usually hard to track down, so I will have some questions.

When I look at the code, I see it create a number of workers, which call GetTensor() in parallel. This assumes that the native CPU backend is thread-safe, which I don't think we know for sure -- maybe it is, maybe it isn't. Taking a lock around that code won't really help, since that just serializes the work, and the parallelism is lost.

Have you measured a speedup on any data set, real or faked?

dayo05:
@NiklasGustafsson 
I made a simple test for this
```cs
class ts : torch.utils.data.Dataset
{
    private ITransform tf = transforms.Compose(transforms.RandomRotation((1, 2)),
        transforms.RandomRotation((1, 2)),
        transforms.RandomRotation((1, 2)),
        transforms.RandomRotation((1, 2)));
    public override long Count { get; } = 10000;
    public override Dictionary<string, torch.Tensor> GetTensor(long index)
    {
        var t = torch.randn(1, 100, 100, torch.ScalarType.Float32);
        t = tf.forward(t);
        return new() {
            {"asdf", t}
        };
    }
}
[Fact]
public void Test1()
{
    var d = new ts();
    foreach (var x in new torch.utils.data.DataLoader(d, 32)) { }
}
[Fact]
public void Test2()
{
    var d = new ts();
    foreach(var x in new torch.utils.data.DataLoader(d, 32, num_worker: 2)) {}
}
```
In this case, Test1 spends time about double of Test2
> Test1: 16sec 768ms
> Test2: 8sec 912ms

Also, this will be advanced feature for dataloader, the default value is also 1 which doesn't uses multiprocessing.

dayo05:
I did more tests with increasing about num_workers and in that case, over the 5 worker spends more times. But less then 4, This feature improve speed.

NiklasGustafsson:
@dayo05 -- thanks for that.

I also wanted to ask again, as I did in the issue (https://github.com/dotnet/TorchSharp/issues/519): Is this the kind of multi-processing that the PyTorch DataLoader implements? From the documentation, it seems like that is used to divide the dataset among multiple processes, each of which trains on a subset of the data. In this case, it's a speedup of the construction of tensors, which seems like an entirely different thing, an implementation detail, in fact.

NiklasGustafsson:
I wrote a stress test to see what happens if we get a few more things happening at once. It did not persuade me that we can trust the native backend to be thread-safe. Given the Python context, it would make any sense to me for LibTorch to be safe, either, that's just a waste of efficiency.

Here's the test:

```C#
private class LargeTestDataset : torch.utils.data.Dataset
        {
            public override long Count { get; } = 128;
            public override Dictionary<string, torch.Tensor> GetTensor(long index)
            {
                return new() { { "data", torch.rand(3,1920,1080) }, { "label", torch.tensor(16) }, { "index", torch.tensor(index) } };
            }
        }

        [Fact]
        public void MultiThreadDataLoaderTest2()
        {
            using var dataset = new LargeTestDataset();
            using var dataloader = new torch.utils.data.DataLoader(dataset, 64, false, torch.CPU, num_worker: 2);
            var iter = dataloader.GetEnumerator();
            iter.MoveNext();
            var x = iter.Current;
            Assert.Equal(new long[] { 64, 3, 1920, 1080 }, x["data"].shape);
            iter.MoveNext();
            x = iter.Current;
            Assert.Equal(new long[] { 64, 3, 1920, 1080 }, x["data"].shape);
            Assert.False(iter.MoveNext());
            iter.Dispose();
        }

        [Fact]
        public void MultiThreadDataLoaderTest3()
        {
            using var dataset = new LargeTestDataset();
            using var dataloader = new torch.utils.data.DataLoader(dataset, 64, false, torch.CPU, num_worker: 10);
            var iter = dataloader.GetEnumerator();
            iter.MoveNext();
            var x = iter.Current;
            Assert.Equal(new long[] { 64, 3, 1920, 1080}, x["data"].shape);
            iter.MoveNext();
            x = iter.Current;
            Assert.Equal(new long[] { 64, 3, 1920, 1080 }, x["data"].shape);
            Assert.False(iter.MoveNext());
            iter.Dispose();
        }

        private void RunMany(object _)
        {
            MultiThreadDataLoaderTest3();
        }

        [Fact]
        public void MultiThreadDataLoaderStress()
        {
            foreach (var _ in System.Linq.Enumerable.Range(1, 25))
                System.Threading.ThreadPool.QueueUserWorkItem(RunMany);

            RunMany(null);
        }
```

When I run this, the process crashes in lots of places, and slowed down my Windows system to a crawl while I was dismissing many many crash dialogs.

NiklasGustafsson:
I did some stress-testing, taking your newly introduced tests and using larger tensors, then run many of them in parallel. This crashed the test run, as well as some apps that were running.

I found some memory inefficiencies that I corrected in a separate PR, which affect the regular (non-parallel) version, too.

In order for us to get comfortable with running DataLoader in parallel, we would have to do a lot of systematic stress testing.

dayo05:
> @dayo05 -- thanks for that.
> 
> I also wanted to ask again, as I did in the issue (#519): Is this the kind of multi-processing that the PyTorch DataLoader implements? From the documentation, it seems like that is used to divide the dataset among multiple processes, each of which trains on a subset of the data. In this case, it's a speedup of the construction of tensors, which seems like an entirely different thing, an implementation detail, in fact.

I know as python has GIL so CPU-based multi-thread tasks will spend more times then which single-thread does. In case of .NET, it supports multi-thread so This will be better way to do.

src/TorchSharp/DataLoader.cs
```diff
@@ -111,17 +121,50 @@ public bool MoveNext()
...
+
+                            var tensorIndexList = new List<long> {currentVal};
+                            for (int i = 1; i < batchSize; i++) {
                                 if (!MoveNextValue()) break;
-                                dic.Add(dataset.GetTensor(currentVal));
+                                tensorIndexList.Add(currentVal);
                             }
 
+                            if (!shuffle)
+                                tensorIndexList.Reverse(); // ConcurrentBag::Add inserts data into first of the collections. So, I'll reverse them.
```
lostmsu, line 132: Why not use `ConcurrentQueue` then?

Why does order the matter here?

src/TorchSharp/DataLoader.cs
```diff
@@ -111,17 +121,50 @@ public bool MoveNext()
...
+                            var taskedBatchCount = 0;
+                            var taskBatchLock = new object();
+
+                            //Run Async
+                            foreach(var _ in Enumerable.Range(1, num_worker - 1))
+                                ThreadPool.QueueUserWorkItem(CreateBatch);
+
+                            CreateBatch(null);
+
+                            while (dic.Count < tensorIndexList.Count) { } //Wait for tensors ready
```
lostmsu, line 144: While this is certainly a way to wait, you should be using a [`Semaphore`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.semaphore?view=net-6.0) instead.

You would not need a concurrent bag here then, an array would do.

lostmsu, line 144: Please, implement proper exception handling for failures in `GetTensor`. As it stands right now, the exception will not be propagated to `MoveNext`.

src/TorchSharp/DataLoader.cs
```diff
@@ -111,17 +121,50 @@ public bool MoveNext()
...
+
                             Current = new();
-                            foreach (var x in dic[0].Keys)
+                            foreach (var x in dic.First().Keys)
                                 Current[x] = cat(dic.Select(k => k[x].unsqueeze(0)).ToArray(), 0).to(device);
```
lostmsu, line 148: The order of tensors in the `cat` operation will not be deterministic. Though this is an existing issue, it gets more complicated with concurrency.

src/TorchSharp/DataLoader.cs
```diff
@@ -111,17 +121,50 @@ public bool MoveNext()
...
+                                }
+                            }
+
+                            long? TaskBatch()
+                            {
+                                lock (taskBatchLock) {
+                                    if (taskedBatchCount < tensorIndexList.Count)
+                                        return tensorIndexList[taskedBatchCount++];
+                                    return null;
+                                }
```
lostmsu, line 166: You don't need a lock here. `Interlocked.Increment` should do.

src/TorchSharp/DataLoader.cs
```diff
@@ -111,17 +121,50 @@ public bool MoveNext()
...
+
+                            while (dic.Count < tensorIndexList.Count) { } //Wait for tensors ready
+
                             Current = new();
-                            foreach (var x in dic[0].Keys)
+                            foreach (var x in dic.First().Keys)
                                 Current[x] = cat(dic.Select(k => k[x].unsqueeze(0)).ToArray(), 0).to(device);
                             return true;
+
+                            void CreateBatch(object _)
```
lostmsu, line 151: NIT: `ProcessPendingBatches`

src/TorchSharp/DataLoader.cs
```diff
@@ -111,17 +121,50 @@ public bool MoveNext()
...
+                            {
+                                while (true) {
+                                    var idx = TaskBatch();
+                                    if (idx is null) break;
+                                    dic.Add(dataset.GetTensor(idx.Value));
+                                }
+                            }
+
+                            long? TaskBatch()
```
lostmsu, line 160: NIT: `ScheduleBatch`.

lostmsu (changes requested):

src/TorchSharp/DataLoader.cs
lostmsu: This is related to the `Semaphore` comment. You might want to consider using .NET async tasks instead of `QueueWorkItem` to properly wire exceptions.

lostmsu: Please also add a test for exception handling.

src/TorchSharp/DataLoader.cs
```diff
@@ -111,17 +121,50 @@ public bool MoveNext()
...
+                            if (!shuffle)
+                                tensorIndexList.Reverse(); // ConcurrentBag::Add inserts data into first of the collections. So, I'll reverse them.
+
+                            ConcurrentBag<Dictionary<string, Tensor>> dic = new();
+                            var taskedBatchCount = 0;
+                            var taskBatchLock = new object();
+
+                            //Run Async
+                            foreach(var _ in Enumerable.Range(1, num_worker - 1))
+                                ThreadPool.QueueUserWorkItem(CreateBatch);
```
NiklasGustafsson, line 140: I would suggest using the more modern Task API, instead of the thread pool. That also allows you to wait more efficiently (not spin in a loop).

See: 

https://dotnetcodr.com/2014/01/01/5-ways-to-start-a-task-in-net-c

and:

https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task?view=net-6.0

That allows you to collect all the task instances and wait for them with one call. You can still use the CreateBatch() that you have.

src/TorchSharp/DataLoader.cs
dayo05: I've test about error handling with code
```cs
class ErrorDataset : torch.utils.data.Dataset
        {
            public override long Count { get; } = 100;
            public override Dictionary<string, torch.Tensor> GetTensor(long index)
            {
                throw new System.NotImplementedException();
            }
        }
        [Fact]
        public void T()
        {
            var d = new torch.utils.data.DataLoader(new ErrorDataset(), 2);
            foreach(var x in d) {}
        }
```
in NUnit test.

This shows error like 
```
System.NotImplementedException: The method or operation is not implemented.

System.NotImplementedException
The method or operation is not implemented.
   at TorchSharp.TestDataLoader.ErrorDataset.GetTensor(Int64 index) in /home/dayo/RiderProjects/TorchSharp/test/TorchSharpTest/TestDataLoader.cs:line 138
   at TorchSharp.torch.utils.data.DataLoader.DataLoaderEnumerator.<>c__DisplayClass10_0.<MoveNext>g__ProcessPendingBatches|1() in /home/dayo/RiderProjects/TorchSharp/src/TorchSharp/DataLoader.cs:line 157
   at TorchSharp.torch.utils.data.DataLoader.DataLoaderEnumerator.MoveNext() in /home/dayo/RiderProjects/TorchSharp/src/TorchSharp/DataLoader.cs:line 142
   at TorchSharp.TestDataLoader.T() in /home/dayo/RiderProjects/TorchSharp/test/TorchSharpTest/TestDataLoader.cs:line 145
```
I think this will be enough. Stacktraces are showing fine.

src/TorchSharp/DataLoader.cs
dayo05: In changed code, I think lock is required because taskedBatchCount used with `(taskedBatchCount, tensorIndexList[taskedBatchCount++])`. Interlocked.Increment will not right way.

NiklasGustafsson:
@dayo05 -- I'll take another look at this PR today. Thanks for your perseverance!

src/TorchSharp/DataLoader.cs
NiklasGustafsson: The following code should be perfectly safe and avoid having the lock:

```C#
var tbc = System.Threading.Interlocked.Increment(ref taskedBatchCount);
if (taskedBatchCount <= tensorIndexList.Count)
    return (taskedBatchCount-1, tensorIndexList[tbc]);
return null;
```

NiklasGustafsson (approved):

PR #541, opened 2022-03-02
Explicit memory management in DataLoader.MoveNext()

NiklasGustafsson:
This will clean up temporaries before returning data from the DataLoader.

### DIFF
--- a/src/TorchSharp/DataLoader.cs
+++ b/src/TorchSharp/DataLoader.cs
@@ -110,18 +110,25 @@ namespace TorchSharp
                         public bool MoveNext()
                         {
                             DisposeCurrent();
-                            if (!MoveNextValue()) return false;
-                            List<Dictionary<string, Tensor>> dic = new();
-                            dic.Add(dataset.GetTensor(currentVal));
-                            for (var i = 1; i < batchSize; i++) {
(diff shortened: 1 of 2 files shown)


NiklasGustafsson:
@lstmsu, @dayo05 -- please review if you have the time.

src/TorchSharp/DataLoader.cs
```diff
@@ -110,18 +110,25 @@ private bool MoveNextValue()
                         public bool MoveNext()
                         {
                             DisposeCurrent();
-                            if (!MoveNextValue()) return false;
-                            List<Dictionary<string, Tensor>> dic = new();
-                            dic.Add(dataset.GetTensor(currentVal));
-                            for (var i = 1; i < batchSize; i++) {
-                                if (!MoveNextValue()) break;
+                            using (var scope = DisposeScopeManager.NewDisposeScope()) {
```
lostmsu, line 113: NIT: using var scope = ...;

NiklasGustafsson: Yeah, I know...

I do usually end up relying on the `using ( ... ) { ... }` syntax in most circumstances. Maybe it's because it was what I got used to when 'using' first was introduced.

PR #543, opened 2022-03-05
Support TorchSharp on .NET Framework

tarekgh:


tarekgh:
CC @NiklasGustafsson @dsyme @michaelgsharp @ericstj

src/TorchSharp/NN/Module.cs
```diff
@@ -722,14 +722,14 @@ public static void SaveStateDictionary(System.IO.BinaryWriter writer, Dictionary
                 {
                     writer.Encode(sd.Count); // 4 bytes
 
-                    foreach (var (k, v) in sd) {
-                        writer.Write(k);
-                        v.Save(writer);
+                    foreach (var kvp in sd) {
+                        writer.Write(kvp.Key);
+                        kvp.Value.Save(writer);
```
lostmsu, line 727: I don't think you need to downgrade C#. You can explicitly set `<LangVersion>10.0</LangVersion>` and continue using new features in `netstandard2.0` target

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -1515,7 +1515,7 @@ public Tensor take_along_dim(Tensor indices, long dimension)
             /// Returns a tensor with the same data and number of elements as self but with the specified shape.
             /// </summary>
             /// <param name="shape">The new tensor shape.</param>
-            
+
```
lostmsu, line 1518: NIT: I'd completely remove those empty lines.

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -5533,13 +5533,16 @@ static public TensorIndex Slice(long? start = null, long? stop = null, long? ste
                 return new TensorIndex() { startIndexOrBoolOrSingle = start, step = step, stopIndex = stop, kind = Kind.Slice };
             }
 
+            static public TensorIndex Slice((int? start, int? end) range) => TensorIndex.Slice((long?)range.start, (long?)range.end);
+
+#if !NETSTANDARD2_0
```
lostmsu, line 5538: Might make sense to add [IndexRange](https://www.nuget.org/packages/IndexRange/) dependency to `netstandard2.0` target.

lostmsu (commented):
This looks simple enough. Though I'd avoid conditional compilation with `#if` by using compatiblity packages such as `System.Memory` (that you already use) and `IndexRange` which you should add, and setting `LangVersion` explicitly.

src/TorchSharp/NN/Module.cs
tarekgh: I am already using the latest version of the compiler. Look at the csproj to see that. I couldn't use the new syntax because the ns2.0 doesn't have the deconstructor api https://source.dot.net/#System.Private.CoreLib/KeyValuePair.cs,50.

src/TorchSharp/Tensor/Tensor.cs
tarekgh: I don't think it is worth it to add a dependency on a third-party library here. doing that comes with other issues like not have enough control to address any security issues on time and also not-guarantee the compatibility moving forward.

lostmsu: @tarekgh with 0.96 we don't need forward compatibility. I'd rather have no extra ifdefs for already niche 4.8 target.

NiklasGustafsson: I don't think we're quite ready to designate 4.8 a 'niche' target yet.

NiklasGustafsson:
General comment -- having the new syntax and capabilities is most important in the public API, and less so in the implementation of TorchSharp itself. 

If we can get rid of some of the #ifdefs by using common capabilities, let's do so unless it in a) and example or tutorial, or b) has some meaningful performance benefit.

tarekgh:
@NiklasGustafsson  

> General comment -- having the new syntax and capabilities is most important in the public API, and less so in the implementation of TorchSharp itself.
> If we can get rid of some of the #ifdefs by using common capabilities, let's do so unless it in a) and example or tutorial, or b) has some meaningful performance benefit.

I have addressed that. Let me know if there is any other specific place we need to change.

src/Examples/Examples.csproj
```diff
@@ -4,26 +4,34 @@
...
+    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT' And '$(IgnoreNetFX)' != 'true'">net472;$(TargetFrameworks)</TargetFrameworks>
+    <TargetFrameworks Condition="'$(TargetFrameworks)' == ''">net5.0</TargetFrameworks>
     <TestUsesLibTorch>true</TestUsesLibTorch>
     <UseMLCodeAnalyzer>false</UseMLCodeAnalyzer>
     <UseStyleCopAnalyzer>false</UseStyleCopAnalyzer>
     <StartupObject>TorchSharp.Examples.Program</StartupObject>
     <IsPackable>false</IsPackable>
     <PlatformTarget>x64</PlatformTarget>
     <RootNamespace>TorchSharp.Examples</RootNamespace>
+    <DefineConstants Condition="$([MSBuild]::GetTargetFrameworkIdentifier('$(TargetFramework)')) == '.NETFramework'">$(DefineConstants);NETFX472</DefineConstants>
```
joperezr, line 18: You don't need to define an additional constant. The SDK already defines constants when targeting netfx, and I believe the one you should use is `NET472_OR_GREATER` which will be true even if this project is retargeted to net48 for example. Same comment applies to all of the projects that you are manually defining this.

src/Examples/SequenceToSequence.cs
```diff
@@ -26,7 +26,11 @@ namespace TorchSharp.Examples
     public class SequenceToSequence
     {
         // This path assumes that you're running this on Windows.
+#if NETFX472
```
joperezr, line 29:
```suggestion
#if NET472_OR_GREATER
```
Same for all of the rest of places where you are using this constant.

src/FSharp.Examples/FSharp.Examples.fsproj
```diff
@@ -21,7 +24,8 @@
 
   <ItemGroup>
     <PackageReference Include="SharpZipLib" Version="1.3.3" />
-    <PackageReference Include="System.Memory" Version="4.5.3" />
+    <PackageReference Include="System.Memory" Version="4.5.4" />
+    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
```
joperezr, line 28: NIT: I suppose this package reference is only needed in net472, if so you can condition it.

src/TorchSharp/TorchSharp.csproj
```diff
@@ -3,19 +3,24 @@
...
   <PropertyGroup>
-      <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
+      <TargetFrameworks>netcoreapp3.1;netstandard2.0</TargetFrameworks>
       <IncludeInPackage>TorchSharp</IncludeInPackage>
       <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
       <UseMLCodeAnalyzer>false</UseMLCodeAnalyzer>
       <UseStyleCopAnalyzer>false</UseStyleCopAnalyzer>
       <IsPackable>false</IsPackable>
       <DefineConstants>$(DefineConstants);LIBTORCH_$(LibTorchPackageVersion.Replace('.', '_'));CUDA_$(CudaVersionDot.Replace('.', '_'))</DefineConstants>
+      <DefineConstants Condition="$([MSBuild]::GetTargetFrameworkIdentifier('$(TargetFramework)')) == 'netstandard2.0'">$(DefineConstants);NETSTANDARD2_0</DefineConstants>
```
joperezr, line 13: Same comment here, you don't need to define this, the SDK already does this for you. The one to use in this case is named: `NETSTANDARD2_0_OR_GREATER`

src/TorchSharp/TorchSharp.csproj
```diff
@@ -44,16 +49,21 @@
 
   <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
 
- <!-- Trigger the download+placement of the redist bits and the build of the C++ project -->
-  <Target Name="BeforeBuild">
+  <!-- Trigger the download+placement of the redist bits and the build of the C++ project -->
+  <Target Name="NativeLibtorch" BeforeTargets="BeforeBuild" >
```
joperezr, line 53: NIT
```suggestion
  <Target Name="BuildNativeLibTorch" BeforeTargets="BeforeBuild" >
```

test/Directory.Build.props
```diff
@@ -2,7 +2,7 @@
   <Import Project="..\Directory.Build.props" />
 
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFrameworks>net5.0;net48</TargetFrameworks>
```
joperezr, line 5: Don't you want to condition that net48 only gets added in Windows? Is this Directory.Build.props file used by any other project apart from the two projects being modified below? If the answer is no, then you can probably consider removing TargetFrameworks from this file and instead have it be defined in the projects directly

joperezr (approved):
Infrastructure-related changes look good to me, just left a few NITs.

Issue #546, opened 2022-03-10
TorchSharp is slower on NetFX comparing to Net Core


tarekgh:
We have enabled TorchSharp to run on .NET Framework. We noticed a performance degradation comparing when running against Net Core.  One obvious place to look at is the pinvoke calls cost. measuring simple pinvoke seeing the numbers:

```
|      Method      |     Mean      |    Error   |   StdDev   |  
| NetCore5OPInvoke |      4.371 ns |  0.0107 ns |  0.0095 ns |  
| NetFxPInvoke     |      7.885 ns |  0.0333 ns |  0.0295 ns | 
``` 

Need to look in general on the full stack to find out if there is any opportunity for optimization.

lostmsu:
FYI in Python.NET we ended up replacing PInvoke with C# unmanaged function pointers, and I have a fork of Roslynator that does conversion automatically: https://github.com/lostmsu/Roslynator/tree/features/ReplacePInvoke

tarekgh:
The interop is already internally using function pointers. It is worth trying to have all pinvokes signatures blittable and not use the built-in marshalling instead.

MovGP0:
This is somewhat expected, since there are a ton of performance optimizations in every new version of .NET. Especially .NET 6 and 7 have changed a lot in favor of performance.

alinpahontu2912:
Hey @tarekgh, Since .Net 5 is no longer supported, with a new upgrade to .Net 8 coming soon, is this still relevant, can it be closed?

tarekgh:
@alinpahontu2912 the issue here the performance degradation on the .NET Framework compared to the .NET Core. Upgrading to .NET 8.0 is irrelevant to this issue. If we don't target .NET Framework anymore then can be closed. Otherwise, I think need to keep it opened.

alinpahontu2912:
Hi @tarekgh,

Thank you for reporting the performance degradation issue with TorchSharp on .NET Framework compared to .NET Core. Your insights, especially regarding the PInvoke costs, have been very valuable. I noticed the linked PR #547 aimed to re-enable optimizations for test runs with .NET Framework, and I’d love to know if that has helped address your concerns.

Given the evolution of TorchSharp and the recent advancements in .NET, do you still find this issue relevant? If so, could you please provide updated information or any further observations? Your feedback is important to us as we continue to improve the library.

Thanks again for your contribution!

alinpahontu2912:
[...]

PR #547, opened 2022-03-10
Re-enable Optimization on test runs with NetFX 4.7.2

tarekgh:
This change is to re-enable the optimization option on the tests when running with NETFX 4.7.2. The change here is to serialize the test execution. This solves the issue of the crashing the test host and avoid the conflict between different tests.

https://github.com/dotnet/TorchSharp/issues/546
https://github.com/dotnet/TorchSharp/pull/543


tarekgh:
CC @NiklasGustafsson

NiklasGustafsson (approved):

PR #550, opened 2022-03-15
CIFAR10 and CIFAR100 Datasets

NiklasGustafsson:
Adding CIFAR 10 and 100 to the torchvision.datasets collection.


src/Examples/SequenceToSequence.cs
```diff
@@ -137,7 +137,7 @@ private static void train(int epoch, Tensor train_data, TransformerModel model,
                         total_loss += loss.to(torch.CPU).item<float>();
                     }
 
-                    if (batch % log_interval == 0 && batch > 0) {
+                    if (batch % log_interval == 0 && batch > 0 || (batch == tdlen / bptt)) {
```
tarekgh, line 140: nit: would be good if you add parentheses to these conditions to show the intention how the conditions  should be evaluated.

src/TorchSharp/TorchVision/dsets/CIFAR.cs
```diff
@@ -0,0 +1,316 @@
...
+                    Directory.CreateDirectory(targetDir);
+                }
+
+                DownloadFile(fileName, targetDir, baseUrl);
+
+                if (!File.Exists(Path.Combine(targetDir, $"{dataset}-binary.bin"))) {
+                    DecompressFile(fileName, targetDir, targetDir);
+                }
+            }
+            private static void DecompressFile(string file, string sourceDir, string targetDir)
```
tarekgh, line 100: nit: insert space line here.

src/TorchSharp/TorchVision/dsets/CIFAR.cs
```diff
@@ -0,0 +1,316 @@
...
+                Utils.Decompress.ExtractTGZ(Path.Combine(sourceDir, file), targetDir);
+            }
+
+            private void DownloadFile(string file, string target, string baseUrl)
+            {
+#if NETSTANDARD2_0_OR_GREATER
+                var filePath = NSPath.Join(target, file);
+#else
+                var filePath = Path.Join(target, file);
+#endif // NETSTANDARD2_0_OR_GREATER
```
tarekgh, line 111: what you think moving these into helper method to reduce the number of '#if' inserted in the sources?

src/TorchSharp/TorchVision/dsets/CIFAR.cs
```diff
@@ -0,0 +1,316 @@
...
+                fine_labels.ForEach(d => d.Dispose());
+                coarse_labels.ForEach(d => d.Dispose());
+            }
+
+            /// <summary>
+            /// Get tensor according to index
+            /// </summary>
+            /// <param name="index">Index for tensor</param>
+            /// <returns>Tensors of index. DataLoader will catenate these tensors into batches.</returns>
+            public override Dictionary<string, Tensor> GetTensor(long index)
```
tarekgh, line 302: >GetTensor [](http://example.com/codeflow?start=55&length=9)

I am wondering, do we claim this is a thread safe? or we don't care?

tarekgh (dismissed):
Added minor comments and question.

src/TorchSharp/TorchVision/dsets/CIFAR.cs
NiklasGustafsson: No, we don't claim that. There's another PR, which is mean to make DataLoader parallel, in which we have to deal with that.

PR #553, opened 2022-03-18
Added FractionalMaxPool{23}d modules.

NiklasGustafsson:


NiklasGustafsson:
@lostmsu, @oluwandabira -- since you both volunteered to do code reviews, could you please take a look at this? It's a straight-forward native interop implementation, so it should be quick.

oluwandabira (approved):
Code look alright and tests pass.

PR #571, opened 2022-03-28
Miscellaneous missing functions

NiklasGustafsson:
For starters, `Tensor.unfold`


ingted:
This is so AMAZING that I need to use Pythonnet to execute unfold and do a lot of converting process to PyObject to get back the unfolded result before and it seems like it is not required anymore!!!

HOW WONDERFUL!!!

PR #572, opened 2022-03-30
Add numpy-style ToString method

dayo05:
New method is exists on `Tensor::str()`. I'll create more tests for that.


src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -5369,6 +5369,79 @@ public override string ToString()
                 return sb.ToString();
             }
 
+            /// <summary>
```
NiklasGustafsson, line 5372: This belongs in TensorExtensionMethods, I think.

test/TorchSharpTest/TestTorchTensor.cs
```diff
@@ -239,6 +231,114 @@ public void Test6DToString()
...
+                        t[i][j] = torch.tensor((1.24f * i, 2.491f * i * 2), torch.complex64);
+                var str = t.str();
+                Assert.Equal("[[0 0 0 0 0]\n [1.24+4.982i 1.24+4.982i 1.24+4.982i 1.24+4.982i 1.24+4.982i]\n [2.48+9.964i 2.48+9.964i 2.48+9.964i 2.48+9.964i 2.48+9.964i]\n [3.72+14.946i 3.72+14.946i 3.72+14.946i 3.72+14.946i 3.72+14.946i]\n [4.96+19.928i 4.96+19.928i 4.96+19.928i 4.96+19.928i 4.96+19.928i]]", str);
+            }
+            Assert.Equal("[[0 0 0 0]\n [0 0 0 0]]", torch.zeros(2, 4, torch.complex64).str());
+            Assert.Equal("[[1 1 1 1]\n [1 1 1 1]]", torch.ones(2, 4, torch.complex64).str());
+            Assert.Equal("[[7.5 10 12.5 ... 125 127.5 130]\n [132.5 135 137.5 ... 250 252.5 255]]", torch.tensor(Enumerable.Range(1, 100).Select(x => x * 2.5 + 5).ToList(), new long[]{2, 50}, ScalarType.Float32).str());
+            Assert.Equal("[[7.5 10 12.5 ... 250 252.5 255]\n [257.5 260 262.5 ... 500 502.5 505]\n [507.5 510 512.5 ... 750 752.5 755]\n ...\n [24258 24260 24262 ... 24500 24502 24505]\n [24508 24510 24512 ... 24750 24752 24755]\n [24758 24760 24762 ... 25000 25002 25005]]",
```
NiklasGustafsson, line 273: The '\n' is problematic -- this test is not likely to work across platforms.  I put a declaration in TestTorchTensor for this reason:

```C#
private string _sep = Environment.NewLine;
```

and used it in the 'expected' strings in the tests.

src/TorchSharp/Tensor/Tensor.cs
NiklasGustafsson: Another possibility is to add a 'style' argument to the verbose ToString() that currently takes a boolean.

Instead of
```C#
public string ToString(bool withData, string fltFormat = "g5", int width = 100, CultureInfo? cultureInfo = null)
```

it would be something like this:

```C#
public enum TensorStringStyle { Metadata, Julia, Numpy }

public string ToString(TensorStringStyle style, string fltFormat = "g5", int width = 100, CultureInfo? cultureInfo = null) 
{
    if (style == TensorStringStyle.Metadata) return ToString();
    ...
}

// For backward compat:

public string ToString(bool withData, string fltFormat = "g5", int width = 100, CultureInfo? cultureInfo = null) 
{
    return ToString(withData ? TensorStringStyle.Julia : TensorStringStyle.Metadata, fltFormat, width, cultureInfo);
}
```

dayo05: Then how about this?
```cs
public static TensorStringStyle DefaultOutputStyle = TensorStringStyle.Metadata;

            public override string ToString() => ToString(DefaultOutputStyle);

            public string ToString(TensorStringStyle style, string fltFormat = "g5", int width = 100,
                CultureInfo? cultureInfo = null) => style switch {
                    TensorStringStyle.Metadata => ToMetadataString(),
                    TensorStringStyle.Julia => ToJuliaString(fltFormat, width, cultureInfo),
                    TensorStringStyle.Numpy => ToNumpyString(this, ndim),
                    _ => throw new InvalidEnumArgumentException("Not supported type")
                };
```

NiklasGustafsson: Sure. Don't you want CultureInfo and fltFormat on the NumPy string?

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -5369,20 +5382,84 @@ public override string ToString()
...
+
+                        for (var i = currentSize - 3; i < currentSize - 1; i++) {
+                            PrintValue(sb, t[i].ToScalar());
+                            sb.Append(' ');
+                        }
+
+                        PrintValue(sb, t[currentSize - 1].ToScalar());
+                    }
+                } else {
+                    var newline = string.Join("", Enumerable.Repeat('\n', (int) dim - 1).ToList());
```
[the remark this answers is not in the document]
NiklasGustafsson: That's why the NumPy tests are failing on Windows.

NiklasGustafsson:
New test failures. This time, I think it's because of the tests doing this:

```C#
.Select(x => x * 2.5 + 5)
```
which leads to rounding differences in various runtimes.

NiklasGustafsson (approved):

NiklasGustafsson:
@dayo05 -- I just merged your changes. Looking good!

With this, I will prepare and make a new release later today.

Issue #566, opened 2022-03-26
Implicit conversion of scalars to tensors


MovGP0:
There should be an implicit conversion of common number types to scalar tensors. Instead of:

```csharp
var points = torch.zeros(6, torch.ScalarType.Float16);
points[0] = torch.tensor(4.0f, torch.ScalarType.Float16);
points[1] = torch.tensor(1.0d, torch.ScalarType.Float32);
points[2] = torch.tensor(7, torch.ScalarType.Int32);
```

we should be able to write

```csharp
var points = torch.zeros(6, torch.ScalarType.Float16);
points[0] = 4.0f;
points[1] = 1.0d;
points[2] = 7;
```

NiklasGustafsson:
@MovGP0 -- that was the intent at one point, but it created a lot of ambiguity with existing APIs with overloads that took both a tensor and a scalar.

There is an explicit conversion, so you should be able to say:

```C#
var points = torch.zeros(6, torch.ScalarType.Float16);
points[0] = (Tensor)4.0f;
points[1] = (Tensor)1.0d;
points[2] = (Tensor)7;
```

which is slightly less cumbersome.

MovGP0:
There is the convention that converting without Information loss should be implicit and converting with Information reduction should be explicit.

The conversion from int/float/double/etc. to Tensor would increase Information and can be implicit, while the conversion from a scalar Tensor to a numeric value would need to be explicit.

Ambiguity shouldn't be a problem in theory. If there is ambiguity, it's usually  better to remove the interfaces that use a primitive type in favor of interfaces that use the Tensor, since there will be an automatic conversion anyway.

NiklasGustafsson:
Hmm, I'm going to have to think about that one. 

I've always thought of that convention in terms of "widening" or "narrowing" conversions, but to implicitly create a collection type from a scalar seems different from widening. I'm not immediately persuaded that it meets the criteria. 

That said, TorchSharp breaks many .NET conventions already, so I wouldn't personally let that stand in the way of usability. It's been a while since I looked at what started breaking when those operators were implicit, and I'll take another look.

NiklasGustafsson:
The specific challenge with indexing is that we can't overload the setters on the value type. Otherwise, we could fix this by taking 'Scalar':

```C#
[IndexerName("TensorItems")]
public Tensor this[long i1] {
    get { ... }
    set { ... }
}

[IndexerName("TensorItems")]
public Scalar this[long i1] {
    set { ... }
[...]

Issue #556, opened 2022-03-18
Implement missing activation modules


NiklasGustafsson:
Hardshrink
Hardsigmoid
Hardtanh
Hardswish
Tanhshrink
GLU
Threshold
Softsign
Softshrink
Softplus

PR #574, opened 2022-04-01
Issue #566

NiklasGustafsson:
Addressing issue #556

### DIFF
--- a/src/TorchSharp/Tensor/Tensor.cs
+++ b/src/TorchSharp/Tensor/Tensor.cs
@@ -1156,6 +1156,15 @@ namespace TorchSharp
             static extern void THSTensor_set1(IntPtr handle, long i1, IntPtr value);
 
             /// <summary>
+            /// This is only here in order to help the C# compiler make the right choice vis-a-vis
+            /// implicit conversions.
+            /// </summary>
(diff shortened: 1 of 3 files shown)


src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -5239,56 +5248,58 @@ public Tensor where(Tensor condition, Tensor other)
...
                 return torch.tensor(value);
             }
 
-            public static explicit operator Tensor(System.Numerics.Complex value)
+            public static implicit operator Tensor(System.Numerics.Complex value)
             {
                 return torch.tensor(value);
             }
 
+            //public static implicit operator Tensor(Scalar value) { throw new InvalidOperationException("Conversion from Scalar to Tensor"); }
```
MadsTorgersen, line 5301: Any reason to keep this commented-out version in here? You have an implementation further up.

MadsTorgersen (approved):

src/TorchSharp/Tensor/Tensor.cs
NiklasGustafsson: Nope. Thanks for catching that!

PR #578, opened 2022-04-19
Support load_state_dict()

NiklasGustafsson:
Support load_state_dict()
Modified version number.
Updated release notes.

### DIFF
--- a/src/TorchSharp/NN/Module.cs
+++ b/src/TorchSharp/NN/Module.cs
@@ -416,6 +416,46 @@ namespace TorchSharp
                     return destination;
                 }
 
+                /// <summary>
+                /// Copies parameters and buffers from state_dict into this module and its descendants.
+                ///
+                /// If strict is True, then the keys of state_dict must exactly match the keys returned by this module’s state_dict() function.
+                /// </summary>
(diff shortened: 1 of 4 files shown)


src/TorchSharp/NN/Module.cs
```diff
@@ -416,6 +416,46 @@ public virtual IEnumerable<(string name, Module module)> named_modules()
...
+                    }
+
+                    if (strict && (missing.Count > 0 || unexpected.Count > 0))
+                        throw new InvalidOperationException("The loaded state_dict is not identica to the target dictionary.");
+
+                    foreach (var key in source.Keys) {
+                        if (destination.ContainsKey(key)) {
+                            destination[key].bytes = source[key].bytes;
+                        }
+                    }
```
tarekgh, line 454: can this be combined with the previous loop in line #435 with checking strict value inside the loop?

tarekgh (approved):
Added a little comment. LGTM otherwise.

src/TorchSharp/NN/Module.cs
NiklasGustafsson: @tarekgh, that was what I wanted to do, but I didn't want to end up with a partially modified destination dictionary in case the exception is caught and dismissed.

PR #584, opened 2022-04-25
Added a couple of tensor APIs missing from 'torch'

NiklasGustafsson:


src/TorchSharp/Tensor/Tensor.torch.cs
```diff
@@ -214,12 +214,37 @@ public static Tensor dstack(IList<Tensor> tensors)
...
+        /// <summary>
+        /// Returns the k largest elements of the given input tensor along a given dimension.
+        /// </summary>
+        /// <param name="tensor">The input tensor</param>
+        /// <param name="k">The 'k' in 'top-k'.</param>
+        /// <param name="dimension">The dimension to sort along. If dim is not given, the last dimension of the input is chosen.</param>
+        /// <param name="largest">Controls whether to return largest or smallest elements</param>
+        /// <param name="sorted">Controls whether to return the elements in sorted order</param>
+        /// <returns></returns>
+        public static (Tensor values, Tensor indexes) topk(Tensor tensor, int k, int dimension = -1, bool largest = true, bool sorted = true) => tensor.topk(k, dimension, largest, sorted);
```
tarekgh, line 226: >Tensor [](http://example.com/codeflow?start=59&length=6)

Do we care in such APIs to check for null? or it is ok to have users getting NullReferenceException instead of ArgumentNullException?

tarekgh (approved):

src/TorchSharp/Tensor/Tensor.torch.cs
NiklasGustafsson: @tarekgh -- we do care. In fact, we should systematically go through all the APIs and turn on nullable, so that we introduce static type checking in most cases. There are a few, but very few, APIs that accept null Tensors.

tarekgh: Just FYI, C# compiler is introducing a new feature in version 11. Basically, append `!!` to the parameter name and the compiler will automatically apply the null check for you. I think this will be more interesting to use in these APIs.

Issue #581, opened 2022-04-23
Storage missing


MovGP0:
In torch there is the `storage()` method to get a `Storage` object that acts as a proxy to access the underlying data array in a linear fashion. 

**Example**
```csharp
var matrix = tensor(new double[] { 0, 1, 2, 3, 4, 5 }, new long[] {2L, 3L});
// matrix is now [[0, 1], [2, 3], [4, 5]]

var storage = matrix.storage();
storage[3] = 6;
// matrix is now [[0, 1], [2, 6], [4, 5]]
```

**See also**
* [Tensor.storage() method](https://pytorch.org/docs/stable/generated/torch.Tensor.storage.html)
* [Storage class](https://pytorch.org/docs/stable/storage.html)

MovGP0:
**Remark** 
It might be tricky to implement the setters properly, because the getter returns a point object in python: 

```python
matrix = torch.tensor([[4.0, 1.0], [5.0, 3.0], [2.0, 1.0]])
point = matrix[3]
matrix.storage_offset() # returns: 4
```

NiklasGustafsson:
Yes, Storage is one of the low-level features that has taken a backseat to other things. It's on the backlog now.

NiklasGustafsson:
BTW, @MovGP0, I'm curious about the use cases for Storage -- is it just linear indexing? What can you do with Tensor.storage() that you cannot do with 'Tensor.data()'?

Thanks,

Niklas

NiklasGustafsson:
> **Remark** the getter returns a Tensor object in python:

It looks like something else than a Tensor when I try:

```
>>> t.storage()
 1.0
 1.0
 1.0
 1.0
[torch.FloatStorage of size 4]
>>>
```

NiklasGustafsson:
There will be a couple of minor challenges with mapping the storage functionality to the .NET static type system:

1.  'storage()' may return one of several classes, each corresponding to the element type. The way I was planning on mapping this to .NET was to declare 'storage()' to return a non-generic base class, while the dynamic type is Storage<T>, where T is the .NET element type. This means that there will be some type-casting before you can use the Storage instance to get to data. For example:

```csharp
            var st = (Storage<long>)x.storage();
```

[...]

PR #587, opened 2022-04-29
Storage

NiklasGustafsson:
Implementing torch.storage().

See issue #581.


src/TorchSharp/Tensor/Storage.cs
```diff
@@ -0,0 +1,522 @@
...
+{
+    public static partial class torch
+    {
+        /// <summary>
+        /// A torch.Storage is a contiguous, one-dimensional array of a single data type.
+        /// Every tensor has a corresponding storage of the same data type.
+        /// </summary>
+        public abstract class Storage
+        {
+            internal Storage()
```
tarekgh, line 19: >internal [](http://example.com/codeflow?start=12&length=8)

should we have the `internal` as `protected` if these are going to be used by the subclasses only?

src/TorchSharp/Tensor/Storage.cs
```diff
@@ -0,0 +1,522 @@
...
+                case Type _ when type == typeof(float):
+                    return new Storage<T>(tensor.@float());
+                case Type _ when type == typeof(double):
+                    return new Storage<T>(tensor.@double());
+                case Type _ when type == typeof((float,float)):
+                    return new Storage<T>(tensor.to_type(ScalarType.ComplexFloat32));
+                case Type _ when type == typeof(System.Numerics.Complex):
+                    return new Storage<T>(tensor.to_type(ScalarType.ComplexFloat64));
+                default:
+                    throw new NotImplementedException();
```
tarekgh, line 56: >NotImplementedException [](http://example.com/codeflow?start=30&length=23)

nit: would be better using `NotSupportedException  instead in such places?

src/TorchSharp/Tensor/Storage.cs
```diff
@@ -0,0 +1,522 @@
...
+                data_ptr();
+            }
+
+            internal Storage(Tensor tensor, IntPtr data_ptr)
+            {
+                _tensor = tensor;
+                _tensor_data_ptr = (data_ptr == IntPtr.Zero) ? this.data_ptr() : data_ptr;
+            }
+
+            internal static Storage<T> CreateTypedStorageInstance<T>(Tensor tensor) where T : unmanaged
```
tarekgh, line 35: >CreateTypedStorageInstance [](http://example.com/codeflow?start=39&length=26)

nit: I wouldn't include the term `Typed` in the method name as the method is already a generic method which indicate that.

src/TorchSharp/Tensor/Storage.cs
```diff
@@ -0,0 +1,522 @@
...
+            protected static extern void THSStorage_set_nbytes(IntPtr tensor, UInt64 nbytes);
+            [DllImport("LibTorchSharp")]
+            protected static extern IntPtr THSStorage_data_ptr(IntPtr tensor);
+        }
+
+        /// <summary>
+        /// A torch.Storage is a contiguous, one-dimensional array of a single data type.
+        /// Every tensor has a corresponding storage of the same data type.
+        /// </summary>
+        public class Storage<T> : torch.Storage, IDisposable, IEnumerable<T> where T : unmanaged
```
tarekgh, line 177: >class [](http://example.com/codeflow?start=15&length=5)

Mark this as sealed if we think no-one want to subclass it. This will have some perf enhancement when it is sealed.

src/TorchSharp/Tensor/Storage.cs
```diff
@@ -0,0 +1,522 @@
...
+            /// <summary>
+            /// Accesses a slice of a storage instance.
+            /// </summary>
+            /// <param name="range">The range, expressed as a tuple [start,end)</param>
+            public Storage<T> this[(int? start, int? end) range] {
+                get {
+                    ValidateNotScalar();
+                    var start = CheckIndex(range.start.HasValue? range.start.Value : 0);
+                    var end = CheckIndex(range.end.HasValue ? range.end.Value : Count);
+                    var count = end - start;
```
tarekgh, line 282: >var count = end - start; [](http://example.com/codeflow?start=20&length=24)

should we validate `end >= start`?

src/TorchSharp/Tensor/Storage.cs
```diff
@@ -0,0 +1,522 @@
...
+                set {
+                    ValidateNotScalar();
+                    if (value._tensor is not null || value._tensor_data_ptr != IntPtr.Zero)
+                        throw new InvalidOperationException();
+                    var v = value._scalarValue;
+                    var start = CheckIndex(range.start.HasValue ? range.start.Value : 0);
+                    var end = CheckIndex(range.end.HasValue ? range.end.Value : Count);
+                    unsafe {
+                        T* ptr = (T*)_tensor_data_ptr;
+                        for (int i = start; i < end; i++)
```
tarekgh, line 297: >for (int i = start; i < end; i++) [](http://example.com/codeflow?start=24&length=33)

should we validate `end >= start`?

src/TorchSharp/Tensor/Storage.cs
```diff
@@ -0,0 +1,522 @@
...
+            }
+
+#if !NETSTANDARD2_0_OR_GREATER
+            /// <summary>
+            /// Accesses a slice of a storage instance.
+            /// </summary>
+            /// <param name="range">The range.</param>
+            public Storage<T> this[System.Range range] {
+                get {
+                    ValidateNotScalar();
```
tarekgh, line 310: >ValidateNotScalar [](http://example.com/codeflow?start=20&length=17)

do we need this validation as we already calling the other indexer which already doing the validation?

src/TorchSharp/Tensor/Storage.cs
```diff
@@ -0,0 +1,522 @@
...
+                    var start = CheckIndex(range.Start);
+                    var end = CheckIndex(range.End);
+                    this[(start, end)] = value;
+                }
+            }
+
+            private int CheckIndex(System.Index index)
+            {
+                var i = index.IsFromEnd ? Count - index.Value : index.Value;
+                if (i < 0 || i >= Count)
```
tarekgh, line 326: >if (i < 0 || i >= Count) [](http://example.com/codeflow?start=16&length=24)

nit: sometimes we optimize conditions like this by doing it as 

```C#
if ((uint)i >= Count)
    ...
```

src/TorchSharp/Tensor/Storage.cs
```diff
@@ -0,0 +1,522 @@
...
+                    for (int i = 0; i < Count; i++)
+                        if (((T*)ptr)[i].Equals(item)) return true;
+                }
+                return false;
+            }
+
+            public void CopyTo(IList<T> array, int arrayIndex)
+            {
+                ValidateNotScalar();
+                int idx = arrayIndex;
```
tarekgh, line 418: >idx [](http://example.com/codeflow?start=20&length=3)

Something looks wrong in this implementation. `idx` is defined and not used. also in the loop should `i` starts from 0?

I think this comment apply to CopyFrom too?

src/TorchSharp/Tensor/Storage.cs
```diff
@@ -0,0 +1,522 @@
...
+                return new StorageEnumerator(this);
+            }
+
+            public void Dispose()
+            {
+                _tensor_data_ptr = IntPtr.Zero;
+                _tensor = null;
+            }
+
+            private class StorageEnumerator: IEnumerator<T>
```
tarekgh, line 472: >class [](http://example.com/codeflow?start=20&length=5)

shoudl you have this as struct? just to avoid allocations if possible?

src/TorchSharp/Tensor/Storage.cs
```diff
@@ -0,0 +1,522 @@
...
+                public bool MoveNext()
+                {
+                    if (_index < 0) {
+                        _index = 0;
+                    } else if (++_index >= _count) {
+                        Reset();
+                        return false;
+                    } else {
+                        _index += 1;
+                    }
```
tarekgh, line 508: should this be simplified as:

```C#

if (_index >= _count - 1)
{
    return false;
}
_index++;

```

src/TorchSharp/Tensor/Storage.cs
NiklasGustafsson: Unless that's a big performance deal, I'd rather not. It's clever, but less readable.

MovGP0: This would cause an `OverflowException` when the calling code is in checked mode. ie. 

```csharp
bool TestCondition(int i, int count) => (uint)i >= count;

checked
{
	return TestCondition(-5, 3); // throws: OverfowException; Arithmetic operation resulted in an overflow.
}
```

lostmsu: @MovGP0 FYI you are wrong. `checked` affects only arithmetic operators inside the `checked`, it does not affect operators inside functions you are calling from `checked` block.

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -689,8 +711,12 @@ public Tensor cpu()
             /// Returns a copy of this object in CUDA memory.
             /// If this object is already in CUDA memory and on the correct device, then no copy is performed and the original object is returned.
             /// </summary>
-            public Tensor cuda()
+            public Tensor cuda(Device? device = null)
             {
+                if (device is null) {
+                    device = torch.CUDA;
```
lostmsu, line 717: This value does not seem to be used anywhere.

lostmsu (changes requested):

src/TorchSharp/Tensor/Tensor.cs
NiklasGustafsson: Thanks for catching that!

src/TorchSharp/Tensor/Storage.cs
```diff
@@ -0,0 +1,531 @@
...
+            {
+                _tensor = tensor;
+                _tensor_data_ptr = (data_ptr == IntPtr.Zero) ? this.data_ptr() : data_ptr;
+            }
+
+            internal static Storage<T> Create<T>(Tensor tensor) where T : unmanaged
+            {
+                var type = typeof(T);
+                switch (type) {
+                case Type _ when type == typeof(byte):
```
michaelgsharp, line 39: Are unsigned types not supported?

NiklasGustafsson: Nope.

src/TorchSharp/Tensor/Storage.cs
```diff
@@ -0,0 +1,531 @@
...
+
+            protected Storage(Tensor tensor, IntPtr data_ptr)
+            {
+                _tensor = tensor;
+                _tensor_data_ptr = (data_ptr == IntPtr.Zero) ? this.data_ptr() : data_ptr;
+            }
+
+            internal static Storage<T> Create<T>(Tensor tensor) where T : unmanaged
+            {
+                var type = typeof(T);
```
tarekgh, line 37: I forgot to mention one thing here, if you want to have better optimization for this method, try to change the `switch` statement into `if` statements. The JIT compiler optimization will kick off better when using  `if`.

PR #595, opened 2022-05-03
Started on torchvision.models.

NiklasGustafsson:
I added ResNet to torchvision.models. There's no direct support for pretrained weights, we'll need a place to store binary data that isn't in the repo itself, but for now, I'd appreciate comments on the UX.

I also added a 'skip' list to Module.load/save, which could be useful for not loading weights for some modules in a model, for example a last linear layer that will be replaced or has a different shape. 

Primary usage: transfer-learning.

Note that to skip a layer, all the state related to it must be skipped. For example, for a linear layer, both the 'weight' and 'bias,' if present, must be skipped.

With the new API, it should be possible to load a ResNet model that has a different number of classes from the one for which trained weights are available, simply by passing in 'fc.weight' and 'fc.bias' as the skip list when loading.

### DIFF
--- a/src/TorchSharp/NN/Module.cs
+++ b/src/TorchSharp/NN/Module.cs
@@ -106,6 +106,11 @@ namespace TorchSharp
                 protected virtual void Dispose(bool disposing)
                 {
                     if (disposing) {
+
+                        foreach (var (_,m) in named_modules()) {
+                            m.Dispose();
+                        }
+
                         handle.Dispose();
                         handle.SetHandleAsInvalid();
                         boxedModule?.Dispose();
--- a/src/TorchSharp/NN/Module.cs
+++ b/src/TorchSharp/NN/Module.cs
@@ -96,8 +96,8 @@ namespace TorchSharp
                 /// </summary>
                 public void Dispose()
                 {
-                    Dispose(true);
                     GC.SuppressFinalize(this);
+                    Dispose(true);
                 }
 
                 /// <summary>
@@ -105,7 +105,14 @@ namespace TorchSharp
                 /// </summary>
                 protected virtual void Dispose(bool disposing)
                 {
-                    if (disposing) {
+                    if (disposing && !handle.IsInvalid) {
+
+                        foreach (var (_, p) in named_buffers(false)) {
+                            p.Dispose();
+                        }
+                        foreach (var (_, b) in named_parameters(false)) {
+                            b.Dispose();
+                        }
 
                         foreach (var (_,m) in named_modules()) {
                             m.Dispose();
@@ -502,8 +509,8 @@ namespace TorchSharp
                     if (!recurse) yield break;
                     foreach (var (submoduleName, subModule) in _internal_submodules) {
                         foreach (var (parameterName, parameter) in subModule.named_parameters(true)) {
-                            if (seen.Contains(parameter.Handle)) continue;
-                            seen.Add(parameter.Handle);
+                            if (seen.Contains(parameter.handle)) continue;
+                            seen.Add(parameter.handle);
                             yield return ($"{submoduleName}.{parameterName}", parameter);
                         }
                     }
--- a/src/TorchSharp/NN/Module.cs
+++ b/src/TorchSharp/NN/Module.cs
@@ -418,7 +418,7 @@ namespace TorchSharp
                 {
                     List<string> missing = new List<string>();
                     List<string> unexpected = new List<string>();
-                    skip ??= new List<string>();
+                    skip ??= Array.Empty<string>();
 
                     var destination = state_dict();
 
@@ -823,7 +823,7 @@ namespace TorchSharp
                 /// </remarks>
                 public virtual Module load(System.IO.BinaryReader reader, bool strict = true, IList<string> skip = null)
                 {
-                    skip ??= new List<string>();
+                    skip ??= Array.Empty<string>();
 
                     var sd = state_dict();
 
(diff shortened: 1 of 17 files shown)


src/TorchSharp/NN/Module.cs
```diff
@@ -418,30 +412,35 @@ public virtual IEnumerable<(string name, Module module)> named_modules()
...
                 /// <param name="source">A dict containing parameters and persistent buffers.</param>
                 /// <param name="strict">Whether to strictly enforce that the keys in state_dict match the keys returned by this module’s state_dict() function.</param>
+                /// <param name="skipped">A list of keys not to consider when loading the dictionary.</param>
                 /// <returns></returns>
-                public virtual (IList<string> missing_keys, IList<string> unexpected_keyes) load_state_dict(Dictionary<string, Tensor> source, bool strict = true)
+                public virtual (IList<string> missing_keys, IList<string> unexpected_keyes) load_state_dict(Dictionary<string, Tensor> source, bool strict = true, IList<string> skipped = null)
                 {
                     List<string> missing = new List<string>();
                     List<string> unexpected = new List<string>();
+                    if (skipped is null) skipped = new List<string>();
```
tarekgh, line 421: >if (skipped is null) skipped = new List<string>(); [](http://example.com/codeflow?start=20&length=50)

nit: 

```C#
skipped ??= new List<string>();
```

NiklasGustafsson: All these new-fangled C# ways of dealing with null... :-)

Thanks!

tarekgh: Actually would be better to have something like:

```
privare static readonly s_emptyList = new List<string>();

.... C#
skipped ??= s_emptyList;
```

NiklasGustafsson: Sure. Array.Empty, perhaps?

src/TorchSharp/NN/Module.cs
```diff
@@ -720,30 +719,50 @@ public virtual Tensor forward(Tensor x, Tensor y)
                 /// Save the parameters and buffers of the module to a disk location.
                 /// </summary>
                 /// <param name="location">The file path.</param>
+                /// <param name="skip">A list of keys not to consider when saving the weights.</param>
                 /// <returns></returns>
-                public Module save(string location)
+                public Module save(string location, IList<string> skip = null)
```
tarekgh, line 724: >skip [](http://example.com/codeflow?start=66&length=4)

nit:
why we called this `skip` and in other places called `skipped`?

src/TorchSharp/NN/Module.cs
```diff
@@ -772,16 +797,34 @@ public Module load(string location, bool strict = true)
...
+                /// <param name="skip">A list of keys not to consider when loading the dictionary.</param>
+                /// <returns>The module, with parameters and buffers loaded.</returns>
+                /// <remarks>
+                /// Using a skip list only prevents tensors in the target module from being modified, it
+                /// does not alter any logic related to checking for matching tensor element types or entries.
+                /// It may be necessary to also pass 'strict=false' to avoid exceptions.
+                /// </remarks>
+                public virtual Module load(System.IO.BinaryReader reader, bool strict = true, IList<string> skip = null)
                 {
+                    if (skip == null) skip = new List<string>();
```
tarekgh, line 826: >if (skip == null) skip = new List<string>(); [](http://example.com/codeflow?start=20&length=44)

ditto as previous comment.

src/TorchSharp/TorchVision/models/ResNet.cs
```diff
@@ -0,0 +1,352 @@
...
+
+    namespace Modules
+    {
+        public class ResNet : Module
+        {
+            // The code here is is loosely based on https://github.com/kuangliu/pytorch-cifar/blob/master/models/resnet.py
+            // Licence and copypright notice at: https://github.com/kuangliu/pytorch-cifar/blob/master/LICENSE
+
+            private readonly long[] planes = new long[] { 64, 128, 128, 256, 256, 512, 512, 512, 512, 512, 512, 1024, 1024 };
+            private readonly long[] strides = new long[] { 1, 2, 1, 2, 1, 2, 1, 1, 1, 1, 1, 2, 1 };
```
tarekgh, line 139: are these can be static? or it can be mutated later?

src/TorchSharp/TorchVision/models/ResNet.cs
```diff
@@ -0,0 +1,352 @@
...
+                flatten = Flatten();
+                fc = Linear(512 * expansion, numClasses);
+
+                RegisterComponents();
+
+                if (device != null && device.type == DeviceType.CUDA)
+                    this.to(device);
+            }
+
+            private void MakeLayer(TorchSharp.Modules.ModuleList modules, Func<int, int, int, Module> block, int expansion, int planes, int num_blocks, int stride)
```
tarekgh, line 232: >planes [](http://example.com/codeflow?start=128&length=6)

using this name here is kind of confusing with the array defined with the same name above. would be nice if we can distinguish them.

src/TorchSharp/TorchVision/models/ResNet.cs
```diff
@@ -0,0 +1,352 @@
...
+                    (in_planes, planes, stride) => new Bottleneck(in_planes, planes, stride),
+                    Bottleneck.expansion, new int[] { 3, 4, 23, 3 },
+                    numClasses,
+                    device);
+            }
+
+            public static ResNet ResNet152(int numClasses, Device device = null)
+            {
+                return new ResNet(
+                    "ResNet101",
```
tarekgh, line 201: >ResNet101 [](http://example.com/codeflow?start=21&length=9)

did you mean ResNet152?

tarekgh (dismissed):

src/TorchSharp/NN/Module.cs
tarekgh: Even better as arrays implement IList.

PR #597, opened 2022-05-06
InceptionV3

NiklasGustafsson:


src/TorchSharp/NN/Pooling/AdaptiveAvgPool2D.cs
```diff
@@ -53,6 +53,42 @@ static public AdaptiveAvgPool2d AdaptiveAvgPool2d(long[] outputSize)
...
+
+            /// <summary>
+            /// Applies a 2D adaptive average pooling over an input signal composed of several input planes.
+            /// The output is of size H x W, for any input size.The number of output features is equal to the number of input planes.
+            /// </summary>
+            /// <param name="outputSize">The target output size (H,W) of the image of the form H x W.</param>
+            /// <returns></returns>
+            static public AdaptiveAvgPool2d AdaptiveAvgPool2d((long,long) outputSize)
+            {
+                long[] os = new long[] { outputSize.Item1, outputSize.Item2 };
```
tarekgh, line 65: >long [](http://example.com/codeflow?start=16&length=4)

does this array be alive or referenced after calling THSNN_AdaptiveAvgPool2d_ctor? 

Just asking if we can have it a stack allocated array?

src/TorchSharp/TorchVision/models/AlexNet.cs
```diff
@@ -81,14 +83,19 @@ public AlexNet(int numClasses, float dropout = 0.5f, torch.Device device = null)
...
                     Linear(4096, 4096),
                     ReLU(inPlace: true),
                     Linear(4096, numClasses)
                 );
 
                 RegisterComponents();
 
+                if (!string.IsNullOrEmpty(weights_file)) {
+
+                    this.load(weights_file, skip: skipfc ? new[] { "classifier.6.weight", "classifier.6.bias" } : null);
```
tarekgh, line 96: > new[] { "classifier.6.weight", "classifier.6.bias" } [](http://example.com/codeflow?start=58&length=53)

Is this a mutable array? I am asking to see if we can have it as static and avoid the allocations?

tarekgh, line 96: >new[] { "classifier.6.weight" [](http://example.com/codeflow?start=59&length=29)

is it possible to have the load method take a span instead of array and avoid the allocation?

NiklasGustafsson: load() is not a high-frequency method, since it's used just once in a while. I've tried to use 'IList<T>' for most APIs, in order to be as general as possible.

src/TorchSharp/NN/Pooling/AdaptiveAvgPool2D.cs
NiklasGustafsson: It applies to a lot of APIs, some yet to be implemented. They are also on high-frequency paths, so if this is something that has a big performance impact, then let's look at that.

src/TorchSharp/TorchVision/models/VGG.cs
```diff
@@ -365,24 +387,40 @@ public VGG(string name, int numClasses, bool batch_norm, float dropout = 0.5f, D
...
-                        break;
-                    case TorchSharp.Modules.Linear ln:
-                        torch.nn.init.normal_(ln.weight, 0, 0.01);
-                        torch.nn.init.constant_(ln.bias, 0);
-                        break;
                     }
+                } else {
+
+                    foreach (var (_, m) in named_modules()) {
+                        switch (m) {
```
tarekgh, line 414: >switch (m) { [](http://example.com/codeflow?start=24&length=12)

Is it expected to add more cases to this switch statement. I am asking if we can have this if statement instead?

tarekgh (dismissed):

src/TorchSharp/NN/Functional.cs
```diff
@@ -404,6 +404,50 @@ public static partial class functional
...
+                /// <param name="padding"></param>
+                /// <param name="dilation"></param>
+                /// <param name="ceil_mode"></param>
+                /// <returns></returns>
+                public static Tensor max_pool2d(Tensor input, long kernelSize, long? stride = null,
+                    long? padding = null, long? dilation = null, bool ceil_mode = false)
+                {
+                    return max_pool2d(input,
+                        new long[] { kernelSize },
+                        stride == null ? new long[] { 1, 1 } : new long [] { stride.Value, stride.Value },
```
tarekgh, line 422: do you think we can stackalloc the arrays here too? or this is not a hot code path? same comment for `max_pool2d` too.

tarekgh (approved):
LGTM. Thanks!

PR #602, opened 2022-05-14
adding tensor factory for arbitrary arrays

MovGP0:
adding a factory method for creating a tensor from an arbitrary array type. 

*Issues*
- some data types are not supported yet, since the appropriate factory methods are missing
- method is not named `tensor`, since the compiler cannot differentiate with the `tensor(IList<> ...)` methods in some cases when given the same name (ie. when the first parameter is of type `List<T>`)
- unit tests are missing, since i was unable to compile the project on my computer (insufficient disk space for the CUDA libs)

### DIFF
--- a/src/TorchSharp/Tensor/Tensor.Factories.cs
+++ b/src/TorchSharp/Tensor/Tensor.Factories.cs
@@ -2080,7 +2080,7 @@ namespace TorchSharp
         /// var tensor = torch.tensor_from_array(rawArray: array, dtype: torch.ScalarType.Float64, device: torch.Device.CPU, requiresGrad: false);
         /// </code>
         /// </example>
-        public static Tensor tensor_from_array(Array rawArray, ScalarType? dtype = null, Device? device = null, bool requiresGrad = false)
+        public static Tensor from_array(Array rawArray, ScalarType? dtype = null, Device? device = null, bool requiresGrad = false)
         {
             // enumerates over all dimensions of the arbitrary array
             // and returns the length of the dimension
--- a/src/TorchSharp/Tensor/Tensor.Factories.cs
+++ b/src/TorchSharp/Tensor/Tensor.Factories.cs
@@ -2045,6 +2045,87 @@ namespace TorchSharp
             return tensor(rawArray, new long[] { dim0, dim1, dim2, dim3 }, dtype, device, requiresGrad);
         }
 
+#nullable enable
+        /// <summary>
+        /// Creates a <see cref="torch.Tensor">torch tensor</see> from an arbitrary <see cref="Array">array</see>.
+        /// </summary>
+        /// <param name="rawArray">The arbitrary array to create the tensor from.</param>
+        /// <param name="dtype">The torch data type.</param>
+        /// <param name="device">The torch device.</param>
+        /// <param name="requiresGrad">Set <value>true</value> if gradients need to be computed for this Tensor; <value>false</value> otherwise.</param>
+        /// <returns>A <see cref="torch.Tensor">torch tensor</see></returns>
+        /// <exception cref="InvalidOperationException">
+        /// When <see cref="Type.GetElementType()">Array.GetType().GetElementType()</see> does not return the .NET element type.
+        /// </exception>
+        /// <exception cref="NotSupportedException">
+        /// When <see cref="Type.GetElementType()">Array.GetType().GetElementType()</see> returns an unsupported .NET element type.
+        /// Supported element types are <see cref="bool" />, <see cref="byte" />, <see cref="sbyte" />, <see cref="short" />,
+        /// <see cref="int" />, <see cref="long" />, <see cref="float" />, <see cref="double" />,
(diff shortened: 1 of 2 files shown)


NiklasGustafsson (changes requested):
Looks reasonable -- I would call it 'from_array,' though.

Also, please add unit tests.

NiklasGustafsson:
@MovGP0 -- I sent you a PR to your repo with some tests and a proposed name change.

src/TorchSharp/Tensor/Tensor.Factories.cs
```diff
@@ -2045,6 +2045,87 @@ public static Tensor tensor(IList<System.Numerics.Complex> rawArray, long dim0,
...
+            IEnumerable<object> GetElements(Array rawArray)
+            {
+                foreach (var element in rawArray) {
+                    yield return element!;
+                }
+            }
+            var elements = GetElements(rawArray);
+
+            // call the existing factory methods to construct the tensor
+            if (t == typeof(bool)) return tensor(elements.Cast<bool>().ToArray(), shape, dtype, device, requiresGrad);
```
NiklasGustafsson, line 2108: Does the Cast<T>().ToArray() lead to a copy being made, I wonder?

src/TorchSharp/Tensor/Tensor.Factories.cs
```diff
@@ -2045,6 +2045,87 @@ public static Tensor tensor(IList<System.Numerics.Complex> rawArray, long dim0,
...
+                    var dimLength = arr.GetLength(dim);
+                    yield return dimLength;
+                }
+            }
+            var shape = GetShape(rawArray).ToArray();
+
+            var t = rawArray.GetType().GetElementType();
+            if (t is null) throw new InvalidOperationException($"{nameof(rawArray)}.GetType().GetElementType() returned null.");
+
+            IEnumerable<object> GetElements(Array rawArray)
```
NiklasGustafsson, line 2099: This function seem unnecessary. 'Array' already implements IEnumerable, so this just adds overhead, I think.

src/TorchSharp/Tensor/Tensor.Factories.cs
```diff
@@ -2045,6 +2045,87 @@ public static Tensor tensor(IList<System.Numerics.Complex> rawArray, long dim0,
...
+            if (t == typeof(long)) return tensor(elements.Cast<long>().ToArray(), shape, dtype, device, requiresGrad);
+#if NET50_OR_GREATER
+            // TODO: implement the required factory method
+            // if (t == typeof(half)) return tensor(elements.Cast<half>().ToArray(), shape, dtype, device, requiresGrad);
+#endif
+            if (t == typeof(float)) return tensor(elements.Cast<float>().ToArray(), shape, dtype, device, requiresGrad);
+            if (t == typeof(double)) return tensor(elements.Cast<double>().ToArray(), shape, dtype, device, requiresGrad);
+            // TODO: implement the required factory methods
+            // if (t == typeof(Tuple<float, float>)) return tensor(elements.Cast<Tuple<float, float>>().ToArray(), shape, dtype, device, requiresGrad);
+            // if (t == typeof(Tuple<double, double>)) return tensor(elements.Cast<Tuple<double, double>>().ToArray(), shape, dtype, device, requiresGrad);
```
NiklasGustafsson, line 2122: We don't use (double,double) to represent 64-bit complex, and the (float,float) factory methods should already exist.

NiklasGustafsson:
@MovGP0 -- just a friendly "ping." I think this is a great addition to the APIs, but I had some comments on it last week.

MovGP0:
> @MovGP0 -- just a friendly "ping." I think this is a great addition to the APIs, but I had some comments on it last week.

much appreciated. Just very busy these weeks.

src/TorchSharp/Tensor/Tensor.Factories.cs
MovGP0: yes, the `.ToArray()` always creates a new array: 
 ```csharp
 var array = new int[]{1,2,3,4};
var newArray = array.Cast<int>().ToArray();
Console.WriteLine(ReferenceEquals(array, newArray)); // false
```

 ```csharp
 var array = new int[]{1,2,3,4};
var newArray = array.ToArray();
Console.WriteLine(ReferenceEquals(array, newArray)); // false
```

MovGP0:
> Looks reasonable -- I would call it 'from_array,' though.

I was thinking about that and opted against it. The more verbose `tensor_from_array` makes the method easier to discover (think of the autocomplete when someone types `torch.tens...`) and makes its function more clear.

NiklasGustafsson:
> * fixing documentation to reflect the name change to `from_array`

I thought you didn't agree with that.

NiklasGustafsson:
I'm going to merge this one as soon as the restarted checks pass. We can always change names later, before the next release.

NiklasGustafsson (approved):

PR #607, opened 2022-05-23
Enable TorchSharp assembly name signing

tarekgh:
Enable name signing for TorchSharp assembly. This will allow avoiding any runtime failure when running with .NET Framework with apps using strong name signing. Otherwise, can run into situations getting exception like the following:

```
System.IO.FileLoadException : Could not load file or assembly 'TorchSharp, Version=0.96.6.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. A strongly-named assembly is required. (Exception from HRESULT: 0x80131044)
```


tarekgh:
CC @NiklasGustafsson @michaelgsharp @ericstj

NiklasGustafsson (approved):

PR #608, opened 2022-05-23
Adding a couple of APIs from torch.nn.utils

NiklasGustafsson:
clip_grad_value_
parameters_to_vector
vector_to_parameters


RELEASENOTES.md
```diff
@@ -7,6 +7,7 @@ Releases, starting with 9/2/2021, are listed with the most recent release at the
 __API Changes:__
 
 Added the 'META' device type, which can be used to examine the affect of shape from tensor operations without actually doing any computations.<br/>
+Added a few methods from the torch.nn.utils namespace.<br/>
```
tarekgh, line 11: should we mention we have made the TorchSharp assembly name signed now?

src/TorchSharp/Torch.cs
```diff
@@ -304,26 +304,82 @@ public static partial class nn
             public static partial class utils
             {
                 [DllImport("LibTorchSharp")]
-                extern static double THSTensor_clip_grad_norm_(IntPtr tensor, int len, double max_norm, double norm_type);
+                extern static double THSTensor_clip_grad_norm_(IntPtr tensors, int len, double max_norm, double norm_type);
+
+                [DllImport("LibTorchSharp")]
```
tarekgh, line 309: nit: for the sake of coding style, could you please decorate these methods with "private" specifiers? It is always better to be explicit about that.

src/TorchSharp/Torch.cs
```diff
@@ -304,26 +304,82 @@ public static partial class nn
...
                 /// <param name="max_norm"></param>
                 /// <param name="norm_type"></param>
-                /// <returns></returns>
+                /// <remarks>Gradients are modified in-place.</remarks>
                 public static double clip_grad_norm_(IEnumerable<Modules.Parameter> tensors, double max_norm, double norm_type = 2.0)
+                {
+                    using (var parray = new PinnedArray<IntPtr>()) {
+                        IntPtr tensorsRef = parray.CreateArray(tensors.Select(p => p.Handle).ToArray());
+                        var value = THSTensor_clip_grad_norm_(tensorsRef, parray.Array.Length, max_norm, norm_type);
+                        CheckForErrors();
```
tarekgh, line 331: >CheckForErrors [](http://example.com/codeflow?start=24&length=14)

Why we don't include the results of the CheckOfErrors inside the pinvoked method like THSTensor_clip_grad_norm_? 

I am asking as this can save pinvoke call. I am not expecting you change that in this PR but may be something we can track to improve?

src/TorchSharp/Torch.cs
```diff
@@ -304,26 +304,82 @@ public static partial class nn
...
+
+                /// <summary>
+                /// Clips gradient of an iterable of parameters at specified value.
+                /// </summary>
+                /// <param name="tensors">An enumeration of Tensors that will have gradients normalized</param>
+                /// <param name="clip_value">Maximum allowed value of the gradients. The gradients are clipped in the range [-clip_value,clip_value]</param>
+                /// <remarks>Gradients are modified in-place.</remarks>
+                public static void clip_grad_value_(IEnumerable<Modules.Parameter> tensors, double clip_value)
+                {
+                    using (var parray = new PinnedArray<IntPtr>()) {
```
tarekgh, line 344: nit: just to mention, C# now has the syntax like 


using var parray = new PinnedArray<IntPtr>();

which will automatically call the dispose method when exiting from the declared block. This is helpful if you want to include the code inside { }

RELEASENOTES.md
NiklasGustafsson: Sure. What should the comment be? Don't know assembly-name signing enough.

src/TorchSharp/Torch.cs
NiklasGustafsson: I know, but I like the old style better. I think it makes it clearer what the scope of the using is, and more readable.

tarekgh (approved):

src/TorchSharp/Torch.cs
NiklasGustafsson: I'm not sure I understand what you are suggesting. The CheckForErrors() method transfers the native exception information, if there is one, to managed code and throws a managed-code exception.

RELEASENOTES.md
tarekgh: Just a simple comment like: Strongly name signing the TorchSharp library to allow loading it in .NET Framework strongly named apps. 

https://docs.microsoft.com/en-us/dotnet/standard/assembly/strong-named

src/TorchSharp/Torch.cs
tarekgh: `CheckForErrors` is calling the following pinvoke method

```C#
            var error = THSTorch_get_and_reset_last_err();
```

You can have the native implementation of methods like `THSTensor_clip_grad_norm_` to call `THSTorch_get_and_reset_last_err()` and return the error as out parameter. Doing that, the code will look something like:

```C#
 var value = THSTensor_clip_grad_norm_(tensorsRef, parray.Array.Length, max_norm, norm_type out error);
 Throw_If_Error(error); // completely managed method
```

src/TorchSharp/Torch.cs
tarekgh: fine as you long you prefer that. I only frankly disagree with the part `and more readable` :-).

src/TorchSharp/Torch.cs
NiklasGustafsson: Got it. That is a huge change, IMO. Maybe it's worth it, but definitely something we would separately, as you mention.

RELEASENOTES.md
NiklasGustafsson: I'll add that to the release notes. FYI, I've taken to adding comments to the release notes with every PR I create.

tarekgh: Thanks for letting me know. I am wondering how frequently we update the NuGet packages? I am asking because @michaelgsharp is waiting for the name signing change I did.

NiklasGustafsson: Got it. There's no set release schedule. I make a release whenever there's a critical mass of changes, or there are just a few very important things to get out.

NiklasGustafsson: We could probably make a release later this week.

tarekgh: This will be great if we do it.

NiklasGustafsson: Okay, let's do that, then.

Issue #471, opened 2021-11-30
torchvision.io missing


oluwandabira:
Are there any plans to implement this in TorchSharp?

NiklasGustafsson:
Possibly, but not exactly like the PyTorch version. We will make sure that it composes well with .NET I/O.

oluwandabira:
Can I try my hand at implementing this?
I'd like this feature sooner than later and I think I have an idea of how to implement `read_image` (although that idea involves `System.Drawing` like ML.Net's `ImageAnalytics.ImageLoader` and my use case is primarily on Linux...)

NiklasGustafsson:
Yes, of course. That would be welcome.

I suggest that you start by writing code examples against a mock API, just so that we can discuss the user experience before you go implementing a whole bunch of code.

NiklasGustafsson:
Please note that it'd be best to introduce some sort of abstraction so that we can support more than one .NET image library. The ImageProcessing example uses SkiaSharp, but that's just one of several relevant image libraries.

NiklasGustafsson:
BTW, it may be time to move torchvision into its own DLL / package soon. No need to hold this up for that decision, but it's something we should consider. It's what the Pytorch distribution does, after all.

NiklasGustafsson:
One big decision to make is how to support the .NET async I/O pattern.

oluwandabira:
I am unsure of where to put the code examples for the mock api. Should I just write them here, open a pr, discussion, gist etc?

NiklasGustafsson:
A new discussion thread would be most appropriate. 

Maybe upload the files to a "junk" branch in your repo, and link to them?

dayo05:
System.Drawing has removed from non-Windows OS from .NET6, for alternative, ImageSharp is one of the recommended method, this is example.

(I'm not sure this is best way)
```cs
Tensor ReadImage(string image)
{
    var img = Image.Load<Rgb24>(image, new JpegDecoder());
    return cat(new[] {ReadImageChannel(img, "R"), ReadImageChannel(img, "G"), ReadImageChannel(img, "B")}, 0).unsqueeze(0);
}


Tensor ReadImageChannel(Image<Rgb24> image, string channel)
=> tensor(image.GetPixelMemoryGroup()[0].Span.ToArray().Select(x => channel switch
{
    "R" => x.R,
    "G" => x.G,
    "B" => x.B,
[...]

PR #609, opened 2022-05-25
Add torchvision.io

oluwandabira:
Potential fix for #471 .

A few things need to be considered before this can be ready for review (aside from the unimplemented methods and documentation):

- Whether `IVideor` should remain. Frankly I don't know any .NET libraries that can serve as an implementation of that.
- Whether `ImageFormat` should remain as is with just `PNG`, `JPEG` and `UNKNOWN` options. This works fine for `ImagerSharp` but may not be accurate for other implementations that supports more image formats.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/TorchSharp?pullRequest=609) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/TorchSharp?pullRequest=609) before we can accept your contribution.<br/><br/>:x: oluwandabira [sign now](https://cla.dotnetfoundation.org/dotnet/TorchSharp?pullRequest=609)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/TorchSharp?pullRequest=609) it.</sub>

oluwandabira:
@NiklasGustafsson The build fails because `File.ReadAllBytesAsync` is not part of netstandard2.0 . What do you recommend I change it to?

NiklasGustafsson:
> @NiklasGustafsson The build fails because `File.ReadAllBytesAsync` is not part of netstandard2.0 . What do you recommend I change it to?

@tarekgh -- do you have any recommendation for @oluwandabira?

tarekgh:
@oluwandabira you may implement `File.ReadAllBytesAsync` using [Stream.ReadAsync](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.readasync?view=net-6.0#system-io-stream-readasync(system-byte()-system-int32-system-int32-system-threading-cancellationtoken)).

NiklasGustafsson:
@oluwandabira -- this is an important additions, so we will probably take a little time reviewing.

Thanks for doing the PR!

NiklasGustafsson:
> * Whether `ImageFormat` should remain as is with just `PNG`, `JPEG` and `UNKNOWN` options. This works fine for `ImagerSharp` but may not be accurate for other implementations that supports more image formats.

I would suggest extending the type with more formats, and then have each implementation throw an exception for formats that are not supported.

src/Examples/Program.cs
```diff
@@ -6,12 +6,13 @@ public static class Program
     {
         public static void Main(string[] args)
         {
-            MNIST.Main(args);
-            AdversarialExampleGeneration.Main(args);
-            CIFAR10.Main(args);
-            SequenceToSequence.Main(args);
-            TextClassification.Main(args);
+            //MNIST.Main(args);
```
NiklasGustafsson, line 9: Should these be left commented?

src/TorchSharp/TorchVision/File.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+using static TorchSharp.torch;
+
+namespace TorchSharp.torchvision
+{
+    partial class io
+    {
+        public static Tensor read_file(string filename)
```
NiklasGustafsson, line 11: What are these? Neither is implemented.

src/TorchSharp/TorchVision/Video.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.IO;
+using System.Threading.Tasks;
+
+using static TorchSharp.torch;
+
+namespace TorchSharp.torchvision
+{
+    public static partial class io
+    {
+        public interface IVideoer
```
NiklasGustafsson, line 13: Based on your comment, let's leave video processing out until someone proposes it.

NiklasGustafsson (changes requested):

NiklasGustafsson (changes requested):
I've added comments in a number of places with questions and suggested changes.

src/TorchSharp/TorchVision/Image.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        public static IImager DefaultImager { get; set; } = new ImagerSharp();
+
+        public enum ImageFormat
+        {
+            PNG,
+            JPEG,
+            UNKNOWN
+        }
+        public interface IImager
```
tarekgh, line 21: >interface [](http://example.com/codeflow?start=15&length=9)

Is it expected to extend this interface in the future? If so, this can be issue for .NET Framework.

You can just make this as abstract class which give flexibility to extend in the future.

src/TorchSharp/TorchVision/Image.cs
```diff
@@ -0,0 +1,72 @@
...
+using static TorchSharp.torch;
+
+namespace TorchSharp.torchvision
+{
+    public static partial class io
+    {
+
+        public static IImager DefaultImager { get; set; } = new ImagerSharp();
+
+        public enum ImageFormat
```
tarekgh, line 15: >ImageFormat [](http://example.com/codeflow?start=20&length=11)

nay reason these enums are defined inside the io class? do we expect these can be used in other scenarios?

src/TorchSharp/TorchVision/Image.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            PNG,
+            JPEG,
+            UNKNOWN
```
tarekgh, line 19: >UNKNOWN [](http://example.com/codeflow?start=12&length=7)

Maybe moving this as first entry will be better when adding more formats to this enum.

src/TorchSharp/TorchVision/Image.cs
```diff
@@ -0,0 +1,72 @@
...
+namespace TorchSharp.torchvision
+{
+    public static partial class io
+    {
+
+        public static IImager DefaultImager { get; set; } = new ImagerSharp();
+
+        public enum ImageFormat
+        {
+            PNG,
```
tarekgh, line 17: >PNG [](http://example.com/codeflow?start=12&length=3)

Can we use the letter casing as we do in [https://docs.microsoft.com/en-us/dotnet/api/system.drawing.imaging.imageformat?view=dotnet-plat-ext-6.0#properties](https://docs.microsoft.com/en-us/dotnet/api/system.drawing.imaging.imageformat?view=dotnet-plat-ext-6.0#properties)?

src/TorchSharp/TorchVision/Image.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            ImageFormat DetectFormat(byte[] image);
+            Tensor DecodeImage(byte[] image, ImageReadMode mode = ImageReadMode.UNCHANGED);
+            Tensor DecodeImage(byte[] image, ImageFormat format, ImageReadMode mode = ImageReadMode.UNCHANGED);
+            byte[] EncodeImage(Tensor image, ImageFormat format);
+        }
+
+        public enum ImageReadMode
+        {
+            UNCHANGED,
```
tarekgh, line 31: >UNCHANGED [](http://example.com/codeflow?start=12&length=9)

Maybe using C# style casing will be better here? i.e. Unchanged, Gray, GrayAlpha, Rgb...etc.

oluwandabira: `ImageReadMode` and its options were written to resemble their [pytorch version](https://pytorch.org/vision/stable/generated/torchvision.io.ImageReadMode.html#torchvision.io.ImageReadMode) but if it is decided that we should change those then we should probably also change the function names as well.

NiklasGustafsson: Generally, I'm in favor of keeping the Pytorch casing wherever it enables 'copy-and-paste' from Python. If we're replacing a string with an enum, it's not as clear.

src/TorchSharp/TorchSharp.csproj
```diff
@@ -22,7 +22,8 @@
 
   <ItemGroup>
     <PackageReference Include="SharpZipLib" Version="1.3.3" />
-    <PackageReference Include="System.Memory" Version="4.5.3" />
+    <PackageReference Include="SixLabors.ImageSharp" Version="2.1.1" />
```
tarekgh, line 25: >SixLabors [](http://example.com/codeflow?start=31&length=9)

@NiklasGustafsson ImageSharp library is under Apache 2.0 license. Do we need to validate anything here before using the library?

src/TorchSharp/TorchVision/Image.cs
oluwandabira: `ImageReadMode` is an enum in pytorch.

src/TorchSharp/TorchVision/Image.cs
oluwandabira: Could you explain a bit or point to where I can learn why this would be an issue? I'll change it to an abstract class, I'm just not familiar enough with .NET to see the problem.

src/Examples/IOReadWrite.cs
```diff
@@ -0,0 +1,63 @@
...
+            }
+        }
+
+        internal static void Main(string[] args)
+        {
+            var filename = args[0];
+
+            Console.WriteLine($"Reading file {filename}");
+
+            //torchvision.io.DefaultImager = new SkiaImager();
```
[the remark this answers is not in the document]
oluwandabira: It should be included, I just haven't gotten around to implementing it. Never worked with Skia.

src/TorchSharp/TorchSharp.csproj
NiklasGustafsson: For a package dependency? I don't think so. It would be different if we used their source code, I think.

src/TorchSharp/TorchVision/Image.cs
NiklasGustafsson: They should be in the same scope as they are in Pytorch, so that the qualified name is the same.

src/TorchSharp/TorchVision/Image.cs
tarekgh: [Default Interface Methods](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/default-interface-methods?ranMID=43674&ranEAID=rl2xnKiLcHs&ranSiteID=rl2xnKiLcHs-YlRlaRrWTrEFNgBuwYx6TQ&epi=rl2xnKiLcHs-YlRlaRrWTrEFNgBuwYx6TQ&irgwc=1&OCID=AID2200057_aff_7795_1243925&tduid=(ir__ixfocyqv29kfqmcc1aulhv0fbf2xv1wna3xvqwfy00)(7795)(1243925)(rl2xnKiLcHs-YlRlaRrWTrEFNgBuwYx6TQ)()&irclickid=_ixfocyqv29kfqmcc1aulhv0fbf2xv1wna3xvqwfy00#summary) is the feature added at some point for .NET Core. But this is not supported on the .NET Framework. If any user build any library/app running on .NET Framework, then the interface gets extended in the future, such library or app can break because of that.

src/TorchSharp/TorchSharp.csproj
tarekgh: That is fine then. I was wondering because I am seeing the following in the NuGet package page

image

src/TorchSharp/TorchVision/Image.cs
tarekgh: Ok then to keep the same casing as pytorch then.

src/TorchSharp/TorchSharp.csproj
NiklasGustafsson: I don't think it's a licensing issue, but we may want to think about whether actual implementations (not the interfaces) of image I/O belong in a different NuGet package.

NiklasGustafsson: I know @michaelgsharp has been looking at this for ML.NET. Michael, could you comment / weigh in?

src/Examples/IOReadWrite.cs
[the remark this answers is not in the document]
NiklasGustafsson: In the 'ImageTransform' example, there's a simplistic image reader / writer based on Skia.

src/TorchSharp/TorchSharp.csproj
tarekgh: We are planning to do some work in ML.NET to use ImageSharp but we haven't gotten into that yet.

NiklasGustafsson: If we were taking a dependency on a commercial library, that would be one thing. This one has both an OSS license and a commercial one, so we're not forcing users of TorchSharp to go with a commercial license.

src/Examples/IOReadWrite.cs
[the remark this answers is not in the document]
oluwandabira: I've added a simple implementation.

NiklasGustafsson:
@oluwandabira -- I suggest that we move this to TorchSharpExamples for the time being. It's valuable, but we don't want to commit to a particular image library dependency at this time. As I said, there's a broader discussion going on for .NET, and ML.NET, regarding imaging, and it remains unresolved.

oluwandabira:
So should I move everything in `torchvision.io` or just the `ImagerSharp` implementation to TorchSharpExamples as it is the only thing that depends on an image library?

NiklasGustafsson:
@oluwandabira  -- that is a good idea. 

I think it would make sense to have everything except the implementations here, and the rest in TorchSharpExamples. Basically, we should not (at this point) take a dependency on either SkiaSharp or ImageSharp in this library, but we can do so in the examples, since we're not distributing a binary library from there.

src/TorchSharp/TorchVision/Image.cs
```diff
@@ -0,0 +1,177 @@
...
+        /// Used by <cref>torchvision.io</cref> image methods.
+        /// </summary>
+        public abstract class Imager
+        {
+            /// <summary>
+            /// Reads the contents of an image file and return the image's file format.
+            /// </summary>
+            /// <param name="image">Image file contents.</param>
+            /// <returns>The detected image format.</returns>
+            public abstract ImageFormat DetectFormat(byte[] image);
```
NiklasGustafsson, line 26: I think this method should disappear. In the SkiaImage example, it requires an additional copy of the data. It's better to have DecodeImage decide how it's going to figure out what format it is in the best way for each implementation. It can still be a private method on the ImageSharp implementation.

src/TorchSharp/TorchVision/Image.cs
```diff
@@ -0,0 +1,177 @@
...
+            /// </summary>
+            /// <param name="image">Image file contents.</param>
+            /// <param name="mode">Image read mode.</param>
+            /// <remarks>
+            /// The image format is detected from image file contents.
+            /// </remarks>
+            /// <returns>
+            /// <cref>Tensor</cref> with <c>shape = [color_channels, image_height, image_width]</c> and <c>dtype = uint8</c>.
+            /// </returns>
+            public Tensor DecodeImage(byte[] image, ImageReadMode mode = ImageReadMode.UNCHANGED)
```
NiklasGustafsson, line 38: I think it makes sense to add an overload that takes a stream, in addition to the one taking a byte[].

src/TorchSharp/TorchVision/Image.cs
NiklasGustafsson: That means making all the DecodeImage methods abstract.

src/TorchSharp/TorchVision/Image.cs
```diff
@@ -0,0 +1,177 @@
...
+            /// <cref>Tensor</cref> with <c>shape = [color_channels, image_height, image_width]</c> and <c>dtype = uint8</c>.
+            /// </returns>
+            public abstract Tensor DecodeImage(byte[] image, ImageFormat format, ImageReadMode mode = ImageReadMode.UNCHANGED);
+            /// <summary>
+            /// Encodes a <cref>Tensor</cref> with <c>shape = [color_channels, image_height, image_width]</c> into an array of bytes.
+            /// </summary>
+            /// <param name="image"><cref>Tensor</cref> with <c>shape = [color_channels, image_height, image_width]</c>.</param>
+            /// <param name="format">Image format.</param>
+            /// <returns>The encoded image.</returns>
+            public abstract byte[] EncodeImage(Tensor image, ImageFormat format);
```
NiklasGustafsson, line 58: Maybe add an EncodeImage that takes a stream?

src/TorchSharp/TorchVision/Image.cs
```diff
@@ -0,0 +1,177 @@
...
+using static TorchSharp.torch;
+
+namespace TorchSharp.torchvision
+{
+    public static partial class io
+    {
+        /// <summary>
+        /// <cref>Imager</cref> to be used when a <cref>torchvision.io</cref> image method's <c>imager</c> is unspecified.
+        /// </summary>
+        public static Imager DefaultImager { get; set; }
```
NiklasGustafsson, line 13: By removing all implementations, this could be null. Would we be better off with a skeleton implementation that just throws an exception, but one that is a little bit more helpful than 'NullReferenceException'?

NiklasGustafsson (changes requested):

dayo05:
This is just suggestion.

If this approved, basic implementation of some libraries(like ImageSharp, System.Drawing(only available on windows) need to move on TorchSharp, not TorchSharp/Examples. This requires new dependency and System.Drawing has removed to non Windows platforms. Maybe like TorchSharp.ImageSharp, style package is looks good.

NiklasGustafsson:
> This is just suggestion.
> 
> If this approved, basic implementation of some libraries(like ImageSharp, System.Drawing(only available on windows) need to move on TorchSharp, not TorchSharp/Examples. This requires new dependency and System.Drawing has removed to non Windows platforms. Maybe like TorchSharp.ImageSharp, style package is looks good. This change will good when API has breaking changes, user don't need to copy-paste from examples, just updating nuget-package.

@dayo05 -- thank you for your input. I agree with you in principle, but there are complications.

As discussed earlier in this PR comment thread, the question of which image library to recommend and/or take a dependency on is one that we are grappling with across several .NET libraries. Until we have some resolution of that discussion, it would be inappropriate to take a binary dependency on one of them in TorchSharp.

src/TorchSharp/TorchVision/Image.cs
oluwandabira: Or did you mean an `EncodeImage` image that takes a stream to write to and doesn't return a byte array?

NiklasGustafsson: Sorry, yes, that's what I meant -- convert directly from a Tensor into a stream.

oluwandabira:
I've added `Stream` overloads to the non-async io methods but I'm not sure about the async ones. I can add them (I think) but it would involve adding async methods to the `Imager` class.

oluwandabira:
Now I'm not so sure that the current async methods are actually async.

NiklasGustafsson:
> Now I'm not so sure that the current async methods are actually async.

It looks to me like they are -- you are using async I/O to fill the buffer, which is then converted synchronously, which is the only available way, since there's no I/O involved. It would be different if you used the synchronous stream APIs.

oluwandabira:
Should I add async stream methods (read_image_async and write_image_async that take streams)?

oluwandabira:
Added them and tweaked the other async methods.

oluwandabira:
The build failure doesn't seem to be my fault.

NiklasGustafsson (approved):

PR #620, opened 2022-06-07
implementing round with decimals

MovGP0:
Open issues: 
* C++ binding might be missing; please check in C++ source
* check if test is working

### DIFF
--- a/src/TorchSharp/Tensor/Tensor.Math.cs
+++ b/src/TorchSharp/Tensor/Tensor.Math.cs
@@ -1743,6 +1743,9 @@ namespace TorchSharp
             [DllImport("LibTorchSharp")]
             static extern IntPtr THSTensor_round(IntPtr tensor);
 
+            [DllImport("LibTorchSharp")]
+            static extern IntPtr THSTensor_round(IntPtr tensor, long decimals);
+
             /// <summary>
             /// Returns a new tensor with each of the elements of input rounded to the closest integer.
             /// </summary>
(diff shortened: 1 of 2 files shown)


src/TorchSharp/Tensor/Tensor.Math.cs
```diff
@@ -1743,6 +1743,9 @@ public Tensor remainder_(Scalar scalar)
             [DllImport("LibTorchSharp")]
             static extern IntPtr THSTensor_round(IntPtr tensor);
 
+            [DllImport("LibTorchSharp")]
```
NiklasGustafsson, line 1746: Since THSTensor_round() is an 'extern "C"' function in the C++ code, we can't overload these. We should simply add the integer argument to the existing function, add it in the C++ code, and then call the C++ function that takes the decimals argument. Then apply the default '0' value to the C# method.

MovGP0: I've removed the round method without the decimals parameter and replaced it with a default value.

MovGP0: unfortunately C++ programming is something I'm unable to do. Especially since the build would fill up my SSD. 

I would be happy about help with this one, until I know enough to do it myself.

NiklasGustafsson: I get it, that's what I think I recall from an earlier conversation.

I sent you a PR against your repo with the C++ changes. I can just take over here and push those changes, or you can merge into your 'main'

Issue #621, opened 2022-06-07
torch.roll missing


oluwandabira:
[torch.roll / Tensor.roll](https://pytorch.org/docs/stable/generated/torch.roll.html) is missing.

PR #623, opened 2022-06-08
Adding Tensor.roll and torch.roll #621

NiklasGustafsson:
Fixed issue #621

### DIFF
--- a/src/TorchSharp/Tensor/Tensor.cs
+++ b/src/TorchSharp/Tensor/Tensor.cs
@@ -5185,13 +5185,13 @@ namespace TorchSharp
             /// Elements that are shifted beyond the last position are re-introduced at the first position.
             /// If a dimension is not specified, the tensor will be flattened before rolling and then restored to the original shape.
             /// </summary>
-            public Tensor roll(IEnumerable<long> shifts, IEnumerable<long>? dims = null)
+            public Tensor roll(ReadOnlySpan<long> shifts, IEnumerable<long>? dims = null)
             {
-                var shLen = shifts.Count();
+                var shLen = shifts.Length;
                 var dmLen = dims is null ? 0 : dims.Count();
 
                 unsafe {
-                    fixed (long* sh = shifts.ToArray(), dm = (dims == null) ? null : dims.ToArray()) {
+                    fixed (long* sh = shifts, dm = (dims == null) ? null : dims.ToArray()) {
                         var res =
                             THSTensor_roll(Handle, (IntPtr)sh, shLen, (IntPtr)dm, dmLen);
                         if (res == IntPtr.Zero) { torch.CheckForErrors(); }
--- a/src/TorchSharp/Tensor/Tensor.cs
+++ b/src/TorchSharp/Tensor/Tensor.cs
@@ -5143,6 +5143,64 @@ namespace TorchSharp
             }
 
             [DllImport("LibTorchSharp")]
+            extern static IntPtr THSTensor_roll(IntPtr tensor, IntPtr shifts, int shLength, IntPtr dims, long dimLength);
+
+            /// <summary>
+            /// Roll the tensor along the given dimension(s).
+            /// Elements that are shifted beyond the last position are re-introduced at the first position.
+            /// If a dimension is not specified, the tensor will be flattened before rolling and then restored to the original shape.
+            /// </summary>
+            public Tensor roll(long shifts, long? dims = null)
+            {
+                if (dims.HasValue) {
+                    return roll(new long[] { shifts }, new long[] { dims.Value });
+                }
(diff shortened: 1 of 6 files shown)


src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -5142,6 +5142,64 @@ public IList<Tensor> nonzero_as_list()
...
+                    return roll(new long[] { shifts }, null);
+                }
+            }
+
+            /// <summary>
+            /// Roll the tensor along the given dimension(s).
+            /// Elements that are shifted beyond the last position are re-introduced at the first position.
+            /// If a dimension is not specified, the tensor will be flattened before rolling and then restored to the original shape.
+            /// </summary>
+            public Tensor roll((long,long) shifts, (long,long) dims)
```
tarekgh, line 5168: >, [](http://example.com/codeflow?start=36&length=1)

nit: insert space.

tarekgh, line 5168: >, [](http://example.com/codeflow?start=56&length=1)

nit: insert space.

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -5142,6 +5142,64 @@ public IList<Tensor> nonzero_as_list()
...
+            {
+                return roll(new long[] { shifts.Item1, shifts.Item2, shifts.Item3 }, new long[] { dims.Item1, dims.Item2, dims.Item3 });
+            }
+
+            /// <summary>
+            /// Roll the tensor along the given dimension(s).
+            /// Elements that are shifted beyond the last position are re-introduced at the first position.
+            /// If a dimension is not specified, the tensor will be flattened before rolling and then restored to the original shape.
+            /// </summary>
+            public Tensor roll(IEnumerable<long> shifts, IEnumerable<long>? dims = null)
```
tarekgh, line 5188: >IEnumerable [](http://example.com/codeflow?start=31&length=11)

would it make sense more if you make this method takes either ReadOnlySpan<long> which will avoid the array allocation in the callsite and also will avoid calling ToArray here too.

tarekgh (dismissed):
Added minor comments and a suggestion if you want to consider it. LGTM, otherwise.

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -5142,6 +5142,64 @@ public IList<Tensor> nonzero_as_list()
...
+
+            /// <summary>
+            /// Roll the tensor along the given dimension(s).
+            /// Elements that are shifted beyond the last position are re-introduced at the first position.
+            /// If a dimension is not specified, the tensor will be flattened before rolling and then restored to the original shape.
+            /// </summary>
+            public Tensor roll(long shifts, long? dims = null)
+            {
+                if (dims.HasValue) {
+                    return roll(new long[] { shifts }, new long[] { dims.Value });
```
tarekgh, line 5156: >new long[]  [](http://example.com/codeflow?start=32&length=11)

Thanks for making Roll take a span. now you can convert all these long array allocations to stackalloc's.

src/TorchSharp/Tensor/Tensor.cs
NiklasGustafsson: I wasn't able to make the 'dims' argument a span, since it didn't let me default to 'null' Is there a solution to that? An empty span?

src/TorchSharp/Tensor/Tensor.cs
NiklasGustafsson: Hmm... When I tried changing to ReadOnlySpan for a few other APIs (torch.zeros, for example), that seems to not work with F#.

@dsyme -- this doesn't work if torch.zeros takes a ReadOnlySpan:

```F#
let mutable pe = torch.zeros([| maxLen; dmodel|])
```

Do I need to overload ROS<long> and IEnumerable<long>?

src/TorchSharp/Tensor/Tensor.cs
tarekgh: You can just pass empty span when having null dims. You can pass `default` value for the dims parameter.

src/TorchSharp/Tensor/Tensor.cs
NiklasGustafsson: >> Do I need to overload ROS and IEnumerable?

It seems I need to have roll(ReadOnlySpan<long>,...) and roll(long[],...) as well as a private _roll(ReadOnlySpan<long>...) with the actual implementation.

NiklasGustafsson:
I've been looking for other places to use ReadOnlySpan<T>, but I will put that into separate PR.

PR #625, opened 2022-06-10
Add 'decimals' argument to Tensor.round()

NiklasGustafsson:


MovGP0:
C++ implementation of `std(decimals)` and unit tests are now merged.  

Question: are the `THS*.h` and `THS*.cpp` files the only ones that need editing, when implementing new methods like [std_mean(https://pytorch.org/docs/stable/generated/torch.std_mean.html)? 

If so, I might be able to do it myself :-)

NiklasGustafsson:
Generally, yes. It gets (slightly) more complicated if you need a new file and have to add it to the CMake files list.

PR #626, opened 2022-06-10
implementation of std_mean

MovGP0:
adding missing `std_mean` methods to the C++ files and in the C# file. 

Open issues: 

* check if build is working
* implement unit tests

### DIFF
--- a/src/Native/LibTorchSharp/THSTensor.h
+++ b/src/Native/LibTorchSharp/THSTensor.h
@@ -1070,9 +1070,9 @@ EXPORT_API(Tensor) THSTensor_std(const Tensor tensor);
 
 EXPORT_API(Tensor) THSTensor_std_along_dimensions(const Tensor tensor, const int64_t* dimensions, int length, bool unbiased, bool keepdim);
 
-EXPORT_API(Tensor) THSTensor_std_mean(const Tensor tensor);
+EXPORT_API(Tensor) THSTensor_std_mean(const Tensor tensor, bool unbiased, Tensor* mean);
 
-EXPORT_API(Tensor) THSTensor_std_mean_along_dimensions(const Tensor tensor, const int64_t* dimensions, int length, bool unbiased, bool keepdim);
+EXPORT_API(Tensor) THSTensor_std_mean_along_dimensions(const Tensor tensor, const int64_t* dimensions, int length, bool unbiased, bool keepdim, Tensor* mean);
 
 EXPORT_API(Tensor) THSTensor_sub(const Tensor left, const Tensor right);
 
--- a/src/Native/LibTorchSharp/THSTensor.h
+++ b/src/Native/LibTorchSharp/THSTensor.h
@@ -1070,6 +1070,10 @@ EXPORT_API(Tensor) THSTensor_std(const Tensor tensor);
 
 EXPORT_API(Tensor) THSTensor_std_along_dimensions(const Tensor tensor, const int64_t* dimensions, int length, bool unbiased, bool keepdim);
 
+EXPORT_API(Tensor) THSTensor_std_mean(const Tensor tensor);
+
+EXPORT_API(Tensor) THSTensor_std_mean_along_dimensions(const Tensor tensor, const int64_t* dimensions, int length, bool unbiased, bool keepdim);
+
 EXPORT_API(Tensor) THSTensor_sub(const Tensor left, const Tensor right);
 
 EXPORT_API(Tensor) THSTensor_sub_(const Tensor left, const Tensor right);
--- a/src/TorchSharp/Tensor/Tensor.cs
+++ b/src/TorchSharp/Tensor/Tensor.cs
@@ -4098,13 +4098,13 @@ namespace TorchSharp
             }
 
             [DllImport("LibTorchSharp")]
-            static extern IntPtr THSTensor_std(IntPtr tensor);
+            static extern IntPtr THSTensor_std(IntPtr tensor, bool unbiased);
 
             /// <summary>Calculates the standard deviation of all elements in the tensor.</summary>
             [System.Diagnostics.Contracts.Pure]
-            public Tensor std()
+            public Tensor std(bool unbiased = true)
             {
-                var res = THSTensor_std(Handle);
+                var res = THSTensor_std(Handle, unbiased);
                 if (res == IntPtr.Zero)
                     torch.CheckForErrors();
                 return new Tensor(res);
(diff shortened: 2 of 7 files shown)


src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -4122,6 +4136,74 @@ public Tensor std(long[] dimensions, bool unbiased = true, bool keepDimension =
...
+            /// If <paramref name="unbiased" /> is <value>true</value>, Bessel’s correction will be used.
+            /// Otherwise, the sample deviation is calculated, without any correction.
+            /// </remarks>
+            /// <param name="dimensions">The dimensions to reduce.</param>
+            /// <param name="unbiased">Whether to use Bessel’s correction (δN=1).</param>
+            /// <param name="keepDimension">Whether the <see cref="Tensor">output tensor</see> has dim retained or not.</param>
+            /// <param name="type"></param>
+            /// <returns>The <see cref="Tensor">output tensor</see>.</returns>
+            [System.Diagnostics.Contracts.Pure]
+            public Tensor std_mean(long[] dimensions, bool unbiased = true, bool keepDimension = false, ScalarType? type = null)
```
NiklasGustafsson, line 4182: I've started adding overloads with `ReadOnlySpan<long>` in addition to the `long[]` version. That helps make the other overloads marginally faster, since they can use 'stackalloc' instead of 'new' when creating the array to pass in for the dimensions argument.

The easiest thing to do is to duplicate the body with the unsafe code -- having the `long[]` version call the `ReadOnlySpan<long>` version risks infinite recursion (which I discovered recently).

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -4122,6 +4136,74 @@ public Tensor std(long[] dimensions, bool unbiased = true, bool keepDimension =
...
+            /// Otherwise, the sample deviation is calculated, without any correction.
+            /// </remarks>
+            /// <param name="dimension">The dimension to reduce.</param>
+            /// <param name="unbiased">Whether to use Bessel’s correction (δN=1).</param>
+            /// <param name="keepDimension">Whether the <see cref="Tensor">output tensor</see> has <paramref name="dimension" /> retained or not.</param>
+            /// <param name="type"></param>
+            /// <returns>The <see cref="Tensor">output tensor</see>.</returns>
+            [System.Diagnostics.Contracts.Pure]
+            public Tensor std_mean(long dimension, bool unbiased = true, bool keepDimension = false, ScalarType? type = null)
+                => std_mean(new[] { dimension }, unbiased, keepDimension, type);
```
NiklasGustafsson, line 4205: See comment above regarding `ReadOnlySpan<long>` and `stackalloc`

test/TorchSharpTest/TestTorchTensor.cs
```diff
@@ -4459,6 +4459,27 @@ public void RoundTest()
...
+        {
+            const long n = 7L;
+            var i = eye(n); // identity matrix
+            var a = rand(new[] { n, n });
+            var b = linalg.inv(a);
+
+            // check non-inline version
+            var r0 = round(matmul(a, b), 2L);
+            var r1 = round(matmul(b, a), 3L);
+            Assert.True(torch.Equals(i, r0), "round() failed");
```
NiklasGustafsson, line 4473: Any chance of rounding causing false assertions here? There's an allclose() comparator that I've been using in some unit tests when comparing tensor arithmetic to some expected outcome.

NiklasGustafsson (changes requested):

NiklasGustafsson:
Any unit tests for std_mean()? Did I miss those?

NiklasGustafsson:
Build failing, because it's apparently implemented as `torch::std_mean()` instead of a method on `at::Tensor`.

NiklasGustafsson:
@MovGP0 -- I just sent you a PR in your repo with some modifications. Turns out that torch::std_mean() returns a tuple, not a single Tensor.

MovGP0:
> @MovGP0 -- I just sent you a PR in your repo with some modifications. Turns out that torch::std_mean() returns a tuple, not a single Tensor.

looks great! PR is merged.

src/TorchSharp/Tensor/Tensor.cs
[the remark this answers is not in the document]
MovGP0: > Could the C# method be merged with the other one, just with a 'dimensions = null' default?

I would expect the following methods to look be the same for the compiler: 

```csharp
public Tensor std_mean(long[]? dimensions = null, bool unbiased = true, bool keepDimension = false, ScalarType? type = null);
public Tensor std_mean(int? dimension = 0, bool unbiased = true, bool keepDimension = false, ScalarType? type = null);
```

src/Native/LibTorchSharp/THSTensor.h
```diff
@@ -1069,6 +1069,10 @@ EXPORT_API(Tensor) THSTensor_std(const Tensor tensor);
 
 EXPORT_API(Tensor) THSTensor_std_along_dimensions(const Tensor tensor, const int64_t* dimensions, int length, bool unbiased, bool keepdim);
 
+EXPORT_API(Tensor) THSTensor_std_mean(const Tensor tensor);
+
+EXPORT_API(Tensor) THSTensor_std_mean_along_dimensions(const Tensor tensor, const int64_t* dimensions, int length, bool unbiased, bool keepdim);
```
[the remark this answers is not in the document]
NiklasGustafsson: It's been such a long time. They are placed under the `obj` folder, somewhere, when the environment is setup with the first build. Following the steps in DEVGUIDE.md should make it happen. If you delete `obj` and `bin` the environment should get refreshed from scratch, I believe.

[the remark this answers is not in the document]
NiklasGustafsson: But you do have to have room for libtorch on your disk.

src/TorchSharp/Tensor/Tensor.cs
MovGP0: seems like only the `ReadOnlySpan<long>` version is required, since an `Array<T>` can be converted to an `ReadOnlySpan<T>` implicitely. I've removed the array version.

NiklasGustafsson: > seems like only the `ReadOnlySpan<long>` version is required, since an `Array<T>` can be converted to an `ReadOnlySpan<T>` implicitely. I've removed the array version.

Unfortunately, not in F#. The array version is required in order for F# array literals to work.

PR #627, opened 2022-06-11
Add torchaudio.datasets.YESNO().

kaiidams:
- Implementation of `torchaudio.datasets.YESNO` that implements `torch.data.IDataset`
- `torchaudio.load, save, info` with no backends. Probably we could use libsndfile like PyTorch for Windows. https://github.com/aybe/libsndfile.NET is C# wrapper of libsndfile but it is licensed with GPL-3. I'm not sure if it is compatible with us.
- `torch.hub.download_url_to_file()` to download archive files.

Issue: @NiklasGustafsson 

- `IDataset.GetTensor()` produces dictionaries of `torch.Tensor`, however, `YESNO` wants to return additional information like the sampling rate of the audio. Maybe more flexibility like `IDataset<T> : IList<T>` is good.
- PyTorch doesn't implement Python async methods. Should we have it?

### DIFF
--- a/src/TorchSharp/Hub.cs
+++ b/src/TorchSharp/Hub.cs
@@ -24,7 +24,7 @@ namespace TorchSharp
                 using (var webClient = new WebClient()) {
                     if (progress) {
                         // TODO: Implement progress.
-                        Console.WriteLine("download_url_to_file(progress: true) is not supported.");
+                        // Console.WriteLine("download_url_to_file(progress: true) is not supported.");
                     }
                     webClient.DownloadFile(url, dst);
                     string fileHash = GetFileChecksum(dst);
--- a/src/TorchSharp/Hub.cs
+++ b/src/TorchSharp/Hub.cs
@@ -1,0 +1,44 @@
+// Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
+using System;
+using System.IO;
+using System.Net;
+using System.Security.Cryptography;
+using System.Text;
+
+namespace TorchSharp
+{
+    public static partial class torch
+    {
+        public static partial class hub
+        {
+            public static void download_url_to_file(string url, string dst, string hash_prefix = null, bool progress = true)
+            {
+                using (var webClient = new WebClient()) {
+                    if (progress) {
+                        // TODO: Implement progress.
+                        Console.WriteLine("download_url_to_file(progress: true) is not supported.");
+                    }
+                    webClient.DownloadFile(url, dst);
+                    string fileHash = GetFileChecksum(dst);
+                    if (hash_prefix != null && !fileHash.StartsWith(hash_prefix)) {
+                        throw new InvalidDataException($"invalid hash value (expected \"{hash_prefix}\", got \"{fileHash}\"");
+                    }
+                }
+            }
+
+            private static string GetFileChecksum(string path)
+            {
+                using (SHA256 sha256 = SHA256.Create()) {
+                    using (var stream = File.OpenRead(path)) {
+                        var hashValue = sha256.ComputeHash(stream);
+                        var sb = new StringBuilder();
+                        foreach (var value in hashValue) {
+                            sb.Append($"{value:x2}");
+                        }
+                        return sb.ToString();
+                    }
+                }
+            }
+        }
+    }
+}
--- a/src/TorchSharp/Hub.cs
+++ b/src/TorchSharp/Hub.cs
@@ -11,6 +11,14 @@ namespace TorchSharp
     {
         public static partial class hub
         {
+            /// <summary>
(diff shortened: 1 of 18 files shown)


NiklasGustafsson:
Hi @kaiidams,

This looks like a really great PR -- could you please add some doc comments before anything else? This is such a new area for me, that I would like to know a more about the APIs before looking into them.

On the 'async' question -- yes, anywhere we're doing I/O underneath, it would be a good idea to offer async versions, too.

GeorgeS2019:
@kaiidams 

As we discussed previously, could you consider .NET Interactive as notebooks to demonstrate the functionalities of torchaudio of TorchSharp.

Many of the simple use cases of torchaudio from pytorch use jupyter notebooks.

If you could deliver those simple pytorch jupyter examples using .NET interactive, this will be VERY exciting for NOT just the Torchsharp and ML.NET But also .NET interactive communities.

basic torchaudio use cases

- [AUDIO I/O](https://pytorch.org/audio/stable/tutorials/audio_io_tutorial.html)
- [AUDIO RESAMPLING](https://pytorch.org/audio/stable/tutorials/audio_resampling_tutorial.html)

These use cases in .NET will also help machine learning users to learn Deep Audio through TorchSharp. 

I understand these use cases are probably too simpler for the .NET/python deep audio technical challenges you typically attempt to solve.  However, these simple examples will encourage more .NET users to try Torchsharp and .NET interactive. 


Great job. Excellent as always.

NiklasGustafsson:
@kaiidams:

>> Maybe more flexibility like IDataset<T> : IList<T> is good.

This sounds reasonable, at first glance. Can you elaborate on the experience -- what would "T" be?

kaiidams:
> @kaiidams:
> 
> > > Maybe more flexibility like IDataset : IList is good.
> 
> This sounds reasonable, at first glance. Can you elaborate on the experience -- what would "T" be?

@NiklasGustafsson Thank you for the advice. `DataLoaderEnumerator` assumes that `T` needs

a. to be collated into a batch
b. to be disposed
c. to be created in worker thread safely

Added generalized version of `DataLoader`. Like PyTorch **a.** can be done by user function. Audio and text datasets need to batch variable length tensors. For **b.** `DataLoaderEnumerator.Current` is detached from the scope as `DataLoaderEnumerator.Dispose()` calls `Dispose()` of it.

kaiidams:
@GeorgeS2019 Thanks. I think it will take some time to make this possible.

> If you could deliver those simple pytorch jupyter examples using .NET interactive, this will be VERY exciting for NOT just the Torchsharp and ML.NET But also .NET interactive communities.

src/TorchSharp/Hub.cs
```diff
@@ -0,0 +1,52 @@
...
+            /// <param name="dst">The file path to download the URL into</param>
+            /// <param name="hash_prefix">If non null, the SHA256 hash of downloaded content must match this prefix</param>
+            /// <param name="progress">Display the progress bar</param>
+            /// <exception cref="InvalidDataException">SHA256 hash doesn't match</exception>
+            public static void download_url_to_file(string url, string dst, string hash_prefix = null, bool progress = true)
+            {
+                using (var webClient = new WebClient()) {
+                    if (progress) {
+                        // TODO: Implement progress.
+                        Console.WriteLine("download_url_to_file(progress: true) is not supported.");
```
tarekgh, line 27: >Console [](http://example.com/codeflow?start=24&length=7)

maybe better commenting this line for now to avoid Console dependency.

src/TorchSharp/Hub.cs
```diff
@@ -0,0 +1,52 @@
...
+            /// Download the url to a file 
+            /// </summary>
+            /// <param name="url">The URL to download</param>
+            /// <param name="dst">The file path to download the URL into</param>
+            /// <param name="hash_prefix">If non null, the SHA256 hash of downloaded content must match this prefix</param>
+            /// <param name="progress">Display the progress bar</param>
+            /// <exception cref="InvalidDataException">SHA256 hash doesn't match</exception>
+            public static void download_url_to_file(string url, string dst, string hash_prefix = null, bool progress = true)
+            {
+                using (var webClient = new WebClient()) {
```
tarekgh, line 24: >WebClient [](http://example.com/codeflow?start=43&length=9)

nit: WebClient is not recommended to use in any new development. Is it possible to use HttpClient here instead? 

[https://docs.microsoft.com/en-us/dotnet/api/system.net.webclient?view=net-6.0#remarks](https://docs.microsoft.com/en-us/dotnet/api/system.net.webclient?view=net-6.0#remarks)

tarekgh:
I put a few little comments. LGTM, otherwise. 

@michaelgsharp do you want to take a look.

NiklasGustafsson (approved):

PR #630, opened 2022-06-17
Add xxxstr() methods

dayo05:
This don't make change about t.str() but it will be useful for who wants to get Metadata string and Numpy style string.
I think `t.npstr()` is better then `t.str(TensorStringStyle.Numpy)`.


NiklasGustafsson:
I'll merge this, and then we'll consider separately whether to move the 'style' parameter to the end.

PR #641, opened 2022-06-26
Added torch.nn.utils.rnn

kaiidams:
- torch.nn.utils.rnn.pack_padded_sequence
- torch.nn.utils.rnn.pad_packed_sequence
- torch.nn.utils.rnn.pad_sequence
- torch.nn.utils.rnn.pack_sequence

### DIFF
--- a/src/TorchSharp/NN/Utils/PackedSequence.cs
+++ b/src/TorchSharp/NN/Utils/PackedSequence.cs
@@ -1,0 +1,97 @@
+// Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
+using System;
+using System.Collections.Generic;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace TorchSharp
+{
+    public static partial class torch
+    {
+        public static partial class nn
+        {
+            public static partial class utils
+            {
+                public static partial class rnn
+                {
+                    /// <summary>
+                    /// A packed batch of variable length sequences.
+                    /// </summary>
+                    public class PackedSequence : IDisposable
+                    {
+                        /// <summary>
+                        /// Class wrapping PyTorch's packedsequence object reference.
+                        /// </summary>
+                        internal sealed class HType : SafeHandle
+                        {
+                            public HType(IntPtr preexistingHandle, bool ownsHandle)
+                                : base(IntPtr.Zero, ownsHandle)
+                            {
+                                SetHandle(preexistingHandle);
+                            }
+
+                            public override bool IsInvalid => handle == IntPtr.Zero;
+
+                            // This is just for marshalling
+                            internal HType() : base(IntPtr.Zero, true)
+                            {
+                            }
+
+                            [DllImport("LibTorchSharp")]
+                            private static extern void THSNN_PackedSequence_dispose(HType handle);
+
+                            protected override bool ReleaseHandle()
+                            {
+                                THSNN_PackedSequence_dispose(this);
+                                return true;
+                            }
+
+                            protected override void Dispose(bool disposing)
+                            {
+                                if (disposing) {
+                                    ReleaseHandle();
+                                }
+                            }
+                        }
+
+                        private HType handle;
+
+                        internal PackedSequence(IntPtr handle)
+                        {
+                            if (handle != IntPtr.Zero) {
+                                this.handle = new HType(handle, true);
+                            }
+                        }
+
+                        ~PackedSequence()
+                        {
+                            Dispose(false);
+                        }
+
+                        internal HType Handle => handle;
+
+                        /// <summary>
+                        ///   Releases the storage.
+                        /// </summary>
+                        public void Dispose()
+                        {
+                            Dispose(true);
+                            GC.SuppressFinalize(this);
+                        }
+
+                        /// <summary>
+                        ///   Implements the .NET Dispose pattern.
+                        /// </summary>
+                        protected virtual void Dispose(bool disposing)
+                        {
+                            if (disposing && handle != null && !handle.IsInvalid) {
+                                handle.Dispose();
+                                handle.SetHandleAsInvalid();
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
--- a/src/TorchSharp/NN/Utils/PackedSequence.cs
+++ b/src/TorchSharp/NN/Utils/PackedSequence.cs
@@ -38,19 +38,12 @@ namespace TorchSharp
                             }
 
                             [DllImport("LibTorchSharp")]
-                            private static extern void THSNN_PackedSequence_dispose(HType handle);
+                            private static extern void THSNN_PackedSequence_dispose(IntPtr handle);
 
                             protected override bool ReleaseHandle()
                             {
-                                THSNN_PackedSequence_dispose(this);
+                                THSNN_PackedSequence_dispose(handle);
                                 return true;
-                            }
-
-                            protected override void Dispose(bool disposing)
-                            {
-                                if (disposing) {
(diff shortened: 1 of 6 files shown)


src/TorchSharp/NN/Utils/PackedSequence.cs
```diff
@@ -0,0 +1,97 @@
...
+                            internal HType() : base(IntPtr.Zero, true)
+                            {
+                            }
+
+                            [DllImport("LibTorchSharp")]
+                            private static extern void THSNN_PackedSequence_dispose(HType handle);
+
+                            protected override bool ReleaseHandle()
+                            {
+                                THSNN_PackedSequence_dispose(this);
```
tarekgh, line 45: >THSNN_PackedSequence_dispose [](http://example.com/codeflow?start=32&length=28)

should you assign the handle back to IntPtr.Zero? 

Also in the native side THSNN_PackedSequence_dispose would be nice to check for null/zero value before deleting it?

src/TorchSharp/NN/Utils/PackedSequence.cs
```diff
@@ -0,0 +1,97 @@
...
+        public static partial class nn
+        {
+            public static partial class utils
+            {
+                public static partial class rnn
+                {
+                    /// <summary>
+                    /// A packed batch of variable length sequences.
+                    /// </summary>
+                    public class PackedSequence : IDisposable
```
tarekgh, line 20: >PackedSequence [](http://example.com/codeflow?start=33&length=14)

can we seal this class? or it expected to have someone subclass it in some scenarios?

tarekgh:
I added a couple of comments/question, LGTM otherwise.

src/TorchSharp/NN/Utils/PackedSequence.cs
kaiidams: There is a example how to use `SafeHandle` https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.safehandle?view=net-6.0 . And also it is aligned with `Optimizer` class. Let me double check how it behave.

src/TorchSharp/NN/Utils/PackedSequence.cs
kaiidams: Thanks. Added seal.

src/TorchSharp/NN/Utils/PackedSequence.cs
```diff
@@ -0,0 +1,97 @@
...
+                            Dispose(true);
+                            GC.SuppressFinalize(this);
+                        }
+
+                        /// <summary>
+                        ///   Implements the .NET Dispose pattern.
+                        /// </summary>
+                        protected virtual void Dispose(bool disposing)
+                        {
+                            if (disposing && handle != null && !handle.IsInvalid) {
```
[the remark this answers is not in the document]
kaiidams: https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.safehandle?view=net-6.0 seems not to check `disposing`. Let me verify this.

NiklasGustafsson:
The compiler is not liking your sealing a class with a virtual member, I guess.

src/TorchSharp/NN/Utils/PackedSequence.cs
tarekgh: What I was trying to say is, in `internal sealed class HType : SafeHandle` you have been explicitly overriding `protected override void Dispose(bool disposing)`. This means when dispose is called the internal state of the safe handle will not be noticed as it is disposed. Look at the https://source.dot.net/#System.Private.CoreLib/SafeHandle.cs,98 to get the idea what the default implementation of the dispose does in the SafeHandle. You may not need to override `Dispose(bool disposing)` at all and let `SafeHandle` automaltically handle the state.  

Note, the sample you have pointed at https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.safehandle?view=net-6.0 is not overriding `Dispose(bool disposing)`.

kaiidams: Thank you @tarekgh. Removed `Dispose(bool)` to let `SafeHandle` handle the state. Also `THSNN_PackedSequence_dispose()` receives `IntPtr` as `NativeMethods.CloseHandle` of https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.safehandle?view=net-6.0 does so.

src/TorchSharp/NN/Utils/PackedSequence.cs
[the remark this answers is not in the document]
kaiidams: @tarekgh Removed `Dispose(bool)`. https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.safehandle?view=net-6.0 says `handle != null` can happen.

Should this be
```
                    [DllImport("LibTorchSharp")]
                    private static extern IntPtr THSNN_pack_sequence(IntPtr[] sequences, int sequences_len, bool enforce_sorted);
```
like
```
                    [DllImport("LibTorchSharp")]
                    private static extern PackedSequence.HType THSNN_pack_sequence(IntPtr[] sequences, int sequences_len, bool enforce_sorted);
```
to avoid leaking?

src/TorchSharp/NN/Utils/RNNUtils.cs
```diff
@@ -0,0 +1,90 @@
...
+                    /// </summary>
+                    /// <param name="sequences">A list of variable length sequences</param>
+                    /// <param name="batch_first">If true, the size of output tensor is (batch, seq, *), otherwise (seq, batch, *)</param>
+                    /// <param name="padding_value">The values of padding.</param>
+                    /// <returns>The padded tensor</returns>
+                    public static torch.Tensor pad_sequence(IEnumerable<torch.Tensor> sequences, bool batch_first = false, double padding_value = 0.0)
+                    {
+                        var sequences_arg = sequences.Select(p => p.Handle).ToArray();
+                        var res = THSNN_pad_sequence(sequences_arg, sequences_arg.Length, batch_first, padding_value);
+                        return new torch.Tensor(res);
```
tarekgh, line 71: >res [](http://example.com/codeflow?start=48&length=3)

who is freeing these resource?

src/TorchSharp/NN/Utils/PackedSequence.cs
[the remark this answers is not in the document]
tarekgh: I asked the same question below before I saw your comment :-)

I am seeing the returned value is used in `new torch.Tensor(res);`. Tensor will call [THSTensor_dispose](https://github.com/dotnet/TorchSharp/blob/49910eceb6a6ed3b99103a5c1782f6b5b802021a/src/TorchSharp/Tensor/Tensor.cs#L87) to dispose the handle. Is this right for disposing the `PackedSequence`?

src/TorchSharp/NN/Utils/RNNUtils.cs
kaiidams: I think `res` is owned by `Tensor` and it is freed in `Tensor.Dispose(bool)`. The caller or `DisposeScope` are responsible to free it, otherwise `~Tensor()` will do.

src/TorchSharp/NN/Utils/PackedSequence.cs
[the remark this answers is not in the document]
kaiidams: `Tensor` and `PackedSequence` are working similarly but `Tensor` is not using `SafeHandle`. I assume that `PacketSequence` owns the handle and disposes it. `pack_sequence()` returns `PackedSequence`, `pad_sequence()` returns `Tensor`.

tarekgh (approved):

NiklasGustafsson (approved):

PR #644, opened 2022-07-02
Load TorchScript modules

NiklasGustafsson:
With this PR, TorchSharp supports loading modules that were created using torch.jit.{trace,script} with Pytorch. 

No support for traced or scripted functions, yet. The PR also does not support tracing or scripting in TorchSharp.


src/TorchSharp/JIT/ScriptModule.cs
```diff
@@ -0,0 +1,333 @@
...
+
+                ~ScriptModule()
+                {
+                    Dispose(false);
+                }
+
+                /// <summary>
+                /// Implements the .NET Dispose pattern.
+                /// </summary>
+                protected override void Dispose(bool disposing)
```
tarekgh, line 30: >Dispose [](http://example.com/codeflow?start=40&length=7)

why you are overriding this Dispose method? wouldn't the base class Dispose will just work? Also wouldn't overriding it will leak the named_* stuff which you are commenting here?

NiklasGustafsson: Thanks for pointing that out. It was a temporary thing, while the named_* stuff was yet to be implemented.

src/TorchSharp/JIT/ScriptModule.cs
```diff
@@ -0,0 +1,333 @@
...
+                public override nn.Module to(DeviceType deviceType, int deviceIndex = -1)
+                {
+                    if (deviceType != DeviceType.CUDA) deviceIndex = -1;
+
+                    if (deviceType == DeviceType.CUDA && !torch.cuda.is_available()) throw new InvalidOperationException("CUDA is not available.");
+
+                    if (deviceType != _deviceType || deviceIndex != _deviceIndex) {
+
+                        InitializeDeviceType(deviceType);
+                        THSJIT_Module_to_device(handle, (int)deviceType, deviceIndex);
```
tarekgh, line 180: >deviceIndex [](http://example.com/codeflow?start=73&length=11)

for my education, what is the behavior when calling THSJIT_Module_to_device and passing the deviceIndex = -1?

src/TorchSharp/JIT/ScriptModule.cs
```diff
@@ -0,0 +1,309 @@
...
+                                ConditionallyRegisterBuffer(fieldName, t);
+                                break;
+                            }
+                        }
+                    }
+
+                    return this;
+                }
+
+#if false   // These functions "work," but the native code doesn't seem to find any interesting information.
```
tarekgh, line 245: >if [](http://example.com/codeflow?start=1&length=2)

we'll keep this code for now? what is the next step here?

src/TorchSharp/JIT/ScriptModule.cs
```diff
@@ -0,0 +1,309 @@
...
+            }
+
+            [DllImport("LibTorchSharp")]
+            private static extern IntPtr THSJIT_load(string filename);
+
+            public static ScriptModule load(string filename)
+            {
+                if (!System.IO.File.Exists(filename))
+                    throw new System.IO.FileNotFoundException(filename);
+                return new ScriptModule(THSJIT_load(filename));
```
tarekgh, line 303: >THSJIT_load(filename) [](http://example.com/codeflow?start=40&length=21)

What happen if I pass a file pass that is exist but not containing a valid script? does the native side somehow throw? how we recognize this failing case?

src/TorchSharp/JIT/Type/Type.cs
```diff
@@ -1,108 +1,126 @@
...
+                            THSJIT_TensorType_dispose(this);
+                            break;
+                        default:
+                            THSJIT_Type_dispose(this);
+                            break;
+                        }
+                        return true;
+                    }
+
+                    protected override void Dispose(bool disposing)
```
tarekgh, line 52: >Dispose [](http://example.com/codeflow?start=44&length=7)

I think you don't need to override this method. The SafeHandle.Dispose(bool) handle the internal disposing state of the object and calls ReleaseHandle as needed. 

[https://source.dot.net/#System.Private.CoreLib/SafeHandle.cs,98](https://source.dot.net/#System.Private.CoreLib/SafeHandle.cs,98)

tarekgh (dismissed):
Added a few comments. LGTM otherwise.

src/TorchSharp/JIT/ScriptModule.cs
NiklasGustafsson: It picks the default GPU.

src/TorchSharp/JIT/ScriptModule.cs
NiklasGustafsson: I want to try to figure out why the scripts don't seem to contain any information on element type or dimensions.

src/TorchSharp/JIT/ScriptModule.cs
NiklasGustafsson: It will blow up in some fashion. We're just interfacing with the libtorch support for TorchScript, so whatever it does. I'll add a negative unit test.

src/TorchSharp/JIT/ScriptModule.cs
```diff
@@ -0,0 +1,338 @@
...
+                    }
+                }
+#endif
+
+                [DllImport("LibTorchSharp")]
+                private static extern IntPtr THSJIT_Module_forward(HType module, IntPtr tensors, int length);
+
+                public unsafe override Tensor forward(Tensor tensor)
+                {
+                    using (var parray = new PinnedArray<IntPtr>()) {
```
NiklasGustafsson, line 292: @tarekgh -- is there some way to avoid the new[]{} and use stackalloc here, instead?

GeorgeS2019:
I assume this is related to this [functionality: torch.jit.save](https://pytorch.org/docs/stable/generated/torch.jit.save.html)

For consistency, would it make sense to rename the extension *.dat in the unit tests to *.pt?

test/TorchSharpTest/func.script.pt
test/TorchSharpTest/l1000_100_10.script.pt
test/TorchSharpTest/linrelu.script.pt
test/TorchSharpTest/scripted.script.pt

NiklasGustafsson:
The file extensions used in unit tests don't really matter.

NiklasGustafsson:
But, yes, there will now be three file formats:

1. *.pth, which TorchSharp does not support. This is for saving Pytorch modules via pickling.
2. *.pt, which TorchSharp will load and save, but not create from scratch. This is for saving Pytorch modules via TorchScript.
3. *.ts/.data/.whatever, which is for saving model weights only in the TorchSharp-specific format.

GeorgeS2019:
I think *.ts is a good TorchSharp format for those involving importsd.py and exportsd.py

GeorgeS2019:
> *.pt, which TorchSharp will load and save, but not create from scratch. This is for saving Pytorch modules via TorchScript.

Using the consensus practice used in Rust LibTorch wrapper, *.pt is meant for Torchscript,   created in pytorch.jit.save,  loaded in RustWrapper, save in RustWrapper.  I believe they are all could be interchanged, I hope this is right. 

If so,  ***.pt** could be a good format interchange between pytorch, Torchsharp and perhaps other LibTorch wrappers

GeorgeS2019:
Another thought!

**.pt** is used between pytorch and LibTorch c++ wrappers, including Torchsharp.

Since .pt includes the architecture of networks and trained weights, .pt could be converted into Onnx using TorchSharp for interchange with ML.NET.

Likewise, that interchange could be done in ML.NET.

However, I think it makes more sense to have a utility to convert **.pt** to onnx in Torchsharp.

NiklasGustafsson:
> If so, ***.pt** could be a good format interchange between pytorch, Torchsharp and perhaps other LibTorch wrappers

That's what it's for. Please note that TorchSharp will lack the ability to __create__ TorchScript files, even after this PR. We will just be able to load and save.

GeorgeS2019:
Another thought!

**.pt** is used between pytorch and LibTorch c++ wrappers, including Torchsharp.

Since .pt includes the architecture of networks and trained weights, .pt could be converted into Onnx using TorchSharp for interchange with ML.NET.

Likewise, that interchange could be done in ML.NET.

However, I think it makes more sense to have a utility to convert **.pt** to onnx in Torchsharp.

GeorgeS2019:
>  TorchSharp will lack the ability to create TorchScript files,

Agree,  however, Torchsharp is able to load TorchSript files and save back TorchScript files though the libtorch c++. => that is loadable by pytorch or other libtorch wrappers

NiklasGustafsson:
> Since .pt includes the architecture of networks and trained weights, .pt could be converted into Onnx

I believe that's exactly how the ONNX exporter for Pytorch works. Since you have to start with Pytorch (not TorchSharp) in order to get a TorchScript file, I don't know why you would want to export to ONNX from TorchSharp. What is the scenario you have in mind?

GeorgeS2019:
> I don't know why you would want to export to ONNX from TorchSharp. What is the scenario you have in mind?

The exported ONNX help keep the .NET communities staying with .NET :-)

Imagine!

Many pytorch communities are sharing models using TorchScript formats.

TorchSharp could serve as the starting point for importing these TorchShript format s and allowing the .NET community to do further training or inference within Torchsharp OR exporting that to ONNX for inference using ML.NET

GeorgeS2019:
By doing that, this addresses the need for Deep Learning support in ML.NET based on a survey done in April 2021.

NiklasGustafsson:
> TorchSharp could serve as the starting point for importing these TorchShript format and allowing the .NET community to do further training or inference within Torchsharp OR exporting that to ONNX for inference using ML.NET

Not until you can start from scratch in TorchSharp. Right now, you cannot have a TorchScript file without starting in Python, where you can also create a ONNX file from the same model. Once we can do torch.jit.script() in TorchSharp, maybe, but until that time...

GeorgeS2019:
> Many pytorch communities are sharing models using TorchScript formats.

We do not have to start from scratch in TorchSharp :-)

GeorgeS2019:
Likewise, many .NET users do not care where Onnx come from as long as they are available for use in ML.NET

NiklasGustafsson:
In order to write an ONNX exporter, we would need to go through the same work that we need to do in order to export to TorchScript. So, until we're in a position to do that, ONNX export will also have to wait. I agree that it's important, but it's separate from this functionality, which allows you to import TorchScript, modify it, and save it, but nothing else.

GeorgeS2019:
Agree ONE THING AT A TIME. 

The final goal of keeping machine learning within the .NET is happening!

src/TorchSharp/JIT/ScriptModule.cs
tarekgh: You don't need to use PinnedArray at all too. You may try the following which will allow writing the whole method without any unsafe code and you'll not to do any managed allocations.

```C#
ReadOnlySpan<IntPtr> buffer = stackalloc IntPtr[1] { tensor.Handle };
var res = THSJIT_Module_forward(handle, (IntPtr)tensorRefs, 1);
```

PR #647, opened 2022-07-06
Web client has been deprecated and should to replaced with HttpClient

### LINKED ISSUE #637: Web client has been deprecated and should to replaced with HttpClient
In Hub.cs, This uses WebClient to download the file. 

This class has marked as obsolete on .NET6 and this class will removed soon.

kaiidams:
Fix #637

### DIFF
--- a/src/TorchSharp/Hub.cs
+++ b/src/TorchSharp/Hub.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Net.Http;
 using System.Security.Cryptography;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace TorchSharp
@@ -18,13 +19,12 @@ namespace TorchSharp
             /// <param name="url">The URL to download</param>
             /// <param name="dst">The file path to download the URL into</param>
             /// <param name="hash_prefix">If non null, the SHA256 hash of downloaded content must match this prefix</param>
-            /// <param name="progress">Display the progress bar</param>
             /// <exception cref="InvalidDataException">SHA256 hash doesn't match</exception>
-            public static void download_url_to_file(string url, string dst, string hash_prefix = null, bool progress = true)
+            public static void download_url_to_file(string url, string dst, string hash_prefix = null)
             {
                 try {
                     Task.Run(async () => {
-                        await download_url_to_file_async(url, dst, hash_prefix, progress);
+                        await download_url_to_file_async(url, dst, hash_prefix);
                     }).Wait();
                 } catch (AggregateException ex) {
                     throw ex.InnerException;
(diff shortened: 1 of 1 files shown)


src/TorchSharp/Hub.cs
```diff
@@ -21,17 +22,46 @@ public static partial class hub
...
+
+            /// <summary>
+            /// Download the url to a file 
+            /// </summary>
+            /// <param name="url">The URL to download</param>
+            /// <param name="dst">The file path to download the URL into</param>
+            /// <param name="hash_prefix">If non null, the SHA256 hash of downloaded content must match this prefix</param>
+            /// <param name="progress">Display the progress bar</param>
+            /// <exception cref="InvalidDataException">SHA256 hash doesn't match</exception>
+            public static async Task download_url_to_file_async(string url, string dst, string hash_prefix = null, bool progress = true)
```
NiklasGustafsson, line 42: Should there be a cancellation token argument?

NiklasGustafsson (changes requested):

src/TorchSharp/Hub.cs
```diff
@@ -21,17 +22,46 @@ public static partial class hub
...
+            {
+                using (var httpClient = new HttpClient()) {
+                    using (var response = await httpClient.GetAsync(url)) {
+                        if (!response.IsSuccessStatusCode) {
+                            throw new InvalidDataException();
+                        }
+                        using (var writer = File.OpenWrite(dst)) {
+                            if (progress) {
+                                // TODO: Implement progress.
```
NiklasGustafsson, line 51: If it's not yet supported, then just remove the 'progress' argument and add an issue to track adding it back in. It'll be confusing if you can pass the flag, but nothing different happens.

NiklasGustafsson:
@kaiidams -- I approved and merged your other PR. We're very close on this one, too. If you have a few moments to take a look at my comments, that would be wonderful. Once we have this merged, I'll make another release.

src/TorchSharp/Hub.cs
```diff
@@ -21,17 +22,46 @@ public static partial class hub
...
-                        // TODO: Implement progress.
-                        // Console.WriteLine("download_url_to_file(progress: true) is not supported.");
-                    }
-                    webClient.DownloadFile(url, dst);
-                    string fileHash = GetFileChecksum(dst);
-                    if (hash_prefix != null && !fileHash.StartsWith(hash_prefix)) {
-                        throw new InvalidDataException($"invalid hash value (expected \"{hash_prefix}\", got \"{fileHash}\"");
+                try {
+                    Task.Run(async () => {
+                        await download_url_to_file_async(url, dst, hash_prefix, progress);
```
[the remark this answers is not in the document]
kaiidams: @NiklasGustafsson 

I'm not very familiar to this, but I think this is needed to avoid deadlock in case the method is called from ASP.NET or desktop apps.

https://docs.microsoft.com/en-us/archive/msdn-magazine/2015/july/async-programming-brownfield-async-development#the-thread-pool-hack

Probably ASP.NET Core doesn't have this issue.

https://blog.stephencleary.com/2017/03/aspnetcore-synchronization-context.html#:~:text=ASP.NET%20Core%20does%20not%20have%20a%20SynchronizationContext%2C%20so,thread%2C%20and%20they%20may%20all%20run%20in%20parallel.

src/TorchSharp/Hub.cs
kaiidams: Added `cancellationToken` to the arguments of `torch.hub.download_url_to_file_async()`.

src/TorchSharp/Hub.cs
kaiidams: Removed `progress` from `torch.hub.download_url_to_file_async()`.

NiklasGustafsson (approved):

PR #656, opened 2022-07-15
adding var and var_mean functions

MovGP0:
- adding the [var](https://pytorch.org/docs/stable/generated/torch.var.html), and [var_mean](https://pytorch.org/docs/stable/generated/torch.var_mean.html) functions
- fixing some documentation details on `std_mean` functions


NiklasGustafsson:
@MovGP0 -- the reason the tests are failing seems to be that two of the native interop extern "C" functions you added have different signatures in the declarations and definitions.

MovGP0:
> different signatures in the declarations and definitions.
I will take a look at it, but unsure if I find the time this week.

PR #667, opened 2022-07-29
implementing isnan

MovGP0:
- implementing `tensor.isnan()`
- `using System.Diagnostics.Contracts`, because `[PureAttribute]` is often used

### DIFF
--- a/test/TorchSharpTest/TestTorchTensor.cs
+++ b/test/TorchSharpTest/TestTorchTensor.cs
@@ -2192,12 +2192,16 @@ namespace TorchSharp
                 Assert.True(tensor.isinf().data<bool>().ToArray().All(b => b));
                 Assert.False(tensor.isneginf().data<bool>().ToArray().All(b => b));
                 Assert.False(tensor.isfinite().data<bool>().ToArray().All(b => b));
+                Assert.False(tensor.isnan().data<bool>().ToArray().All(b => b));
+                Assert.False(torch.isnan(tensor).data<bool>().ToArray().All(b => b));
 
                 tensor.fill_(Single.NegativeInfinity);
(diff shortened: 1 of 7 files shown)


test/TorchSharpTest/TestTorchTensor.cs
```diff
@@ -2192,12 +2192,16 @@ public void InfinityTest()
...
                 Assert.False(tensor.isfinite().data<bool>().ToArray().All(b => b));
+                Assert.False(tensor.isnan().data<bool>().ToArray().All(b => b));
+                Assert.False(torch.isnan(tensor).data<bool>().ToArray().All(b => b));
 
                 tensor.fill_(Single.NegativeInfinity);
                 Assert.True(tensor.isneginf().data<bool>().ToArray().All(b => b));
                 Assert.True(tensor.isinf().data<bool>().ToArray().All(b => b));
                 Assert.False(tensor.isposinf().data<bool>().ToArray().All(b => b));
                 Assert.False(tensor.isfinite().data<bool>().ToArray().All(b => b));
+                Assert.False(tensor.isnan().data<bool>().ToArray().All(b => b));
```
NiklasGustafsson, line 2203: Should we have a unit test where isnan is true?

MovGP0: > Should we have a unit test where isnan is true?

I've implemented a better test of `isnan`.

MovGP0:
addressed some nullability issues in `Tensor.Factories.cs` and `TensorExtensionMethods.cs`. 

Tests are running on my machine. Failing build does not seem to be related to this commit.

NiklasGustafsson (approved):

Issue #670, opened 2022-07-31
Better align methods for creating tensors from .NET arrays with Pytorch APIs.


TheBlackPlague:
Hello,

It seems to me that this library's tensor allocation performance is not good at all. See the following code:

```cs
    [MethodImpl(MethodImplOptions.AggressiveOptimization)]
    public static (Tensor, Tensor) Batch(this Board[] boards, float[,] sideToMove, float[,] notSideToMove, int batchSize)
    {
        Stopwatch stopwatch = Stopwatch.StartNew();
        for (int i = 0; i < batchSize; i++) {
            FillArrayFromBoard(boards[i], sideToMove, notSideToMove, i);
        }
        stopwatch.Stop();
        LastElapsedMs = (int)stopwatch.ElapsedMilliseconds;

        return (from_array(sideToMove, dtype:float32), from_array(notSideToMove, dtype:float32));
    }
```

When using arrays of size `[100000, 768]` for both `sideToMove` and `notSideToMove`, `LastElapsedMs = 201ms` on a run. However, the whole method takes `6570ms`. That means the tensor allocation alone is taking `> 6000ms` even accounting for the time `Stopwatch.StartNew()` and `Stopwatch.Stop()` take.

To test it, you can just do `from_array()` 2x on a `[100000, 768]` array. Is there a faster way to do this? Because otherwise, it slows down performance quite a lot.

I'm using `TorchSharp-cuda-windows v0.97.0`.

TheBlackPlague:
I would like to add that I've confirmed the Tensors are being allocated on the CPU & not the GPU.

TheBlackPlague:
It seems that normal tensor allocation isn't slow.

```cs
Stopwatch stopwatch = Stopwatch.StartNew();
Tensor zero = zeros(100000, 768);
stopwatch.Stop();
zero.Dispose();

Console.WriteLine("Allocation took: " + stopwatch.ElapsedMilliseconds + "ms");
```
The output of this code is:
```
Allocation took: 103ms
```

However, the `from_array()` method is insanely slow.

NiklasGustafsson:
@TheBlackPlague -- from_array is a very general tensor factory, accepting any number of dimensions. That generality probably has some overhead. Can you try 'tensor(float[,]) instead and see what that does to performance? 'tensor()' is the canonical way to create a tensor from an array, but it only has overloads up to **four** dimensions. If you need more, we can add...

NiklasGustafsson:
[...]

PR #671, opened 2022-08-01
Fix performance issue reported in #670

NiklasGustafsson:
Fix the performance issue associated with tensor().

### DIFF
--- a/src/TorchSharp/Tensor/Tensor.Factories.cs
+++ b/src/TorchSharp/Tensor/Tensor.Factories.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using static TorchSharp.TensorExtensionMethods;
 
@@ -1369,40 +1370,46 @@ namespace TorchSharp
         /// <summary>
         /// Create a tensor from an array of values, shaping it based on the shape passed in.
         /// </summary>
+        [System.Diagnostics.Contracts.Pure]
         public static Tensor tensor(bool[] rawArray, long[]? dimensions = null, torch.ScalarType? dtype = null, torch.Device? device = null, bool requiresGrad = false)
         {
-            if (dimensions is null)
-                dimensions = new long[] { rawArray.LongLength };
-
-            var dataHandle = GCHandle.Alloc(rawArray, GCHandleType.Pinned);
-            return _tensor_generic(dataHandle, dimensions, (sbyte)ScalarType.Bool, dtype, device, requiresGrad);
+            return (dimensions is null)
+                ? _tensor_generic(rawArray, stackalloc long[] { rawArray.LongLength }, (sbyte)ScalarType.Bool, dtype, device, requiresGrad)
+                : _tensor_generic(rawArray, dimensions, (sbyte)ScalarType.Bool, dtype, device, requiresGrad); 
         }
 
-        private static Tensor _tensor_generic(GCHandle dataHandle, long[] dimensions, sbyte origType, ScalarType? dtype, Device? device, bool requiresGrad)
+        private static Tensor _tensor_generic(Array rawArray, ReadOnlySpan<long> dimensions, sbyte origType, ScalarType? dtype, Device? device, bool requiresGrad)
         {
+            var dataHandle = GCHandle.Alloc(rawArray, GCHandleType.Pinned);
+            var dataArrayAddr = dataHandle.AddrOfPinnedObject();
+            var gchp = GCHandle.ToIntPtr(dataHandle);
+            GCHandleDeleter deleter = null!;
+            deleter = new GCHandleDeleter((IntPtr ptr) => {
+                GCHandle.FromIntPtr(gchp).Free();
+                deleters.TryRemove(deleter, out deleter!);
+            });
+            deleters.TryAdd(deleter, deleter); // keep the delegate alive
+
             unsafe {
-                var dataArrayAddr = dataHandle.AddrOfPinnedObject();
-                var gchp = GCHandle.ToIntPtr(dataHandle);
-                GCHandleDeleter deleter = null!;
-                deleter = new GCHandleDeleter((IntPtr ptr) => {
-                    GCHandle.FromIntPtr(gchp).Free();
-                    deleters.TryRemove(deleter, out deleter!);
-                });
-                deleters.TryAdd(deleter, deleter); // keep the delegate alive
-                var handle = THSTensor_new(dataArrayAddr, deleter, dimensions, dimensions.Length, origType, requiresGrad);
-                if (handle == IntPtr.Zero) {
-                    GC.Collect();
-                    GC.WaitForPendingFinalizers();
-                    handle = THSTensor_new(dataArrayAddr, deleter, dimensions, dimensions.Length, origType, requiresGrad);
+                fixed (long* shape = dimensions) {
+                    var handle = THSTensor_new(dataArrayAddr, deleter, (IntPtr)shape, dimensions.Length, origType, requiresGrad);
+
+                    if (handle == IntPtr.Zero) {
+                        GC.Collect();
+                        GC.WaitForPendingFinalizers();
+                        handle = THSTensor_new(dataArrayAddr, deleter, (IntPtr)shape, dimensions.Length, origType, requiresGrad);
+                    }
+
+                    if (handle == IntPtr.Zero) { torch.CheckForErrors(); }
+                    var tensor = new Tensor(handle);
+
+                    if (device is not null) {
+                        tensor = dtype.HasValue ? tensor.to(dtype.Value, device) : tensor.to(device);
+                    } else if (dtype.HasValue) {
+                        tensor = tensor.to_type(dtype.Value);
+                    }
+                    return tensor;
                 }
-                if (handle == IntPtr.Zero) { torch.CheckForErrors(); }
-                var tensor = new Tensor(handle);
-                if (device is not null) {
-                    tensor = dtype.HasValue ? tensor.to(dtype.Value, device) : tensor.to(device);
-                } else if (dtype.HasValue) {
-                    tensor = tensor.to_type(dtype.Value);
-                }
-                return tensor;
             }
         }
 
@@ -1410,6 +1417,7 @@ namespace TorchSharp
         /// Create a 1-D tensor from an array of values, shaping it based on the input array.
         /// </summary>
         /// <remarks>The Torch runtime does not take ownership of the data, so there is no device argument.</remarks>
+        [System.Diagnostics.Contracts.Pure]
         public static Tensor tensor(IList<bool> rawArray, torch.ScalarType? dtype = null, torch.Device? device = null, bool requiresGrad = false)
         {
             return tensor(rawArray, new long[] { (long)rawArray.Count }, dtype, device, requiresGrad);
@@ -1421,6 +1429,7 @@ namespace TorchSharp
         /// <remarks>The Torch runtime does not take ownership of the data, so there is no device argument.
         ///          The input array must have rows * columns elements.
         /// </remarks>
+        [System.Diagnostics.Contracts.Pure]
         public static Tensor tensor(IList<bool> rawArray, long rows, long columns, torch.ScalarType? dtype = null, torch.Device? device = null, bool requiresGrad = false)
         {
             return tensor(rawArray, new long[] { rows, columns }, dtype, device, requiresGrad);
@@ -1432,6 +1441,7 @@ namespace TorchSharp
         /// <remarks>The Torch runtime does not take ownership of the data, so there is no device argument.
         ///          The input array must have dim0*dim1*dim2 elements.
         /// </remarks>
+        [System.Diagnostics.Contracts.Pure]
         public static Tensor tensor(IList<bool> rawArray, long dim0, long dim1, long dim2, torch.ScalarType? dtype = null, torch.Device? device = null, bool requiresGrad = false)
         {
             return tensor(rawArray, new long[] { dim0, dim1, dim2 }, dtype, device, requiresGrad);
@@ -1443,6 +1453,7 @@ namespace TorchSharp
         /// <remarks>The Torch runtime does not take ownership of the data, so there is no device argument.
         ///          The input array must have dim0*dim1*dim2*dim3 elements.
         /// </remarks>
+        [System.Diagnostics.Contracts.Pure]
         public static Tensor tensor(IList<bool> rawArray, long dim0, long dim1, long dim2, long dim3, torch.ScalarType? dtype = null, torch.Device? device = null, bool requiresGrad = false)
         {
             return tensor(rawArray, new long[] { dim0, dim1, dim2, dim3 }, dtype, device, requiresGrad);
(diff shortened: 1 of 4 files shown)


src/TorchSharp/Tensor/Tensor.Factories.cs
```diff
@@ -1369,10 +1369,18 @@ public static Tensor tensor(IList<bool> rawArray, long[] dimensions, torch.Scala
         /// <summary>
         /// Create a tensor from an array of values, shaping it based on the shape passed in.
         /// </summary>
-        public static Tensor tensor(bool[] dataArray, long[] dimensions, torch.ScalarType? dtype = null, torch.Device? device = null, bool requiresGrad = false)
+        public static Tensor tensor(bool[] rawArray, long[]? dimensions = null, torch.ScalarType? dtype = null, torch.Device? device = null, bool requiresGrad = false)
+        {
+            if (dimensions is null)
+                dimensions = new long[] { rawArray.LongLength };
```
tarekgh, line 1375: >new long[]  [](http://example.com/codeflow?start=29&length=11)

can you allocate this on the stack instead?

src/TorchSharp/Tensor/Tensor.Factories.cs
```diff
@@ -1369,10 +1369,18 @@ public static Tensor tensor(IList<bool> rawArray, long[] dimensions, torch.Scala
...
+
+            var dataHandle = GCHandle.Alloc(rawArray, GCHandleType.Pinned);
+            return _tensor_generic(dataHandle, dimensions, (sbyte)ScalarType.Bool, dtype, device, requiresGrad);
+        }
+
+        private static Tensor _tensor_generic(GCHandle dataHandle, long[] dimensions, sbyte origType, ScalarType? dtype, Device? device, bool requiresGrad)
         {
             unsafe {
```
tarekgh, line 1383: >unsafe [](http://example.com/codeflow?start=12&length=6)

why this is needed?

tarekgh:
@NiklasGustafsson do you have any perf numbers before and after your change?

NiklasGustafsson:
> @NiklasGustafsson do you have any perf numbers before and after your change?

Yes, it's about six orders of magnitude faster.

src/TorchSharp/Tensor/Tensor.Factories.cs
NiklasGustafsson: It was there before, so I kept it. Maybe it's not needed?

src/TorchSharp/Tensor/Tensor.Factories.cs
NiklasGustafsson: Yeah, I was just thinking about that. It should work.

src/TorchSharp/Tensor/Tensor.Factories.cs
```diff
@@ -1381,11 +1389,11 @@ public static Tensor tensor(bool[] dataArray, long[] dimensions, torch.ScalarTyp
                     deleters.TryRemove(deleter, out deleter!);
                 });
```
tarekgh, line 1390: This is a little confusing form me. it takes IntPtr as a parameter and not used while it uses `gchp` which is defined outside the delegate. does the caller of this delegate pass the right value to use inside? 

Note, using `gchp` it will force allocating a closure object wrapping `gchp` to use when the delegate get called.

NiklasGustafsson: This really old code that has survived since before I started working on TorchSharp, so I don't actually know the exact details. The basic idea is that the underlying storage is shared, so there's no copy, and once the tensor is dead, the pinning of the .NET array is no longer needed. If there's a better way to accomplish that...

NiklasGustafsson: The delegate will be called back from the native runtime when the tensor goes away.

tarekgh: and the native caller will pass the value to the delegate which be exact to `gchp` in this case. right?

src/TorchSharp/Tensor/Tensor.Factories.cs
```diff
@@ -1369,10 +1369,18 @@ public static Tensor tensor(IList<bool> rawArray, long[] dimensions, torch.Scala
...
+        public static Tensor tensor(bool[] rawArray, long[]? dimensions = null, torch.ScalarType? dtype = null, torch.Device? device = null, bool requiresGrad = false)
+        {
+            if (dimensions is null)
+                dimensions = new long[] { rawArray.LongLength };
+
+            var dataHandle = GCHandle.Alloc(rawArray, GCHandleType.Pinned);
+            return _tensor_generic(dataHandle, dimensions, (sbyte)ScalarType.Bool, dtype, device, requiresGrad);
+        }
+
+        private static Tensor _tensor_generic(GCHandle dataHandle, long[] dimensions, sbyte origType, ScalarType? dtype, Device? device, bool requiresGrad)
```
tarekgh, line 1381: >_tensor_generic [](http://example.com/codeflow?start=30&length=15)

why this method not taking the rawArray as object and does the allocation inside? this will not require all caller to do the allocation manually before calling it.

src/TorchSharp/Tensor/Tensor.Factories.cs
NiklasGustafsson: Just ran it in the debugger -- ptr and ghcp are different when the delegate is called, so it seems we need to capture ghcp.

tarekgh: ok, but this makes me worry more as the native code asks to delete something else. I hope we are not leaking anything here. It would be nice if someone could investigate later to understand how things work here. This is not a blocker to this PR as we already had this logic before your change.

src/TorchSharp/Tensor/Tensor.Factories.cs
NiklasGustafsson: So, I would have to use a fixed block, to pass the shape, right? It should probably do that, anyway, just not correctly implemented today.

src/TorchSharp/Tensor/Tensor.Factories.cs
tarekgh: Could be the native code is calling with the address `dataArrayAddr`. It is better to keep as it is for now.

TheBlackPlague (changes requested):
There are a few changes I'd request, but otherwise, LGTM. Great job! 👍

src/TorchSharp/Tensor/Tensor.Factories.cs
TheBlackPlague: I think `Span<long> dimensions = stackalloc long[1]` would work just fine. 👍

tarekgh: If you area going to use pointers, like allocation using stackalloc, then you don't need fixed block but you will use unsafe block. 

If you use Span, you will not need fixed nor unsafe block.

TheBlackPlague: > If you area going to use pointers, like allocation using stackalloc, then you don't need fixed block but you will use unsafe block.
> 
> If you use Span, you will not need fixed nor unsafe block.

I think `Span<>` should be fine here. It is practically the same thing, and you avoid the dirty `unsafe` keyword.

NiklasGustafsson: I've only ever used pointer when passing arrays to native code -- can I go straight from a ReadOnlySpan to IntPtr without unsafe or fixed?

src/TorchSharp/Tensor/Tensor.Factories.cs
NiklasGustafsson: I'll open an issue on this -- investigate why ghcp and ptr are different.

src/TorchSharp/Tensor/Tensor.Factories.cs
```diff
@@ -2312,23 +2192,56 @@ public static Tensor tensor(IList<(float Real, float Imaginary)> rawArray, long
         /// </summary>
         public static Tensor tensor((float Real, float Imaginary)[,] rawArray, torch.ScalarType? dtype = null, torch.Device? device = null, bool requiresGrad = false)
         {
-            return tensor(rawArray.Cast<(float Real, float Imaginary)>().ToArray(), new long[] { rawArray.GetLongLength(0), rawArray.GetLongLength(1) }, dtype, device, requiresGrad);
+            var dataArray = new float[rawArray.GetLongLength(0), rawArray.GetLongLength(1) * 2];
+            for (var i = 0; i < rawArray.GetLongLength(0); i++) {
+                for (var j = 0; j < rawArray.GetLongLength(0); j++) {
```
tarekgh, line 2197: >0 [](http://example.com/codeflow?start=59&length=1)

should be 1?

src/TorchSharp/Tensor/Tensor.Factories.cs
```diff
@@ -2312,23 +2192,56 @@ public static Tensor tensor(IList<(float Real, float Imaginary)> rawArray, long
...
 
         /// <summary>
         /// Create a three-dimensional tensor from a three-dimensional array of values.
         /// </summary>
         public static Tensor tensor((float Real, float Imaginary)[,,] rawArray, torch.ScalarType? dtype = null, torch.Device? device = null, bool requiresGrad = false)
         {
-            return tensor(rawArray.Cast<(float Real, float Imaginary)>().ToArray(), new long[] { rawArray.GetLongLength(0), rawArray.GetLongLength(1), rawArray.GetLongLength(2) }, dtype, device, requiresGrad);
+            var dataArray = new float[rawArray.GetLongLength(0), rawArray.GetLongLength(1), rawArray.GetLongLength(2) * 2];
+            for (var i = 0; i < rawArray.GetLongLength(0); i++) {
+                for (var j = 0; j < rawArray.GetLongLength(0); j++) {
```
tarekgh, line 2214: >0 [](http://example.com/codeflow?start=59&length=1)

should be 1?

src/TorchSharp/Tensor/Tensor.Factories.cs
```diff
@@ -2312,23 +2192,56 @@ public static Tensor tensor(IList<(float Real, float Imaginary)> rawArray, long
...
+                    for (var k = 0; k < rawArray.GetLongLength(0); k++) {
```
tarekgh, line 2215: >0 [](http://example.com/codeflow?start=63&length=1)

should be 2?

src/TorchSharp/Tensor/Tensor.Factories.cs
```diff
@@ -2312,23 +2192,56 @@ public static Tensor tensor(IList<(float Real, float Imaginary)> rawArray, long
...
-            return tensor(rawArray.Cast<(float Real, float Imaginary)>().ToArray(), new long[] { rawArray.GetLongLength(0), rawArray.GetLongLength(1), rawArray.GetLongLength(2), rawArray.GetLongLength(3) }, dtype, device, requiresGrad);
+            var dataArray = new float[rawArray.GetLongLength(0), rawArray.GetLongLength(1), rawArray.GetLongLength(2), rawArray.GetLongLength(3) * 2];
+            for (var i = 0; i < rawArray.GetLongLength(0); i++) {
+                for (var j = 0; j < rawArray.GetLongLength(0); j++) {
+                    for (var k = 0; k < rawArray.GetLongLength(0); k++) {
+                        for (var l = 0; l < rawArray.GetLongLength(0); l++) {
+                            dataArray[i, j, k, l * 2] = rawArray[i, j, k, l].Real;
+                            dataArray[i, j, k, l * 2 + 1] = rawArray[i, j, k, l].Imaginary;
+                        }
+                    }
```
tarekgh, line 2239: GetLongLength(1), (2), (3)?

src/TorchSharp/Tensor/Tensor.Factories.cs
NiklasGustafsson: Well, that's just embarrassing! Thank you for catching it.

src/TorchSharp/Tensor/Tensor.Factories.cs
```diff
@@ -2420,23 +2337,56 @@ public static Tensor tensor(IList<System.Numerics.Complex> rawArray, long dim0,
         /// </summary>
         public static Tensor tensor(System.Numerics.Complex[,] rawArray, torch.ScalarType? dtype = null, torch.Device? device = null, bool requiresGrad = false)
         {
-            return tensor(rawArray.Cast<System.Numerics.Complex>().ToArray(), new long[] { rawArray.GetLongLength(0), rawArray.GetLongLength(1) }, dtype, device, requiresGrad);
+            var dataArray = new double[rawArray.GetLongLength(0), rawArray.GetLongLength(1) * 2];
+            for (var i = 0; i < rawArray.GetLongLength(0); i++) {
+                for (var j = 0; j < rawArray.GetLongLength(0); j++) {
```
tarekgh, line 2342: >0 [](http://example.com/codeflow?start=59&length=1)

1?

src/TorchSharp/Tensor/Tensor.Factories.cs
```diff
@@ -2420,23 +2337,56 @@ public static Tensor tensor(IList<System.Numerics.Complex> rawArray, long dim0,
...
         /// Create a four-dimensional tensor from a four-dimensional array of values.
         /// </summary>
         public static Tensor tensor(System.Numerics.Complex[,,,] rawArray, torch.ScalarType? dtype = null, torch.Device? device = null, bool requiresGrad = false)
         {
-            return tensor(rawArray.Cast<System.Numerics.Complex>().ToArray(), new long[] { rawArray.GetLongLength(0), rawArray.GetLongLength(1), rawArray.GetLongLength(2), rawArray.GetLongLength(3) }, dtype, device, requiresGrad);
+            var dataArray = new double[rawArray.GetLongLength(0), rawArray.GetLongLength(1), rawArray.GetLongLength(2), rawArray.GetLongLength(3) * 2];
+            for (var i = 0; i < rawArray.GetLongLength(0); i++) {
+                for (var j = 0; j < rawArray.GetLongLength(0); j++) {
+                    for (var k = 0; k < rawArray.GetLongLength(0); k++) {
+                        for (var l = 0; l < rawArray.GetLongLength(0); l++) {
```
tarekgh, line 2380: (1) and (2) and (3)?

src/TorchSharp/Tensor/Tensor.Factories.cs
```diff
@@ -2420,23 +2337,56 @@ public static Tensor tensor(IList<System.Numerics.Complex> rawArray, long dim0,
...
         /// <summary>
         /// Create a three-dimensional tensor from a three-dimensional array of values.
         /// </summary>
         public static Tensor tensor(System.Numerics.Complex[,,] rawArray, torch.ScalarType? dtype = null, torch.Device? device = null, bool requiresGrad = false)
         {
-            return tensor(rawArray.Cast<System.Numerics.Complex>().ToArray(), new long[] { rawArray.GetLongLength(0), rawArray.GetLongLength(1), rawArray.GetLongLength(2) }, dtype, device, requiresGrad);
+            var dataArray = new double[rawArray.GetLongLength(0), rawArray.GetLongLength(1), rawArray.GetLongLength(2) * 2];
+            for (var i = 0; i < rawArray.GetLongLength(0); i++) {
+                for (var j = 0; j < rawArray.GetLongLength(0); j++) {
+                    for (var k = 0; k < rawArray.GetLongLength(0); k++) {
```
tarekgh, line 2360: (1) and (2)?

tarekgh (dismissed):
Modulo the comments and suggestions, LGTM.

src/TorchSharp/Tensor/Tensor.Factories.cs
NiklasGustafsson: https://github.com/dotnet/TorchSharp/issues/672

NiklasGustafsson:
> Modulo the comments and suggestions, LGTM.

Thank you! As always, couldn't have done it without you! :-)

test/TorchSharpTest/TestTorchTensor.cs
```diff
@@ -891,6 +982,7 @@ public void TestMDTensorFactoryFloat64()
                 Assert.Equal(3, t.ndim);
                 Assert.Equal(new long[] { 2, 2, 2 }, t.shape);
                 Assert.Equal(ScalarType.Float64, t.dtype);
+                Assert.Equal(array.Cast<double>().ToArray(), t.data<double>().ToArray());
             }
         }
```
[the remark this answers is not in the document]
NiklasGustafsson: Sure.

src/TorchSharp/Tensor/Tensor.Factories.cs
tarekgh: `(IntPtr)MemoryMarshal.GetReference(buffer)` where buffer is the span.

test/TorchSharpTest/TestTorchTensor.cs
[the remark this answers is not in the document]
NiklasGustafsson: 1GB big enough?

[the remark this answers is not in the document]
TheBlackPlague: Well, I think tests for `[10000, 10000]` should be more than enough.

src/TorchSharp/Tensor/Tensor.Factories.cs
```diff
@@ -1477,39 +1500,19 @@ public static Tensor tensor(IList<byte> rawArray, long[] dimensions, torch.Scala
         /// <summary>
         /// Create a tensor from an array of values, shaping it based on the shape passed in.
         /// </summary>
-        public static Tensor tensor(byte[] dataArray, long[] dimensions, torch.ScalarType? dtype = null, torch.Device? device = null, bool requiresGrad = false)
+        [System.Diagnostics.Contracts.Pure]
+        public static Tensor tensor(byte[] rawArray, long[]? dimensions = null, torch.ScalarType? dtype = null, torch.Device? device = null, bool requiresGrad = false)
```
TheBlackPlague, line 1504: I believe the input type of dimensions could be `Span`. Since, `Span` can be created from heap arrays too.  Overall, will allow both ways increasing user control while not affecting performance (and giving better performance for those that may want it).

TheBlackPlague (changes requested):
I think some expressions are left that could be changed to `Span`. Other than that, the PR looks great!

src/TorchSharp/Tensor/Tensor.Factories.cs
NiklasGustafsson: ReadOnlySpan<long> can't be nullable, I believe, so I'll have to duplicate every single method...

NiklasGustafsson: But not a big deal.

tarekgh: can you use empty span for the null case?

TheBlackPlague: Does `ReadOnlySpan<T>?` not work as a type?

tarekgh: Yes, it cannot be nullable. The span is defined as `ref struct` which cannot be nullable.

NiklasGustafsson: > Does `ReadOnlySpan<long>?` not work as a type? I imagined it would.

Seems like no. Not a big deal, duplicating works, too.

tarekgh: @NiklasGustafsson can you use empty span for the null case?

Issue #679, opened 2022-08-04
The default value of `onesided` of `torch.istft()`  is not aligned with PyTorch


kaiidams:
PyTorch uses the tensor shape to determine `onesided` (n_fft != fft_size). TorchShape defaults to `onesided=true` when spec is not complex (like `torch.stft()`).

```python
dtype = torch.float32
spec = torch.rand([1, 1024, 500, 2], dtype=dtype)
x = torch.istft(spec, 512, 160, 400, None)
=> RuntimeError

spec = torch.rand([1, 512, 500, 2], dtype=dtype)
x = torch.istft(spec, 512, 160, 400, None)
=> OK

spec = torch.rand([1, 257, 500, 2], dtype=dtype)
x = torch.istft(spec, 512, 160, 400, None)
=> OK

dtype = torch.complex64
spec = torch.rand([1, 1024, 500], dtype=dtype)
x = torch.istft(spec, 512, 160, 400, None)
=> RuntimeError

spec = torch.rand([1, 512, 500], dtype=dtype)
x = torch.istft(spec, 512, 160, 400, None)
=> OK

spec = torch.rand([1, 257, 500], dtype=dtype)
x = torch.istft(spec, 512, 160, 400, None)
=> OK
```

NiklasGustafsson:
@kaiidams -- I repro'd this and have a fix. It'll be included in an upcoming PR.

PR #680, opened 2022-08-04
Rationalize how tensors are created from arrays.

NiklasGustafsson:
Per discussion in #670, the following will be how to create a tensor from arrays, aligned with Pytorch, treating `from_array()` as the equivalent of `from_numpy()`:

```C#
// Never copy:
public static Tensor from_array(Array rawArray)

// Copy if necessary:
public static Tensor frombuffer(Array rawArray, ScalarType dtype, long count = -1, long offset = 0, bool requiresGrad = false, Device? device = null)
public static Tensor as_tensor(Array data, ScalarType? dtype = null, Device? device = null)
public static Tensor as_tensor(Tensor data, ScalarType? dtype = null, Device? device = null)

// Always copy:
public static Tensor tensor(<<VARIOUS>>, torch.Device? device = null, bool requiresGrad = false)
```

This PR also addresses #679.

### DIFF
--- a/src/TorchSharp/Tensor/Tensor.cs
+++ b/src/TorchSharp/Tensor/Tensor.cs
@@ -6998,7 +6998,7 @@ namespace TorchSharp
             }
 
             [DllImport("LibTorchSharp")]
-            extern static IntPtr THSTensor_stft(IntPtr x, long n_fft, long hop_length, long win_length, IntPtr window, bool normalized, bool onesided, bool return_complex);
+            extern static IntPtr THSTensor_stft(IntPtr x, long n_fft, long hop_length, long win_length, IntPtr window, bool normalized, long onesided, bool return_complex);
 
             /// <summary>
             /// Short-time Fourier transform (STFT).
@@ -7017,7 +7017,11 @@ namespace TorchSharp
             {
                 IntPtr _input = Handle;
                 IntPtr _window = (window is null) ? IntPtr.Zero : window.Handle;
-                bool _onesided = onesided.HasValue ? onesided.Value : !is_complex();
+
+                long _onesided = -1; // encoding of null
+                if (onesided.HasValue) {
+                    _onesided = (onesided.Value ? 1 : 0);
+                }
                 bool _return_complex = return_complex.HasValue ? return_complex.Value : is_complex();
 
                 if (center) {
@@ -7068,11 +7072,9 @@ namespace TorchSharp
                 var fft_size = shape[1];
                 IntPtr _window = (window is null) ? IntPtr.Zero : window.Handle;
 
-                int _onesided = -1; // encoding of Python 'None'
+                long _onesided = -1; // encoding of null
                 if (onesided.HasValue) {
                     _onesided = (onesided.Value ? 1 : 0);
-                } else if (n_fft != fft_size) {
-                    _onesided = 1;
                 }
 
                 var res = THSTensor_istft(Handle, n_fft, hop_length, win_length, _window, center, normalized, _onesided, length, return_complex);
--- a/src/TorchSharp/Tensor/Tensor.cs
+++ b/src/TorchSharp/Tensor/Tensor.cs
@@ -7176,6 +7176,35 @@ namespace TorchSharp
             BFloat16 = 15
         }
 
+        internal static ScalarType ToScalarType(Type t)
+        {
+            switch (true) {
+            case bool _ when t == typeof(bool):
+                return ScalarType.Bool;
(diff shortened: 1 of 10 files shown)


NiklasGustafsson:
@kaiidams -- please take a look at the modifications to istft().

@tarekgh, @MovGP0, @TheBlackPlague -- please take a look at the modifications to tensor creation.

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -7065,8 +7065,16 @@ public Tensor stft(long n_fft, long hop_length = -1, long win_length = -1, Tenso
...
+                var fft_size = shape[1];
                 IntPtr _window = (window is null) ? IntPtr.Zero : window.Handle;
-                bool _onesided = onesided.HasValue ? onesided.Value : !is_complex();
+
+                int _onesided = -1; // encoding of Python 'None'
+                if (onesided.HasValue) {
+                    _onesided = (onesided.Value ? 1 : 0);
+                } else if (n_fft != fft_size) {
+                    _onesided = 1;
+                }
```
kaiidams, line 7076: Probably these are handled in C++ code of PyTorch.

https://github.com/pytorch/pytorch/blob/4906a956181fb337e99860fa3e6071473c0e30e6/aten/src/ATen/native/SpectralOps.cpp#L1005

```
  const bool onesided = onesidedOpt.value_or(fft_size != n_fft);
```

kaiidams (dismissed):

NiklasGustafsson:
By the way, I know that this PR violates the last of the DO NOT rules in the CONTRIBUTING.md document, but mixing unrelated changes in one PR. Sorry about that.

src/TorchSharp/Tensor/Tensor.cs
NiklasGustafsson: I've added hfft2/hfftn/ihfft2/ihfftn to 'torch' in this PR, too.

GeorgeS2019:
@NiklasGustafsson 

Often when we are using pytorch to prepare tensor data for input to ONNX, `from_numpy()` is often used to convert raw data to tensor. => [**TORCH.FROM_NUMPY**](https://pytorch.org/docs/stable/generated/torch.from_numpy.html)

### Question 1
Not sure if this discussion would eventually evolve towards [**Torch and numpy interoperability**](https://github.com/dotnet/TorchSharp/discussions/394)

### Question 2
Not sure how much of the [thinking considerations **below** ](https://pytorch.org/docs/stable/generated/torch.from_numpy.html)will be considered.

> The function torch.from_numpy() provides support for the conversion of a numpy array into a tensor in PyTorch. It expects the input as a numpy array (numpy.ndarray). The output type is tensor. The returned tensor and ndarray share the same memory. The returned tensor is not resizable.

> It currently accepts **ndarray with dtypes of numpy.float64, numpy.float32, numpy.float16, numpy.int64, numpy.int32, numpy.int16, numpy.int8, numpy.uint8, and numpy.bool.**

### Question 3
Will this evolve into such that the following code will no longer involve` import numpy`. Instead of using `torch.from_numpy(a)`, now there is a substitute `torch.from_array(a)` Torchsharp equivalent as implemented here?
```python
# Importing the PyTorch library
import torch
import numpy

# A numpy array of size 6
a = numpy.array([1.0, -0.5, 3.4, -2.1, 0.0, -6.5])
print(a)

# Applying the from_numpy function and
# storing the resulting tensor in 't'
t = torch.from_numpy(a)
print(t)
```

NiklasGustafsson:
@GeorgeS2019 -- this is unrelated to support for numpy, which is a Python library. The only reason that `from_numpy()` is mentioned here is because it served as the template for what I propose that the semantics of `from_array()` should be. That's it.

Question 1 -- no, I don't see that this discussion will do so. It's unrelated.
Question 2 -- the modified semantics of `from_array()` that I'm proposing is specifically based on that thinking, i.e. the semantics of `from_numpy()`
Question 3 -- I'm not sure I understand your question. This PR has nothing to do with what you import in Python code.
Question 4 -- since TorchSharp does not support `torch.from_numpy()`, I don't see why we would do that.

NiklasGustafsson:
BTW, @GeorgeS2019, since you are such a frequent contributor to discussions here, it would be great if you would introduce yourself to the community here: https://github.com/dotnet/TorchSharp/discussions/333

Not a requirement, just something I think everyone would appreciate.

NiklasGustafsson:
I feel good about the implementation, and have received no pushback on the modified semantics, so I will merge this PR now.

PR #688, opened 2022-08-08
Add 'training' flag to dropout functions in torch.nn.functional

NiklasGustafsson:


NiklasGustafsson:
@dsyme -- I mistakenly requested a review from you. Please feel free to ignore.

NiklasGustafsson:
@kaiidams -- maybe you can take a look, too?

src/TorchSharp/NN/AlphaDropout.cs
```diff
@@ -60,19 +60,19 @@ static public AlphaDropout AlphaDropout(double probability = 0.5, bool inPlace =
             }
             public static partial class functional
             {
+                [DllImport("LibTorchSharp")]
+                extern static IntPtr THSNN_alpha_dropout(IntPtr input, double probability, bool training, bool inPlace);
```
tarekgh, line 64: > extern static IntPtr THSNN_alpha_dropout(IntPtr input, double probability, bool training, bool inPlace); [](http://example.com/codeflow?start=15&length=105)

Could you please double check the rule [https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1414?ranMID=43674&ranEAID=rl2xnKiLcHs&ranSiteID=rl2xnKiLcHs-HFITgw5K.GZz_fG_GrdzpQ&epi=rl2xnKiLcHs-HFITgw5K.GZz_fG_GrdzpQ&irgwc=1&OCID=AID2200057_aff_7795_1243925&tduid=(ir__2m3q0nl02wkf6gcatnnkkvci0e2xvdeew13xgf9200)(7795)(1243925)(rl2xnKiLcHs-HFITgw5K.GZz_fG_GrdzpQ)()&irclickid=_2m3q0nl02wkf6gcatnnkkvci0e2xvdeew13xgf9200&view=vs-2022&tabs=csharp#rule-description](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1414?ranMID=43674&ranEAID=rl2xnKiLcHs&ranSiteID=rl2xnKiLcHs-HFITgw5K.GZz_fG_GrdzpQ&epi=rl2xnKiLcHs-HFITgw5K.GZz_fG_GrdzpQ&irgwc=1&OCID=AID2200057_aff_7795_1243925&tduid=(ir__2m3q0nl02wkf6gcatnnkkvci0e2xvdeew13xgf9200)(7795)(1243925)(rl2xnKiLcHs-HFITgw5K.GZz_fG_GrdzpQ)()&irclickid=_2m3q0nl02wkf6gcatnnkkvci0e2xvdeew13xgf9200&view=vs-2022&tabs=csharp#rule-description) for the boolean parameters to ensure we are doing the right thing?

NiklasGustafsson: Well, that's interesting. We've been ignoring the MarshalAs attribute since the beginning. Surprised it has worked.

NiklasGustafsson: Lots of native methods in TorchSharp use bools without the attribute.

tarekgh: We could be lucky enough with that :-) the issue with the boolean is the interop uses 32-bit for it which could be not the exact size used in the native.

NiklasGustafsson: Not sure that I understand, but let's fix this across the board. If I understand correctly, each bool should either be fixed on the native side as a uint32_t, or on the managed side using [MarshalAs(UnmanagedType.U1)]

NiklasGustafsson: Presumably, the C++ compiler was aligning everything on 32-bit boundaries -- is that what you're saying.

Latest update adds the MarshalAs on all booleans.

tarekgh: My understanding is size of bool in C++ is one byte.  and the managed interop will pass it as 32-bit value. So, C++ will use just the first byte which will work. I am not sure if we'll have the same luck if running on machines with [big-endian](https://en.wikipedia.org/wiki/Endianness) machines.

tarekgh (approved):
LGTM.  Thanks!

src/TorchSharp/TorchAudio/Modules/Tacotron2.cs
```diff
@@ -384,13 +384,13 @@ public override Tensor forward(Tensor x)
...
                     if (i < this.n_convs - 1) {
                         if (this.training) {
-                            x = F.Dropout(torch.tanh(conv.forward(x)), 0.5);
+                            x = F.dropout(torch.tanh(conv.forward(x)), 0.5);
                         } else {
                             x = torch.tanh(conv.forward(x));
                         }
                     } else {
                         if (this.training) {
-                            x = F.Dropout(conv.forward(x), 0.5);
```
kaiidams, line 393: This can be written like
```
x = F.dropout(conv.forward(x), p: 0.5, training: this.training);
```
I can update `tacotron2.cs` later.

kaiidams (approved):

PR #700, opened 2022-08-16
Additional distributions.

NiklasGustafsson:


src/TorchSharp/Distributions/Constraints.cs
```diff
@@ -0,0 +1,637 @@
...
+                {
+                    /// <summary>
+                    /// Constructor
+                    /// </summary>
+                    /// <param name="is_discrete">Whether constrained space is discrete.</param>
+                    /// <param name="event_dim">
+                    /// Number of rightmost dimensions that together define an event.
+                    /// The check() method will remove this many dimensions when computing validity.
+                    /// </param>
+                    protected Constraint(bool is_discrete = false, int event_dim = 0)
```
tarekgh, line 33: I suggest adding default constructor something like:

protected Constraint() : this(false, 0) {}

then make this constructor with no default parameters 

protected Constraint(bool is_discrete, int event_dim) {...}

This to be explicit and avoid having the compiler adding the default constructor.

src/TorchSharp/Distributions/Constraints.cs
```diff
@@ -0,0 +1,637 @@
...
+                    /// The check() method will remove this many dimensions when computing validity.
+                    /// </summary>
+                    public override int event_dim { get => base_constraint.event_dim; }
+
+                }
+
+                /// <summary>
+                /// Constrain to the two values {0, 1}.
+                /// </summary>
+                public class _Boolean : Constraint
```
tarekgh, line 117: >_Boolean [](http://example.com/codeflow?start=29&length=8)

Is it how it is defined with pytorch (started with underscore)?

src/TorchSharp/Distributions/Constraints.cs
```diff
@@ -0,0 +1,637 @@
...
+                        this.upper_bound = upper_bound;
+                    }
+
+                    public long lower_bound { get; private set; }
+
+                    public long upper_bound { get; private set; }
+
+                    public override Tensor check(Tensor value)
+                    {
+                        return (value % 1 == 0) & (lower_bound <= value) & (value <= upper_bound);
```
tarekgh, line 156: >& [](http://example.com/codeflow?start=48&length=1)

I assume it is intentional to use the bit-operators here.

src/TorchSharp/Distributions/Constraints.cs
```diff
@@ -0,0 +1,637 @@
...
+                    public _LessThan(double upper_bound) : base(true)
+                    {
+                        this.upper_bound = upper_bound;
+                    }
+
+                    public double upper_bound { get; private set; }
+
+                    public override Tensor check(Tensor value)
+                    {
+                        return (value <= upper_bound);
```
tarekgh, line 333: >= [](http://example.com/codeflow?start=39&length=1)

is it intentional to include `=`?

src/TorchSharp/Distributions/GumbEL.cs
```diff
@@ -0,0 +1,81 @@
...
+            public override Tensor variance => stddev.pow(2);
+
+            public override Tensor stddev => pioversqrtsix * scale;
+
+            /// <summary>
+            /// Returns entropy of distribution, batched over batch_shape.
+            /// </summary>
+            public override Tensor entropy() => scale.log() + (1 + euler_constant);
+
+            private double pioversqrtsix = Math.PI / Math.Sqrt(6);
```
tarekgh, line 45: >pioversqrtsix [](http://example.com/codeflow?start=27&length=13)

this should be const?

src/TorchSharp/Distributions/Constraints.cs
NiklasGustafsson: Yes. Those are boolean tensors being ANDed together, not boolean expressions, where short-circuit operators would be appropriate.

src/TorchSharp/Distributions/Constraints.cs
NiklasGustafsson: Yes. I held my .NET naming conventions nose and called everything the same as Pytorch does.

src/TorchSharp/Distributions/GumbEL.cs
NiklasGustafsson: Yes, it should. Thanks!

NiklasGustafsson: Take that back -- it's not initialized to a compile-time constant, so that will generate and error. 'readonly,' perhaps?

src/TorchSharp/Tensor/Size.cs
```diff
@@ -20,6 +21,19 @@ public static partial class torch
...
+            /// Represents an empty size.
+            /// </summary>
+            public static Size Empty = new Size();
+
+            /// <summary>
+            /// Constructor
+            /// </summary>
+            public Size()
+            {
+                _shape = new long[0];
```
tarekgh, line 34: >new long[0]; [](http://example.com/codeflow?start=25&length=12)

`Array.Empty<long>`

src/TorchSharp/Distributions/GumbEL.cs
tarekgh: I meant calculate it offline and use the value. 

```C#
private const pioversqrtsix  = 1.2825498301618641; //  Math.PI / Math.Sqrt(6);
```

NiklasGustafsson: Of course. Thanks!

tarekgh (approved):

src/TorchSharp/Distributions/Constraints.cs
NiklasGustafsson: Okay. I don't understand why this is useful/important, but I'll take your word for it.

PR #711, opened 2022-08-27
Add function methods `torch.sum()`, and etc.

kaiidams:
- torch.argmax()
- torch.argmin()
- torch.argsort()
- torch.normal(double, double)
- torch.sum()

### DIFF
--- a/src/TorchSharp/Tensor/Tensor.Factories.cs
+++ b/src/TorchSharp/Tensor/Tensor.Factories.cs
@@ -951,6 +951,22 @@ namespace TorchSharp
         }
 
         /// <summary>
+        /// Similar to the function above, but the means and standard deviations are shared among all drawn elements. The resulting tensor has size given by size.
+        /// </summary>
+        /// <param name="mean">The mean for all distributions</param>
+        /// <param name="std">The standard deviation for all distributions</param>
+        /// <param name="size">A sequence of integers defining the shape of the output tensor.</param>
+        /// <param name="dtype"></param>
+        /// <param name="device"></param>
+        /// <param name="requires_grad"></param>
+        /// <param name="generator">An optional random number generator</param>
+        /// <returns></returns>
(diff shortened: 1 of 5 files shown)


GeorgeS2019:
## Math operations
### Pointwise Ops
- [x] Torch.Sum

src/TorchSharp/Tensor/Tensor.Factories.cs
```diff
@@ -950,6 +950,22 @@ public static Tensor normal(Tensor mean, Tensor std, torch.Generator? generator
...
+        /// <param name="std">The standard deviation for all distributions</param>
+        /// <param name="size">A sequence of integers defining the shape of the output tensor.</param>
+        /// <param name="dtype"></param>
+        /// <param name="device"></param>
+        /// <param name="requires_grad"></param>
+        /// <param name="generator">An optional random number generator</param>
+        /// <returns></returns>
+        public static Tensor normal(double mean, double std, ReadOnlySpan<long> size, torch.ScalarType? dtype = null, torch.Device? device = null, bool requires_grad = false, torch.Generator? generator = null)
+        {
+            return randn(size, dtype: dtype, device: device, requiresGrad: requires_grad, generator: generator) * std + mean;
```
NiklasGustafsson, line 966: If device is 'null,' should it use the std.device instead of the default device, perhaps?

kaiidams: `torch.rand()` should handle the case where `device` is `null`. As the user may call it like `torch.rand(size, dtype: torch.float64)` without `device`. `torch.InitializeDevice(null)` falls back to `torch.CPU`. We may want to call C++ `torch::rand()` function without device option if we want to make sure it is compatible with PyTorch?

NiklasGustafsson: @kaiidams -- I was confused. I was thinking that std and mean were tensors, and that the device should follow them unless non-null. std and mean are, of course, doubles, so the default device makes sense.

NiklasGustafsson (approved):

PR #716, opened 2022-09-12
Tensorboard support

NiklasGustafsson:
Adding support for Tensorboard -- scalars.


NiklasGustafsson:
@kaiidams -- could you also take a look at this PR?

NiklasGustafsson:
@QingtaoLi1 -- can you also take a look at this PR?

src/Native/LibTorchSharp/crc32c.c
```diff
@@ -0,0 +1,930 @@
+/* Part of CRC-32C library: https://crc32c.machinezoo.com/ */
```
tarekgh, line 1: >Part of CRC-32C library: https://crc32c.machinezoo.com/ */ [](http://example.com/codeflow?start=3&length=58)

do we want to add something to [https://github.com/dotnet/TorchSharp/blob/main/THIRD-PARTY-NOTICES.TXT](https://github.com/dotnet/TorchSharp/blob/main/THIRD-PARTY-NOTICES.TXT) about this?

src/TorchSharp/Utils/CRC32C.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.Runtime.InteropServices;
+
+namespace TorchSharp.Utils
+{
+    public static class CRC32C
+    {
+        // uint32_t crc32c_append(uint32_t crc, buffer input, size_t length)
+
+        [DllImport("LibTorchSharp")]
+        extern static uint crc32c_append(uint crc, IntPtr value, ulong length);
```
lostmsu, line 37: IMHO, it would be better to rewrite the original C file in C#. The differences would be minimal.

src/TorchSharp/Utils/tensorboard/protobuf/Graph.cs
```diff
@@ -0,0 +1,310 @@
+// <auto-generated>
+//     Generated by the protocol buffer compiler.  DO NOT EDIT!
+//     source: tensorboard/compat/proto/graph.proto
+// </auto-generated>
+#pragma warning disable 1591, 0612, 3021
+#region Designer generated code
```
lostmsu, line 6: IMHO, non-torch specific parts of the TensorBoard code should be built as a separate NuGet package.

src/TorchSharp/Utils/CRC32C.cs
```diff
@@ -0,0 +1,70 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Compule the CRC32C value for a 32-bit integer.
+        /// </summary>
+        /// <param name="data">A byte array.</param>
+        public static unsafe uint process(int data)
+        {
+            fixed (int* ptr = stackalloc int[] { data })
```
tarekgh, line 56: >fixed [](http://example.com/codeflow?start=12&length=5)

I don't think you need `fixed` statement here. This apply to other method below too.

src/TorchSharp/Utils/tensorboard/SummaryWriter.cs
```diff
@@ -0,0 +1,262 @@
...
+                        log_dir = CreateRunName(log_dir);
+                    }
+
+                    return new Modules.SummaryWriter(log_dir, filename_suffix);
+                }
+
+                internal static string CreateRunName(string log_dir)
+                {
+                    var now = DateTime.Now;
+                    var name = $"{Modules.SummaryWriter._months[now.Month - 1]}{now.Day}_{now.Hour}-{now.Minute}-{now.Second}_{System.Net.Dns.GetHostName()}";
```
tarekgh, line 46: >_months [](http://example.com/codeflow?start=56&length=7)

nit: may be can get rid of `_months` and use `CultureInfo.InvariantCulture.DateTimeFormat.AbbreviatedMonthNames` instead?

NiklasGustafsson: Thanks for that hint -- I wasn't aware of its existence.

src/TorchSharp/Utils/CRC32C.cs
NiklasGustafsson: Hmm. The compiler complains when I take it out.

src/Native/LibTorchSharp/crc32c.c
NiklasGustafsson: Yes, of course. Thanks for catching that.

src/TorchSharp/Utils/CRC32C.cs
NiklasGustafsson: I'm holding out hope that the HW acceleration code in the original source might be something we can bring back in when computing CRC on larger data blobs, such as images, and such. That'll require C.

src/TorchSharp/Utils/tensorboard/protobuf/AllocationDescription.cs
```diff
@@ -0,0 +1,337 @@
...
+    #region Descriptor
+    /// <summary>File descriptor for tensorboard/compat/proto/allocation_description.proto</summary>
+    public static pbr::FileDescriptor Descriptor {
+      get { return descriptor; }
+    }
+    private static pbr::FileDescriptor descriptor;
+
+    static AllocationDescriptionReflection() {
+      byte[] descriptorData = global::System.Convert.FromBase64String(
+          string.Concat(
```
tarekgh, line 26: >string.Concat( [](http://example.com/codeflow?start=10&length=14)

nit: instead of calling it, you may just use `+` with the literal string?

NiklasGustafsson: Those are files generated by the protoc compiler.

src/TorchSharp/Utils/tensorboard/protobuf/Graph.cs
NiklasGustafsson: You are definitely right on principle. Right now, with the minimal feature set, I think I'll keep it in TorchSharp, but we can definitely pull it out later. It might be useful for ML.NET, too.

src/TorchSharp/Utils/tensorboard/protobuf/AllocationDescription.cs
NiklasGustafsson: Sorry, should have mentioned -- skip all the files in the 'protobuf' folder.

tarekgh: That make the review faster :-) thanks!

src/TorchSharp/Utils/CRC32C.cs
lostmsu: Looks like simple bit shifting to me. No manual AVX calls or anything like that.

src/TorchSharp/Utils/CRC32C.cs
tarekgh: Could you please share the compiler error with me?

tarekgh (approved):

PR #720, opened 2022-09-13
add non-maximum suppression (NMS) function to torchsharp

huangyangyu:
This is a C# implementation of [NMS](https://pytorch.org/vision/0.9/ops.html?highlight=nms#torchvision.ops.nms), which is verified in object detection feature.

### DIFF
--- a/src/TorchSharp/TorchVision/Ops.cs
+++ b/src/TorchSharp/TorchVision/Ops.cs
@@ -1,9 +1,10 @@
 // Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
 
 using static TorchSharp.torch;
-using System.Range;
-using System.Index;
+using System;
+using System.Linq;
 using System.Collections.Generic;
+using System.Runtime.InteropServices;
 
 namespace TorchSharp.torchvision
 {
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/TorchSharp?pullRequest=720) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/TorchSharp?pullRequest=720) before we can accept your contribution.<br/><br/>:x: huangyangyu [sign now](https://cla.dotnetfoundation.org/dotnet/TorchSharp?pullRequest=720)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/TorchSharp?pullRequest=720) it.</sub>

src/TorchSharp/TorchVision/Ops.cs
```diff
@@ -30,5 +33,48 @@ public static Tensor sigmoid_focal_loss(Tensor inputs, Tensor targets, float alp
                 return loss.MoveToOuterDisposeScope();
             }
         }
+
+        public static Tensor nme(Tensor boxes, Tensor scores, double iou_threshold = 0.5)
```
NiklasGustafsson, line 37: Is it NMS or NME?

src/TorchSharp/TorchVision/Ops.cs
```diff
@@ -30,5 +33,48 @@ public static Tensor sigmoid_focal_loss(Tensor inputs, Tensor targets, float alp
...
+        {
+            using (var nmsScope = torch.NewDisposeScope()) {
+                var x1 = boxes[.., 0];
```
NiklasGustafsson, line 40: System.Range is only available in later version of .NET, and we need to be compatible with .NET FX 4.7.2. Therefore, you need to use the uglier syntax that System.Range turns into.

src/TorchSharp/TorchVision/Ops.cs
```diff
@@ -30,5 +33,48 @@ public static Tensor sigmoid_focal_loss(Tensor inputs, Tensor targets, float alp
...
+                    if (order.numel() == 1) {
+                        i = order.item<long>();
+                        keep.Add(i);
+                        break;
+                    } else {
+                        i = order[0].item<long>();
+                        keep.Add(i);
+                    }
+
+                    var xx1 = x1[order[1..]].clamp(min: x1[i]);
```
NiklasGustafsson, line 60: Same here -- can't use ranges in implementation code.

src/TorchSharp/TorchVision/Ops.cs
```diff
@@ -1,6 +1,9 @@
 // Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
 
 using static TorchSharp.torch;
+using System.Range;
```
NiklasGustafsson, line 4: Not available on all versions of .NET that we are targeting.

src/TorchSharp/TorchVision/Ops.cs
```diff
@@ -30,5 +33,48 @@ public static Tensor sigmoid_focal_loss(Tensor inputs, Tensor targets, float alp
                 return loss.MoveToOuterDisposeScope();
             }
         }
+
```
NiklasGustafsson, line 36: Please add doc comments.

NiklasGustafsson (changes requested):

NiklasGustafsson (commented):
Also, please add some unit tests.

src/TorchSharp/TorchVision/Ops.cs
huangyangyu: Thanks for pointing out the typo.

src/TorchSharp/TorchVision/Ops.cs
huangyangyu: changed it to tensor.select(..)

src/TorchSharp/TorchVision/Ops.cs
huangyangyu: changed it to tensor indices, like this: torch.arange(..).

src/TorchSharp/TorchVision/Ops.cs
huangyangyu: Removed the unsupported dependencies.

src/TorchSharp/TorchVision/Ops.cs
huangyangyu: Added.

src/TorchSharp/TorchVision/Ops.cs
NiklasGustafsson: You could also use: 

```C#
var idx = TensorIndex.Slice(); // Default values is the same as '..'
var x1 = boxes[idx, 0];
```

I don't know if that or 'select()' is faster.

src/TorchSharp/TorchVision/Ops.cs
NiklasGustafsson: TensorIndex.Slice(1,null) may be faster.

NiklasGustafsson (approved):

PR #737, opened 2022-09-16
Added modifiable global default for tensor string formatting.

NiklasGustafsson:
This PR allows you to set the default tensor string format using `torch.TensorStringStyle = ...`

### DIFF
--- a/src/TorchSharp/Tensor/Tensor.cs
+++ b/src/TorchSharp/Tensor/Tensor.cs
@@ -6659,7 +6659,7 @@ namespace TorchSharp
             /// Tensor-specific ToString()
             /// </summary>
             /// <param name="style">
-            /// The style to use -- either 'metadata,' 'julia,' or 'numpy'
+            /// The style to use -- either 'default,' 'metadata,' 'julia,' or 'numpy'
             /// </param>
             /// <param name="fltFormat">The floating point format to use for each individual number.</param>
             /// <param name="width">The line width to enforce</param>
--- a/src/TorchSharp/Tensor/Tensor.cs
+++ b/src/TorchSharp/Tensor/Tensor.cs
@@ -6653,7 +6653,7 @@ namespace TorchSharp
                                    string fltFormat = "g5",
                                    int width = 100,
                                    CultureInfo? cultureInfo = null,
-                                   string newLine = "") => disamb ? ToString(TensorStringStyle.Julia, fltFormat, width, cultureInfo, newLine) : ToMetadataString();
+                                   string newLine = "") => disamb ? ToString(torch.TensorStringStyle, fltFormat, width, cultureInfo, newLine) : ToMetadataString();
 
             /// <summary>
             /// Tensor-specific ToString()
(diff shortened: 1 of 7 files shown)


NiklasGustafsson:
@George2019, @dayo05 -- please take a look at this PR. The idea is to be able to set the global default string format rather than have the default be a default argument. 

This allows you to control the behavior of all `ToString()`, `print()`, and `str()` that don't take a style parameter. 'Julia' is still the static default, so you have to set `torch.TensorStringStyle = TensorStringStyle.NumPy;` to change it. Once you do that, though, everything will obey it.

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -6679,6 +6679,7 @@ public Tensor where(Tensor condition, Tensor y)
                     return ToMetadataString();
```
kaiidams, line 6679: Doc comment needs update. ` /// The style to use -- either 'metadata,' 'julia,' or 'numpy'`

kaiidams (dismissed):

dayo05:
LGTM!

GeorgeS2019:
@NiklasGustafsson  I could not get torch.metadata to work!

GeorgeS2019:
@dayo05  what u mean? 
>LGTM!

NiklasGustafsson:
> @NiklasGustafsson I could not get torch.metadata to work!

Try TensorStringStyle.Metadata

NiklasGustafsson:
> LGTM!



> @dayo05 what u mean?
> 
> > LGTM!

LGTM == Looks Good To Me

GeorgeS2019:
@NiklasGustafsson 

Why on notebook, whenever I use e.g. .`str() ` I get this as additional message?

> warning CS1701: Assuming assembly reference 'System.Runtime.Numerics, Version=4.1.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' used by 'TorchSharp' matches identity 'System.Runtime.Numerics, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' of 'System.Runtime.Numerics', you may need to supply runtime policy

tarekgh:
@GeorgeS2019 do you have link to that notebook? Maybe the notebook is using .NET Framework and referencing the old runtime version of System.Numerics.

GeorgeS2019:
FYI: https://github.com/dotnet/interactive/issues/762

NiklasGustafsson:
@GeorgeS2019 -- there must be something funky in your setup for that to happen. I've never seen anything like that going on with any notebook. Is it one of the books from TorchSharpExamples?

GeorgeS2019:
I run out of idea :-(

I compare using local and remote nuget TorchSharp. same happens for the TorchSharpExamples notebooks. Not sure where the problems :-(

GeorgeS2019:
>  referencing the old runtime version of System.Numerics.

I cleaned up my .nuget packages and all the old System.Runtime.Numerics, e.g. Version=4.1.2.0 to ONLY the latest 4.3.0 
No effect.

PR #741, opened 2022-09-17
Add torchvision.models.mobilenet_v2()

kaiidams:
https://pytorch.org/vision/stable/models/generated/torchvision.models.mobilenet_v2.html#torchvision.models.mobilenet_v2

Compatible with these state_dict's converted. https://github.com/dotnet/TorchSharp/issues/588

- https://download.pytorch.org/models/mobilenet_v2-7ebf99e0.pth
- https://download.pytorch.org/models/mobilenet_v2-b0353104.pth

The code is from main branch of torchvision that has difference from v0.12.0 (for PyTorch v1.11.0)

### DIFF
--- a/src/TorchSharp/TorchVision/Ops/Misc.cs
+++ b/src/TorchSharp/TorchVision/Ops/Misc.cs
@@ -7,7 +7,7 @@
 //
 // The origin has the following copyright notice and license:
 //
-// https://github.com/pytorch/audio/blob/main/LICENSE
+// https://github.com/pytorch/vision/blob/main/LICENSE
 //
 
 using System;
--- a/src/TorchSharp/TorchVision/Ops/Misc.cs
+++ b/src/TorchSharp/TorchVision/Ops/Misc.cs
@@ -1,0 +1,145 @@
+// Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
+
+// A number of implementation details in this file have been translated from the Python version of torchvision,
+// largely located in the files found in this folder:
+//
+// https://github.com/pytorch/vision/blob/a4f53308b2d0f1aa9191686e326f45c26053f686/torchvision/ops/misc.py
+//
+// The origin has the following copyright notice and license:
+//
+// https://github.com/pytorch/audio/blob/main/LICENSE
+//
+
+using System;
+using System.Collections.Generic;
+using static TorchSharp.torch;
+
+#nullable enable
+namespace TorchSharp
+{
+    public static partial class torchvision
+    {
+        public static partial class ops
+        {
+            /// <summary>
+            /// Configurable block used for Convolution2d-Normalization-Activation blocks.
+            /// </summary>
+            /// <param name="in_channels">Number of channels in the input image</param>
+            /// <param name="out_channels">Number of channels produced by the Convolution-Normalization-Activation block</param>
+            /// <param name="kernel_size">Size of the convolving kernel.</param>
+            /// <param name="stride">Stride of the convolution.</param>
+            /// <param name="padding">Padding added to all four sides of the input. Default: null, in which case it will calculated as ``padding = (kernel_size - 1) // 2 * dilation``</param>
+            /// <param name="groups">Number of blocked connections from input channels to output channels.</param>
+            /// <param name="norm_layer">Norm layer that will be stacked on top of the convolution layer. If ``null``, ``torch.nn.BatchNorm2d`` is used.</param>
+            /// <param name="activation_layer">Activation function which will be stacked on top of the normalization layer (if not None), otherwise on top of the conv layer. If ``null``, ``torch.nn.ReLU`` is used.</param>
+            /// <param name="dilation">Spacing between kernel elements.</param>
+            /// <param name="inplace">Parameter for the activation layer, which can optionally do the operation in-place.</param>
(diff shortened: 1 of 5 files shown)


GeorgeS2019:
> FYI

*.pth, which TorchSharp does not support. This is for saving Pytorch modules via pickling.
*.pt, which TorchSharp will load and save, but not create from scratch. This is for saving Pytorch modules via TorchScript.
*.ts/.data/.whatever, which is for saving model weights only in the TorchSharp-specific format.

src/TorchSharp/TorchVision/Ops/Misc.cs
```diff
@@ -0,0 +1,145 @@
...
+            /// <param name="stride">Stride of the convolution.</param>
+            /// <param name="padding">Padding added to all four sides of the input. Default: null, in which case it will calculated as ``padding = (kernel_size - 1) // 2 * dilation``</param>
+            /// <param name="groups">Number of blocked connections from input channels to output channels.</param>
+            /// <param name="norm_layer">Norm layer that will be stacked on top of the convolution layer. If ``null``, ``torch.nn.BatchNorm2d`` is used.</param>
+            /// <param name="activation_layer">Activation function which will be stacked on top of the normalization layer (if not None), otherwise on top of the conv layer. If ``null``, ``torch.nn.ReLU`` is used.</param>
+            /// <param name="dilation">Spacing between kernel elements.</param>
+            /// <param name="inplace">Parameter for the activation layer, which can optionally do the operation in-place.</param>
+            /// <param name="bias">Whether to use bias in the convolution layer. By default, is false.</param>
+            public static nn.Module Conv2dNormActivation(
```
NiklasGustafsson, line 38: Wat about Conv1d?

src/TorchSharp/TorchVision/models/Utils.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        public static partial class models
+        {
+            internal static partial class _utils
+            {
+                /// <summary>
+                /// This function is taken from the original tf repo.
+                /// It ensures that all layers have a channel number that is divisible by 8
+                /// It can be seen here:
+                /// https://github.com/tensorflow/models/blob/master/research/slim/nets/mobilenet/mobilenet.py
```
NiklasGustafsson, line 19: Please add a reference to the TF license and copyright notice from here.

NiklasGustafsson (changes requested):

src/TorchSharp/TorchVision/Ops/Misc.cs
kaiidams: `Conv1dNormActivation` is not included in the original `torchvision`

src/TorchSharp/TorchVision/models/Utils.cs
kaiidams: Added references. Also fix wrong mentions to torchaudio with torchvision.

src/TorchSharp/TorchVision/Ops/Misc.cs
```diff
@@ -0,0 +1,145 @@
...
+                Func<bool, nn.Module>? activation_layer = null,
+                long dilation = 1,
+                bool inplace = true,
+                bool? bias = null)
+            {
+                if (padding == null) {
+                    padding = (kernel_size - 1) / 2 * dilation;
+                }
+
+                if (bias == null) {
```
NiklasGustafsson, line 55: Given that 'null' doesn't seem to mean anything special, why can't the parameter just be a 'bool' with 'false' as the default?

src/TorchSharp/TorchVision/Ops/Misc.cs
```diff
@@ -0,0 +1,145 @@
...
+                    bias = false;
```
NiklasGustafsson, line 116: Same comment as earlier.

NiklasGustafsson (changes requested):

src/TorchSharp/TorchVision/Ops/Misc.cs
kaiidams: I changed the code not to fall back to `torch.nn.BatchNorm2d` when `norm_layer` is `null`.

The original signature of torchvision is as below. It accepts the optional `norm_layer` with a default callable. In C# the default value must be `null` as it must be compile time static, but I wanted to emulate the behavior.

https://github.com/pytorch/vision/blob/841b9a19a5ae6fb92b517438061b9846153cc0c6/torchvision/ops/misc.py#L144

```
    def __init__(
        self,
        in_channels: int,
        out_channels: int,
        kernel_size: Union[int, Tuple[int, int]] = 3,
        stride: Union[int, Tuple[int, int]] = 1,
        padding: Optional[Union[int, Tuple[int, int], str]] = None,
        groups: int = 1,
        norm_layer: Optional[Callable[..., torch.nn.Module]] = torch.nn.BatchNorm2d,
        activation_layer: Optional[Callable[..., torch.nn.Module]] = torch.nn.ReLU,
        dilation: Union[int, Tuple[int, int]] = 1,
        inplace: Optional[bool] = True,
        bias: Optional[bool] = None,
    ) -> None:
```

src/TorchSharp/TorchVision/Ops/Misc.cs
kaiidams: Same as earlier

NiklasGustafsson (approved):

Issue #749, opened 2022-09-20
`functional.linear` is wrong


lostmsu:
It is supposed to have signature `Tensor linear(Tensor input, Tensor weight, Tensor? bias = null)` and perform corresponding action.

NiklasGustafsson:
Thank you for catching that -- should be easy to fix.

PR #751, opened 2022-09-20
Address issue #749 -- incorrect nn.functional.linear function signature.

NiklasGustafsson:
Also changed bilinear to use native function directly.


tarekgh (dismissed):
LGTM!

PR #753, opened 2022-09-21
Adding device and dtype to a number of module factories.

NiklasGustafsson:
Renamed a few parameters to align with PyTorch.
Added parameters and buffers to InstanceNorm{123}d.

### DIFF
--- a/src/TorchSharp/JIT/ScriptModule.cs
+++ b/src/TorchSharp/JIT/ScriptModule.cs
@@ -169,7 +169,7 @@ namespace TorchSharp
                 [DllImport("LibTorchSharp")]
                 static extern void THSJIT_Module_to_dtype(HType module, sbyte dtype);
 
-                public override nn.Module to(Device device, ScalarType dtype)
+                internal protected override nn.Module _to(Device device, ScalarType dtype)
                 {
                     if (device.type != DeviceType.CUDA) { device = new Device(device.type, -1); };
 
@@ -191,7 +191,7 @@ namespace TorchSharp
                 /// <param name="deviceType">The device type, e.g. 'CPU' or 'CUDA'.</param>
                 /// <param name="deviceIndex">The optional device index.</param>
                 /// <returns></returns>
-                public override nn.Module to(DeviceType deviceType, int deviceIndex = -1)
+                internal protected override nn.Module _to(DeviceType deviceType, int deviceIndex = -1)
                 {
                     if (deviceType != DeviceType.CUDA) deviceIndex = -1;
 
@@ -218,7 +218,7 @@ namespace TorchSharp
                 /// Convert the parameters and buffers.
                 /// </summary>
                 /// <returns></returns>
-                public override nn.Module to(ScalarType dtype)
+                internal protected override nn.Module _to(ScalarType dtype)
                 {
                     THSJIT_Module_to_dtype(handle, (sbyte)dtype);
                     CheckForErrors();
--- a/src/TorchSharp/JIT/ScriptModule.cs
+++ b/src/TorchSharp/JIT/ScriptModule.cs
@@ -175,47 +175,12 @@ namespace TorchSharp
 
                     if (device.type == DeviceType.CUDA && !torch.cuda.is_available()) throw new InvalidOperationException("CUDA is not available.");
 
-                    if (device.type != _deviceType || device.index != _deviceIndex) {
 
-                        InitializeDeviceType(device.type);
-                        THSJIT_Module_to_device_dtype(handle, (sbyte)dtype, (int)device.type, device.index);
-                        CheckForErrors();
+                    InitializeDeviceType(device.type);
+                    THSJIT_Module_to_device_dtype(handle, (sbyte)dtype, (int)device.type, device.index);
+                    CheckForErrors();
 
-                        foreach (var (_, sm) in named_children()) sm.to(device.type, device.index);
-
-                        foreach (var field in GetType().GetFields(BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Instance)) {
-
-                            var fieldName = field.Name;
-                            var value = field.GetValue(this);
-
-                            switch (value) {
-                            // This test must come before the Tensor test
-                            case Modules.Parameter param when dtype == param.dtype && device.type == param.device_type && device.index == param.device_index:
-                                continue;
-
-                            case Modules.Parameter param: {
-                                    var t = param.to(dtype, device);
-                                    t.retain_grad();
-                                    var p = new Modules.Parameter(t, param.requires_grad);
-                                    field.SetValue(this, p);
-                                    ConditionallyRegisterParameter(fieldName, p);
-                                    break;
-                                }
-
-                            case Tensor tensor when (device.type != tensor.device_type || device.index != tensor.device_index): {
-                                    var t = tensor.to(dtype, device);
-                                    field.SetValue(this, t);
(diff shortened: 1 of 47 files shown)


NiklasGustafsson:
@tarekgh -- this is massive, but most of the changes follow exactly the same pattern. The most important changes to review are in Module.cs and in ScriptModule.cs, the rest are pretty templated.

@lostmsu -- if you could give it a quick check, that would be great, but don't feel obligated to. It's the "ease of use" aspect of the fix to the issue you reported, not the rewrite of the modules in C#.

src/TorchSharp/JIT/ScriptModule.cs
```diff
@@ -163,9 +163,63 @@ public override void eval()
...
+                            var value = field.GetValue(this);
+
+                            switch (value) {
+                            // This test must come before the Tensor test
+                            case Modules.Parameter param when dtype == param.dtype && device.type == param.device_type && device.index == param.device_index:
+                                continue;
+
+                            case Modules.Parameter param: {
+                                    var t = param.to(dtype, device);
+                                    t.retain_grad();
```
lostmsu, line 198: Why is this needed? Should it only be called when the original parameter `requires_grad`?

src/TorchSharp/JIT/ScriptModule.cs
```diff
@@ -163,9 +163,63 @@ public override void eval()
...
+                                    var t = tensor.to(dtype, device);
+                                    field.SetValue(this, t);
+                                    ConditionallyRegisterBuffer(fieldName, t);
+                                    break;
+                                }
+                            }
+                        }
+
+                        _deviceType = device.type;
+                        _deviceIndex = device.index;
```
lostmsu, line 215: Not related to this change, but is having device info cached in the object safe? Can I obtain two C# objects pointing to the same module? If so, and I move the module using one object, the other's state becomes inconsistent?

src/TorchSharp/NN/Module.cs
```diff
@@ -142,12 +143,66 @@ protected virtual void Dispose(bool disposing)
...
+                [DllImport("LibTorchSharp")]
+                static extern void THSNN_Module_to_device_dtype(HType module, sbyte dtype, long deviceType, long deviceIndex);
+
                 [DllImport("LibTorchSharp")]
                 static extern void THSNN_Module_to_device(HType module, long deviceType, long deviceIndex);
 
                 [DllImport("LibTorchSharp")]
                 static extern void THSNN_Module_to_dtype(HType module, sbyte dtype);
 
+                public virtual Module to(Device device, ScalarType dtype)
```
lostmsu, line 155: Haven't I seen this exact method body before. Can it be reused?

src/TorchSharp/NN/Module.cs
```diff
@@ -980,6 +1035,30 @@ protected virtual void RegisterComponents()
...
+                    if (!dtype.HasValue)
+                        dtype = get_default_dtype();
+
+                    if (device == null)
+                        device = torch.CPU;
+
+                    return (device, dtype.Value);
+                }
+
+                internal T MoveModule<T>(Device device, ScalarType? dtype) where T : Module
```
lostmsu, line 1049: Make this a static extension method, then C# will infer the return type.

test/TorchSharpTest/TestTorchTensorBugs.cs
```diff
@@ -531,8 +531,8 @@ public Module510(int in_channels, int out_channels, int kernel_size = 3, int str
                     ReLU(inplace: true)
                 );
 
-                temp.weight = Parameter(torch.randn(temp.weight.shape));
-                temp.bias = Parameter(torch.randn(temp.bias.shape));
+                temp.weight = Parameter(torch.randn(temp.weight!.shape));
```
lostmsu, line 534: Can you actually null `weight` of `BatchNorm1d`? Is it a legal operation that does not break things?

NiklasGustafsson: PyTorch allows it. Not sure what happens after...

src/TorchSharp/NN/Module.cs
NiklasGustafsson: The ScriptModule version is an override that calls to a different native method than non-script modules do.

src/TorchSharp/JIT/ScriptModule.cs
NiklasGustafsson: That needs to be investigated. As you say, not related to this change, specifically.

src/TorchSharp/NN/Module.cs
NiklasGustafsson: I'll take a look at that.

src/TorchSharp/JIT/ScriptModule.cs
NiklasGustafsson: Parameters should always require gradients.

src/TorchSharp/NN/Module.cs
NiklasGustafsson: ... but there are still opportunities for sharing. Thanks for pointing that out.

tarekgh: Right, I think there is no need to override this method in SxriptModule.

src/TorchSharp/NN/Module.cs
```diff
@@ -142,12 +143,74 @@ protected virtual void Dispose(bool disposing)
...
+                                field.SetValue(this, p);
+                                ConditionallyRegisterParameter(fieldName, p);
+                                break;
+                            }
+
+                        case Tensor tensor when (device.type != tensor.device_type || device.index != tensor.device_index): {
+                                var t = tensor.to(dtype, device);
+                                field.SetValue(this, t);
+                                ConditionallyRegisterBuffer(fieldName, t);
+                                break;
```
tarekgh, line 202: nit: please align this block.

src/TorchSharp/NN/Module.cs
```diff
@@ -142,12 +143,74 @@ protected virtual void Dispose(bool disposing)
...
+                    InitializeDeviceType(device.type);
+                    THSNN_Module_to_device_dtype(handle, (sbyte)dtype, (int)device.type, device.index);
+                    CheckForErrors();
+
+                    _toEpilog(device, dtype);
+
+                    return this;
+                }
+
+                protected void _toEpilog(Device device, ScalarType dtype)
```
tarekgh, line 175: >_toEpilog [](http://example.com/codeflow?start=31&length=9)

The code here is identical to the code in the same method that takes DeviceType. Is there a way we can share the code instead of duplicating it? Or we can make this method a generic method?

src/TorchSharp/NN/Module.cs
```diff
@@ -980,6 +1053,30 @@ protected virtual void RegisterComponents()
...
+                internal T MoveModule<T>(Device device, ScalarType? dtype) where T : Module
+                {
+                    T module = (T)this;
+                    if (device != null && dtype.HasValue) {
+                        module = (T)module.to(device, dtype.Value);
+                    } else if (device != null) {
+                        module = (T)module.to(device);
+                    } else if (dtype.HasValue) {
+                        module = (T)module.to(dtype.Value);
+                    }
```
tarekgh, line 1076: nit:

```C#
return device != null ? 
           (dtype.HasValue ? (T)module.to(device, dtype.Value) : (T)module.to(device)) :
           (dtype.HasValue ? (T)module.to(dtype.Value) : module);
````

tarekgh (approved):

PR #754, opened 2022-09-21
Add `torch.nn.Hardswish()` and `torch.nn.functional.hardswish()`

kaiidams:
- Add torch.nn.Hardswish() and torch.nn.functional.hardswish()
- `torch.nn.functional.glu(): fix type of `dim`.
- `torch.nn.functional.hardsigmoid(): Call Tensor.hardsigmoid directly.
- `torch.nn.functional.hardsigmoid(): Uncapitalize function name
- `torch.nn.functional.leaky_relu(): Wrong default value of negative_slope. https://pytorch.org/docs/stable/generated/torch.nn.functional.leaky_relu.html#torch.nn.functional.leaky_relu
- `ChannelShuffle.cs` moved as it is not an activation function.

### DIFF
--- a/src/TorchSharp/NN/Activation/Hardswish.cs
+++ b/src/TorchSharp/NN/Activation/Hardswish.cs
@@ -1,0 +1,68 @@
+// Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
+
+using static TorchSharp.torch;
+
+namespace TorchSharp
+{
+    using Modules;
+
+    namespace Modules
+    {
+        /// <summary>
+        /// This class is used to represent a Hardswish module.
+        /// </summary>
+        public class Hardswish : torch.nn.Module
+        {
+            private readonly bool inplace;
+
+            internal Hardswish(bool inplace = false) : base(nameof(Hardswish))
+            {
+                this.inplace = inplace;
+            }
+
+            public override Tensor forward(Tensor tensor)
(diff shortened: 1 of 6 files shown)


src/TorchSharp/NN/Activation/Hardswish.cs
```diff
@@ -0,0 +1,68 @@
...
+                /// Applies the Hardswish function, element-wise, as described in the paper:
+                /// `Searching for MobileNetV3 https://arxiv.org/abs/1905.02244`.
+                /// </summary>
+                /// <param name="input">The input tensor</param>
+                /// <param name="inplace">Do the operation in-place</param>
+                /// <returns></returns>
+                static public Tensor hardswish(Tensor input, bool inplace = false)
+                {
+                    using (var m = nn.Hardswish(inplace)) {
+                        return m.forward(input);
```
NiklasGustafsson, line 62: Maybe this can just invoke Tensor.hardswish instead of creating a module?

NiklasGustafsson (changes requested):
Use Tensor.hardswish in the nn.functional version.

src/TorchSharp/NN/Activation/Hardswish.cs
kaiidams: Thanks. Fixed `hardsigmoid()` and `hardswish()`

NiklasGustafsson (approved):

PR #757, opened 2022-09-22
Draft sketching a solution to the forward() problem.

NiklasGustafsson:
Here's a sketch of what it would look like to have a general `object forward(object)` and how it would work with Sequential's need to pass arbitrary data from one layer to the next.

I have modified MaxPool1d in the way that I would expect modules like it to behave, but not systematically gone through everything to look for others that require the same treatment. Also, I have not sketched out how this would affect ScriptModule.


kaiidams:
PyTorch `Sequential` doesn't support modules with multiple arguments. I think it is out of scope of TorchSharp when PyTorch doesn't do it. Probably DiffSharp can provide cleaner module-chaining with static typing. I think the current issue of `forward()` is that you have to know the actual class to call the right method. 

```cs
    public class TestModule1 : nn.Module
    {
        public Tensor forward(Tensor x, Tensor? length = null)
        {
            return length is null ? x : x + length; // Do something
        }
    }

    public class TestModule2 : nn.Module
    {
        public new (Tensor, Tensor) forward(Tensor x)
        {
            return (x + 1, x - 1);
        }
    }

    // These call the wrong `forward()` method.
    nn.Module m1 = new TestModule1();
    m1.forward(x);
    nn.Module m2 = new TestModule2();
    m1.forward(x);
```

This is issue for `Sequential` which only know about `nn.Module`. Probably this can be resolved by using `dynamic`.

```cs
    public class Sequential : nn.Module
    {
        List<nn.Module> layers;
        public Tensor forward(Tensor input)
        {
            foreach (dynamic layer in layers)
            {
                input = (Tensor)layer.forward(input);
            }
            return input;
        }
    }
```

`MaxPool1d` has another issue that it may return values of different types from a single function. I personally prefer to avoid this issue by having two classes `MaxPool1d` and `MaxPool1dWithIndices` or two methods like we do now.

`ScriptModule` has also another issue. It can have methods which names are not known statically.

```python
from typing import Tuple
import torch
from torch import nn

class MyModule(nn.Module):
	def __init__(self):
		super().__init__()
		self.p = nn.Parameter(torch.rand(10))

	def forward(self, x: torch.Tensor, y: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
		return x + y, x - y

	@torch.jit.export
	def predict(self, x: torch.Tensor) -> torch.Tensor:
		return x + self.p


m = MyModule()

m = torch.jit.script(m)
m.save("test.pt")
```

NiklasGustafsson:
@kaiidams, I added a new issue for supporting method invocation for ScriptModule. It will probably require that we add the use of 'dynamic'

PR #764, opened 2022-09-29
Support saving and restoring optimizer state.

NiklasGustafsson:
As mentioned in #761

### DIFF
--- a/src/TorchSharp/Optimizers/ASGD.cs
+++ b/src/TorchSharp/Optimizers/ASGD.cs
@@ -1,13 +1,12 @@
 // Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
 using System;
+using System.IO;
 using System.Collections.Generic;
 using System.Linq;
 using static TorchSharp.torch;
 
 namespace TorchSharp
 {
-    using System.IO;
-    using Microsoft.VisualBasic;
     using Modules;
 
     public static partial class torch
--- a/src/TorchSharp/Optimizers/ASGD.cs
+++ b/src/TorchSharp/Optimizers/ASGD.cs
@@ -271,7 +271,7 @@ namespace TorchSharp
                     state.step = 0;
                     state.eta = param_group.LearningRate;
                     state.mu = 1;
-                    state.ax = torch.zeros_like(p);
+                    state.ax = torch.zeros_like(p).DetatchFromDisposeScope();
                 }
             }
 
--- a/src/TorchSharp/Optimizers/ASGD.cs
+++ b/src/TorchSharp/Optimizers/ASGD.cs
@@ -202,6 +202,10 @@ namespace TorchSharp
                     ax.to(device);
                 }
 
+                /// <summary>
+                /// Load the optimizer parameter state from a stream.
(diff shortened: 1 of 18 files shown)


src/TorchSharp/Optimizers/Adagrad.cs
```diff
@@ -189,6 +191,58 @@ public void Dispose()
...
+                }
+
+                /// <summary>
+                /// Useful for tests, allows comparison of one state with another.
+                /// </summary>
+                /// <param name="other">The other optimizer state</param>
+                /// <returns></returns>
+                public override bool ApproximatelyEquals(OptimizerState other)
+                {
+                    var rhs = other as State;
```
tarekgh, line 243: nit: this can be written as 

```C#
    return (other is State rhs) && step == rhs.step && sum.allclose(rhs.sum);
```
if you like this style, can be applied to other Equals methods in the rest of the classes.

src/TorchSharp/Optimizers/ASGD.cs
```diff
@@ -6,6 +6,8 @@
 
 namespace TorchSharp
 {
+    using System.IO;
+    using Microsoft.VisualBasic;
```
tarekgh, line 10: >Microsoft.VisualBasic [](http://example.com/codeflow?start=10&length=21)

this is interesting :-)

src/TorchSharp/Optimizers/RMSprop.cs
```diff
@@ -212,6 +213,71 @@ public void Dispose()
...
+                /// <returns></returns>
+                public override bool ApproximatelyEquals(OptimizerState other)
+                {
+                    var rhs = other as State;
+                    return (rhs is not null) && step == rhs.step &&
+                        square_avg.allclose(rhs.square_avg) &&
+                        grad_avg.allclose(rhs.grad_avg) &&
+                        momentum_buffer.allclose(rhs.momentum_buffer);
+                }
             }
```
tarekgh, line 283: nit: remove extra spaced lines.

src/TorchSharp/Optimizers/Rprop.cs
```diff
@@ -3,9 +3,15 @@
...
 using static TorchSharp.torch;
+using static TorchSharp.torchvision;
+using static TorchSharp.Utils.LEB128Codec;
 
 namespace TorchSharp
 {
+    using System.IO;
+    using System.Net;
+    using Google.Protobuf.WellKnownTypes;
+    using ICSharpCode.SharpZipLib.Core;
```
tarekgh, line 14: why we don't add these to the top?

tarekgh (dismissed):

src/TorchSharp/Optimizers/Rprop.cs
NiklasGustafsson: Something VS added. No idea why not at the top. I'll fix.

src/TorchSharp/Optimizers/ASGD.cs
NiklasGustafsson: Must be something VS added automatically. Seems to be doing that every now and then. Kinda funny, I think.

PR #767, opened 2022-09-30
Align loss functions with PyTorch's APIs.

NiklasGustafsson:
The most immediate consequence is that losses are modules rather than delegates, which means you need to call `.forward()` to actually compute the loss. Also, the factories are in `torch.nn` rather than `torch.nn.functional` and have the same Pascal-casing as the corresponding types.

The members of the `torch.nn.functional` static class are now proper immediate loss functions, whereas the previous ones returned a loss delegate.

### DIFF
--- a/test/TorchSharpTest/NN.cs
+++ b/test/TorchSharpTest/NN.cs
@@ -677,8 +677,8 @@ namespace TorchSharp
             var y = torch.randn(new long[] { 64, 10 });
 
             var eval = seq.forward(x);
-            var loss = mse_loss(Reduction.Sum);
-            var output = loss(eval, y);
+            var loss = MSELoss(Reduction.Sum);
+            var output = loss.forward(eval, y);
 
             var result = output.ToSingle();
         }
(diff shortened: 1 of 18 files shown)


NiklasGustafsson:
@kaiidams -- here's the loss function fix.

tarekgh (approved):

kaiidams (dismissed):

test/TorchSharpTest/NN.cs
```diff
@@ -1048,6 +1234,23 @@ public void TestTripleMarginWithDistanceLoss()
...
+                (x, y) => {
+                    return (x - y).abs();
+                };
+
+            using (Tensor anchor = torch.rand(new long[] { 15, 5 }, requiresGrad: true).neg())
+            using (Tensor positive = torch.randn(new long[] { 15, 5 }, requiresGrad: true))
+            using (Tensor negative = torch.randn(new long[] { 15, 5 })) {
+
+                var result = triplet_margin_with_distance_loss(anchor, positive, negative, distance);
+                Assert.True(true);
```
kaiidams, line 1250: Probably we could check `Assert.Equal(Array.Empty<long>(), result.shape);` or `Assert.False(float.IsNaN(result.item<float>())` to make sure `result` is a float number.

kaiidams (approved):

test/TorchSharpTest/NN.cs
NiklasGustafsson: Excellent idea.

Issue #323, opened 2021-08-11
forward() should take a variable-length list of arguments


NiklasGustafsson:
The pytorch Module.forward() function is defined to take a variable-length list of argument, while the TorchSharp implementation more or less assumes that you can only forward one Tensor "automatically" in a construct like Sequential.

NiklasGustafsson:
Here's a design question I have. The most general solution would be to have a forward() that is declared as such:

```C#
public virtual object[] forward(object[])
```

but that would mean that there will be a lot of type casting, array indexing, and array creation in the common scenario, which is one tensor in, one tensor out.

We can easily overload the forward() function to take different kinds of arguments and return 'object', but that would complicate the implementation of Sequential, which won't statically know what is being passed from one layer to the next (in the current implementation it assumes one in, one out). A more complex implementation of Sequential would have to inspect the output of one layer and decide which overloaded version of forward() to invoke.

I'm not sure I like the options here, so I would be happy to get some thoughts on this.

dsyme:
@NiklasGustafsson There's not really good solution.  DiffSharp has a generic `Model<TIn, TOut>` and we could do that but it doesn't really solve that much.  Having one virtual forward with the signature you show seems reasonable in the nature of things, pushing the complexity to Sequential.

NiklasGustafsson:
@kaiidams, @MovGP0, @lostmsu -- looking for your input on this old issue. Most models fall into a few patterns for what forward() takes and returns, but the current implementation of Sequential expects it to be forward: Tensor -> Tensor. It would take a little bit of complexity and overhead to generalize fully, but...

NiklasGustafsson:
We could, for example, make the only version of forward 'Tensor[] forward(params Tensor[] arguments);` and then have virtual methods `_forward(...)` for the common cases, which built-in modules override. The general version is then what is used by Sequential, but it will require a lot of allocations of arrays for the return values, which is a waste of time and space.

lostmsu:
How about this option: simply add this method to `Module`:

```csharp
public virtual (Tensor, T) forward<T>(Tensor x, T extras)
  => (this.forward(x), extras);
```

lostmsu:
I realized the previous suggestion does not cover scenarios with `T` being different for input and output. To cover that maybe something like

```csharp
public virtual (Tensor, TOut) forward<TIn, TOut>(Tensor x, T extras)
  => (this.forward(x), this.forward_extras<TIn, TOut>(extras));
[...]

PR #768, opened 2022-10-01
WIP: Yet another draft solution for the forward() problem.

kaiidams:
https://github.com/dotnet/TorchSharp/issues/323#issuecomment-927715929

> @NiklasGustafsson There's not really good solution. DiffSharp has a generic Model<TIn, TOut> and we could do that but it doesn't really solve that much. Having one virtual forward with the signature you show seems reasonable in the nature of things, pushing the complexity to Sequential.

This implements generic `nn.Model<T1, ..., TResult>` that is derived from `nn.Module`. `nn.Module` doesn't declare `forward()` functions, but generic subclasses do.

https://github.com/dotnet/TorchSharp/blob/51f37e11f7a7d0ddf5e53bad67281027c0ed7690/src/TorchSharp/NN/Module.cs#L1115
```cs
            public abstract class Module<T1, TResult> : Module
            {
                protected Module(string name) : base(name) { }
                protected Module(IntPtr handle, IntPtr boxedHandle) : base(handle, boxedHandle) { }
                internal Module(HType handle, IntPtr? boxedHandle) : base(handle, boxedHandle) { }
                public abstract TResult forward(T1 input1);
            }
```

This can avoid confusing `forward()` functions with `new` and `override`, with explicit type casting. A module may implement both `forward(Tensor)` and `forward(Tensor, Tensor)`, however, it can be derived only from one of `nn.Model<Tensor, Tensor>` and `nn.Model<Tensor, Tensor, Tensor>`.

### DIFF
--- a/src/Examples/AlexNet.cs
+++ b/src/Examples/AlexNet.cs
@@ -8,11 +8,11 @@ namespace TorchSharp.Examples
     /// <summary>
     /// Modified version of original AlexNet to fix CIFAR10 32x32 images.
     /// </summary>
-    class AlexNet : Module
+    class AlexNet : Module<Tensor, Tensor>
     {
-        private readonly Module features;
-        private readonly Module avgPool;
-        private readonly Module classifier;
+        private readonly Module<Tensor, Tensor> features;
+        private readonly Module<Tensor, Tensor> avgPool;
+        private readonly Module<Tensor, Tensor> classifier;
 
         public AlexNet(string name, int numClasses, torch.Device device = null) : base(name)
         {
--- a/src/Examples/CIFAR10.cs
+++ b/src/Examples/CIFAR10.cs
@@ -56,7 +56,7 @@ namespace TorchSharp.Examples
 
             Console.WriteLine($"\tCreating the model...");
 
-            Module model = null;
+            Module<torch.Tensor, torch.Tensor> model = null;
 
             switch (modelName.ToLower()) {
             case "alexnet":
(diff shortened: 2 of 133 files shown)


NiklasGustafsson:
Wow, that's a lot of editing. I'll take a look at your changes. Not sure I'll finish over the weekend, though.

NiklasGustafsson:
@kaiidams -- I like this version. 

There are some modules, I think, that could possibly have more than one forward(). If the composition is based on IModule<T,TResult>, instead of injecting an abstract type in the middle, a module could potentially implement several versions of forward. Not sure I like the complication of that.

Another thing -- the PyTorch version of Sequential doesn't care if what a module returns, as long as the next module in the chain accepts the same kind of data. That won't be possible without an untyped forward().

And the MaxPool problem isn't addressed, but maybe that's not a problem we need to worry about.

NiklasGustafsson:
And ScriptModule isn't necessarily a Module<Tensor,Tensor>.

lostmsu:
+1 for interfaces

NiklasGustafsson:
> +1 for interfaces

@kaiidams -- I'm working on another sketch, with your proposal as the baseline, but adding the interfaces. I'll post it later this weekend.

NiklasGustafsson:
With interfaces, I believe MaxPool could implement IModule<Tensor,Tensor> as well as IModule<Tensor,(Tensor,Tensor)>, if the latter interface is implemented explicitly.

PR #769, opened 2022-10-01
WIP: forward(), yet again.

NiklasGustafsson:
@kaiidams, @lostmsu -- here's another version. This adds interfaces to Katsuya's version. The interface is useful for Sequential composition. It also helps make ScriptModule more general.


src/TorchSharp/JIT/ScriptModule.cs
```diff
@@ -268,91 +269,102 @@ private Type GetType(Type type)
...
+
+
+                    switch (typeCode) {
+                    case 0:
+                        // Nothing.
+                        throw new NotImplementedException("ScriptModule.forward() returning something else than a tensor, a tuple of tensors, or list of tensors.");
+                    case 1:
+                        // Tensor
+                        return new Tensor(ptrArray[0]);
+                    case 2:
```
lostmsu, line 303: Can C++ modules return a tuple of 0 elements?

src/TorchSharp/NN/Activation/ELU.cs
```diff
@@ -12,12 +12,12 @@ namespace Modules
...
         /// This class is used to represent a ELU module.
         /// </summary>
-        public class ELU : torch.nn.Module
+        public class ELU : torch.nn.Module<Tensor, Tensor>
         {
             internal ELU(IntPtr handle, IntPtr boxedHandle) : base(handle, boxedHandle) { }
 
             [DllImport("LibTorchSharp")]
-            private static extern IntPtr THSNN_ELU_forward(torch.nn.Module.HType module, IntPtr tensor);
+            private static extern IntPtr THSNN_ELU_forward(torch.nn.Module<Tensor, Tensor>.HType module, IntPtr tensor);
```
lostmsu, line 20: Does every generic module type need its own `HType`?

src/TorchSharp/NN/Losses.cs
```diff
@@ -12,16 +12,56 @@ namespace TorchSharp
...
+    {
+        public WeightedLoss(Tensor? weight = null, torch.nn.Reduction reduction = nn.Reduction.Mean) : base(reduction)
+        {
+            this.weight = weight;
+        }
+
+        public Tensor? weight { get; }
+    }
+
+    public abstract class WeightedLoss<T1, T2, T3, TResult> : Loss<T1, T2, T3, TResult>
```
lostmsu, line 54: Weighted loss should imho also implement `ILoss<T, TResult>`, cause the weight is optional?

src/TorchSharp/NN/Losses.cs
```diff
@@ -12,16 +12,56 @@ namespace TorchSharp
 {
     using Modules;
 
-    public class Loss : nn.Module
+    public abstract class Loss<T1, T2, TResult> : nn.Module<T1, T2, TResult>
```
lostmsu, line 15: Are there any losses, that do not return `Tensor`? I am curious if it makes sense to have it as a generic parameter.

src/TorchSharp/NN/Module.cs
```diff
@@ -1120,6 +1111,59 @@ protected void Dispose(bool disposing)
                     }
                 }
             }
+
+            public interface IModule<T1, TResult>
```
lostmsu, line 1115: Nitpick which I am unsure if it makes sense: it is `Func<T,TResult>`, not `Func<T1, TResult>`.

src/TorchSharp/NN/Recurrent/GRU.cs
```diff
@@ -25,7 +25,7 @@ internal GRU(IntPtr handle, IntPtr boxedHandle, long hiddenSize, long numLayers,
             private bool _batch_first;
 
             [DllImport("LibTorchSharp")]
-            extern static IntPtr THSNN_GRU_forward(torch.nn.Module.HType module, IntPtr input, IntPtr h_0, out IntPtr h_n);
+            extern static IntPtr THSNN_GRU_forward(torch.nn.Module<Tensor, Tensor>.HType module, IntPtr input, IntPtr h_0, out IntPtr h_n);
```
lostmsu, line 28: Unrelated to this change: are there many name conflicts? Does it make sense for every TorchSharp file to have `using static torch;` and `using static torch.nn` to be concise? Maybe even as a `global using`.

src/TorchSharp/NN/Sequential.cs
```diff
@@ -16,30 +16,32 @@ namespace Modules
         /// <summary>
         /// This class is used to represent a Sequential module.
         /// </summary>
-        public class Sequential : torch.nn.Module
+        public class Sequential : torch.nn.Module<Tensor, Tensor>
```
lostmsu, line 19: As I mentioned in the discussion, it would make sense for a `Sequential<T>: Module<T, T>`

src/TorchSharp/NN/Activation/ELU.cs
NiklasGustafsson: No.

src/TorchSharp/NN/Losses.cs
NiklasGustafsson: Not sure what you mean -- implementing `ILoss<T, TResult>` would mean that there's a `TResult forward(T);` and none of the loss functions have one like that.

src/TorchSharp/NN/Losses.cs
NiklasGustafsson: You can imagine some that return "double" or something like that, even if there's none that currently return anything but a tensor.

src/TorchSharp/NN/Losses.cs
lostmsu: Sorry, I misread that one thinking it provides an optional parameter `weight` instead of having it on the class instance itself.

src/TorchSharp/NN/Recurrent/GRU.cs
NiklasGustafsson: As you said, it's unrelated. Maybe something to fix in the future.

src/TorchSharp/JIT/ScriptModule.cs
NiklasGustafsson: I don't think so. If we encounter that, we can revisit.

NiklasGustafsson:
Thank you @kaiidams for demonstrating the value of using generics, and thank you @lostmsu for reviewing the draft. It's ready for review now.

PR #784, opened 2022-10-08
Add `torchaudio.transforms.GriffinLim()`.

kaiidams:
These are braking changes
- Replace `torchaudio.ITransform` with `IModule<Tensor, Tensor>`.
- Remove `torchaudio.transforms.Compose()`.


src/Examples/SpeechCommands.cs
```diff
@@ -87,7 +87,7 @@ private static BatchItem Collate(IEnumerable<SpeechCommandsDatasetItem> items, t
             };
         }
 
-        internal static void TrainingLoop(string dataset, Device device, M5 model, ITransform transform, Dataset<SpeechCommandsDatasetItem> train_data, Dataset<SpeechCommandsDatasetItem> test_data)
```
NiklasGustafsson, line 90: Maybe, in the interest of avoiding a breaking change, ITransform could simply be a derivation from IModule<Tensor,Tensor>? Maybe we don't have enough audio users for it to matter, but it's worth considering.

kaiidams: As you said, I'm not aware of many audio users. Compared to generic `nn.Module` change, I guess this change causes less impact. I'll add ITransform back if you think needed.

Meta has their internal usage logging https://github.com/pytorch/vision/pull/4799#issuecomment-954921854 to estimate the impact. .NET SDK also has wider telemetry https://learn.microsoft.com/en-us/dotnet/core/tools/telemetry .

NiklasGustafsson: We have a similar situation in TorchVision, btw, so let's address both at the same time. I like fewer interfaces, but at the same time, having ITransform : IModule<Tensor,Tensor> may increase readability for pre-processing blocks... I don't know.

kaiidams:
```
commit 8ba1587760633abe776970089bddc5ff75ade827 (HEAD -> griffinlim2, origin/griffinlim2)
Author: Katsuya Iida <katsuya.iida@gmail.com>
Date:   Sun Oct 9 12:53:34 2022 +0900

    Add `torchaudio.ITransform` back
```

NiklasGustafsson (approved):

PR #789, opened 2022-10-09
Generic ModuleList & ModuleDict

NiklasGustafsson:


src/TorchSharp/NN/ModuleDict.cs
```diff
@@ -21,7 +21,7 @@ namespace Modules
         /// ModuleDict is an ordered dictionary that respects the order of insertion, and
         /// in update(), the order of the merged OrderedDict or another ModuleDict (the argument to update()).
         /// </summary>
-        public class ModuleDict : Module, IDictionary<string, Module>, IList<(string, Module)>
+        public class ModuleDict<T> : Module, IDictionary<string, T>, IList<(string, T)> where T : Module
```
kaiidams, line 24: Should we have `IModule` so that we can do this? 
```cs
Modules.ModuleDict<IModule<Tensor, Tensor, Tensor>> dict
```

QingtaoLi1: The purpose to introduce `IModule<>` is to separate `forward()` methods with different number of inputs. IMHO, if we introduce a base interface `IModule`, it should be an empty interface. And it should not be public, otherwise it may lead to a custom Module without any managed `forward()` method, which disobey the idea of `IModule`.

QingtaoLi1 (approved):

QingtaoLi1 (approved):

src/TorchSharp/NN/ModuleDict.cs
NiklasGustafsson: ModuleList and ModuleDict are primarily used to trigger automatic registration of the modules that are held by them, i.e. during module construction. Registration is useful for the implementation of of named_parameters, named_buffers, and named_children. IModule would have to contain those members, which wouldn't make sense, IMO.

lostmsu: Why not? Actually makes sense for `IModule` to have `named_parameters` etc

The way I see it they are the only actual difference between a module in PyTorch sense and a `Func<T1, ..., TResult>` in .NET.

lostmsu: In fact, I believe I have one or two PyTorch modules with placeholder `forward` implementation for the sole purpose of having a separate namespace for a group of parameters.

kaiidams: In the current implementation, `Module` implements model structures and parameters. `IModule<>`s declares `forward()` function prototypes. A module must be a subclass of `Module` and implements some of `IModule<>`s. `Module<>` (without `I`), I think, is a utility abstract class which can be used when the module implements only one `IModule<>`. As managing model structures and parameters are complicated, it makes sense for `IModule<>` to have only `forward()`. If the purpose of `ModuleDict` is managing model structures and parameters, then `ModuleDict` should use `Module` or subclasses of `Module` as Niklas says. The issue is that it is not sure if a model is a subclass of `Module<>` when it implements more than one `IModule<>`s.

QingtaoLi1: @lostmsu Could you share the link of the PyTorch modules with placeholder `forward` mentioned above? I agree that `IModule` can have `named_parameters` etc., but the problem now is about the `forward`s. I'm using a `Dictionary<string, object>` to pass in multiple method arguments, which I think is not elegant.

QingtaoLi1: @kaiidams `Module`s with more than one kind of `forward()` can be unified by using `?` operator, e.g. `Module<Tensor, Tensor?, Tensor?>` and handled in the method logic. I'm not sure in which case a model can have multiple `forward()` methods. Can you provide an example of this? If you mean forwarding part of the model to get some intermediate representations, I think `forward` is not a proper name for this functionality.

kaiidams: This one implements multiple `IModule<>`s.

https://github.com/dotnet/TorchSharp/blob/97fa3df7fc62706bd19c332273d07f8f29eb7b7f/src/TorchSharp/NN/TransformerEncoderLayer.cs#L13

`Module<Tensor, Tensor?, Tensor?>` doesn't have default arguments, still you need to call it as `m.forward(x, null)`, which `nn.Seqential` cannot handle.

Also, a module can return different return values.

```cs
public MyModule : IModule<Tensor, Tensor, (Tensor, Tensor)>, IModule<Tensor, Tensor> {
...
}
```

QingtaoLi1: `nn.Sequential` does matter.

For different return values, one common practice in PyTorch is returning a tuple or dict containing all extra return values.

kaiidams: > For different return values, one common practice in PyTorch is returning a tuple or dict containing all extra return values.

That's true. These returns different types
- https://pytorch.org/docs/stable/generated/torch.nn.MaxPool2d.html
- https://pytorch.org/docs/stable/generated/torch.unique.html

NiklasGustafsson:
So, it sounds like there's no real consensus on this. 

IModule<T> is meant to represent the static signature of the forward() function(s) of a module. Maybe it's misnamed? There are other types with 'forward(),' like image and audio transforms, which are not necessarily modules in the nn namespace.

kaiidams:
Please go ahead. The review is already accepted.

> There are other types with 'forward(),' like image and audio transforms, which are not necessarily modules in the nn namespace.

 PyTorch `torchaudio`'s transforms are subclasses of `nn.Module`s. They are TorchScript-able.
Most of PyTorch `torchvision`'s transforms are subclasses of `nn.Module`s and TorchScript-able. TorchSharp's `ITransform`s are not `nn.Module`.

https://pytorch.org/vision/stable/transforms.html#scriptable-transforms
> For any custom transformations to be used with torch.jit.script, they should be derived from torch.nn.Module.

PR #792, opened 2022-10-10
Add SkiaSharp-based default imager for torchvision.io.

NiklasGustafsson:


NiklasGustafsson:
@kaiidams -- time to take a quick look?

src/TorchSharp/TorchSharp.csproj
```diff
@@ -23,6 +23,8 @@
   <ItemGroup>
     <PackageReference Include="Google.Protobuf" Version="3.21.5" />
     <PackageReference Include="SharpZipLib" Version="1.3.3" />
+    <PackageReference Include="SkiaSharp" Version="2.88.3" />
+    <PackageReference Include="SkiaSharp.NativeAssets.Linux.NoDependencies" Version="2.88.3" />
```
lostmsu, line 27: IMHO creating a dependency from TorchSharp to Skia is a bad idea. There are pure .NET image libraries that could be used instead, including legacy `System.Drawing`, `ImageSharp`, and many more. Using Skia adds native library dependencies further restricting platforms where TorchSharp can run.

NiklasGustafsson: SkiaSharp has good platform coverage. It's only meant to be the default image reader/writer, and it's only being used for encode/decode, not for any processing. Anyone can extend with their preferred alternative, and there's already an example using ImageSharp.

kaiidams: @NiklasGustafsson Adding a dependency to Skia makes it more difficult to use. However, if we have separated packages like `TorchSharp.Vision` and `TorchSharp.Audio`, then I think adding a dependency to Skia it not bad. At the end, `TorchSharp.Vision` and `TorchSharp.Audio` have to depend on more I/O and utility libraries.

NiklasGustafsson: Hmm... maybe the time has come to separate out TorchVision and TorchAudio in to their own packages. There's not much in TorchText at this point...

GeorgeS2019: - `system.drawing` will be phased out in .NET7 due to windows OS specific and ncompatibility with other OS 
- `imagesharp` is a serious option to substitute `system.drawing`. Unless Microsoft plans to acquire the company behind imagesharp, skiasharp seems to be a reasonable choice for now.

GeorgeS2019: > maybe the time has come to separate out TorchVision and TorchAudio into their own packages.

I agree that it is time for TorchSharp to be compatible with PyTorch in terms of packages and file structure arrangements to ease the process of **identifying yet-to-implement gaps** and of porting more PyTorch codes to TorchSharp.

GeorgeS2019: > There's not much in TorchText at this point...

We need to continue this discussion.  

Tentatively, I see no reason why we need to adhere to TorchText but perhaps we could go around it and have a synergistic strategy of addressing unmet needs for deep NLP for .NET communities. This means a coordinated strategy is needed from TorchSharp, ML.NET, Onnx runtime, and extension.

src/TorchSharp/TorchVision/IO/Image.cs
```diff
@@ -1,3 +1,4 @@
+// Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
```
kaiidams, line 1: Some may want equivalents of `pil_to_tensor()` and `tensor_to_pil()` to convert to/from `SkImage` or `ImageSharp.Image` directly. Not necessarily included in this PR.

https://pytorch.org/vision/stable/generated/torchvision.transforms.functional.pil_to_tensor.html#torchvision.transforms.functional.pil_to_tensor

src/TorchSharp/TorchVision/IO/Image.cs
```diff
@@ -1,3 +1,4 @@
...
 using System.IO;
```
kaiidams, line 2: Many of datasets in `torchvision.datasets` return `PIL` image with `__getitem__()`. Which TorchSharp's datasets should return, `Tensor` or objects of `Imager`?

kaiidams (approved):

src/TorchSharp/TorchVision/IO/Image.cs
NiklasGustafsson: Tensor.

src/TorchSharp/TorchVision/IO/Image.cs
NiklasGustafsson: I really would prefer not to allow types from underlying image libraries to escape into the TorchSharp public APIs -- that creates a tight coupling that will be hard to walk away from later. 

If converting to images (Skia, ImageSharp, or others) is important, I would prefer to provide well-documented example code that shows how to do it. There's already an example for ImageSharp, and the SkiaSharp code was lifted from the example I had in place.

kaiidams: I thought you wanted to make SkiaSharp the first-class library by adding SkiaSharp dependency to TorchSharp(Vision). If TorchSharp API should be clean from SkiaSharp. I think the package should be divided into TorchSharp and TorchSharp.Skia instead.

kaiidams: I think that's Python way. They make all things included so that novice programmers can do many things out of box. I think C# libraries tend to have less dependencies, that enables wider platform support. That's one of its strong points.

src/TorchSharp/TorchSharp.csproj
NiklasGustafsson: Okay, so what I'll do it merge this, and then start a PR to separate TV and TA from the core TorchSharp package. The next release won't go out until the second PR comes in.

tarekgh, line 27: >SkiaSharp.NativeAssets.Linux.NoDependencies [](http://example.com/codeflow?start=31&length=43)

@NiklasGustafsson what about Windows and MacOS dependency? or we don't run there?

NiklasGustafsson: Those are actually included in the SkiaSharp package. Only the Linux natives are outside of it.

lostmsu: Even if separated, what is the advantage of having Skia vs ImageSharp?

tarekgh:
>         public enum ImageFormat

Should compare this list with what SkiaSharp support? 

[https://learn.microsoft.com/en-us/dotnet/api/skiasharp.skencodedimageformat?view=skiasharp-2.88](https://learn.microsoft.com/en-us/dotnet/api/skiasharp.skencodedimageformat?view=skiasharp-2.88)

---
Refers to: src/TorchSharp/TorchVision/IO/ImageFormat.cs:8 in d72ce6c. [](commit_id = d72ce6c37d2639bf2e76d724c649e6dabc30ddb0, deletion_comment = False)

src/TorchSharp/TorchVision/IO/SkiaSharpImager.cs
```diff
@@ -0,0 +1,179 @@
...
+
+                    if (outputGrayScale) {
+                        result = torchvision.transforms.functional.rgb_to_grayscale(result);
+                    }
+
+                    return result.MoveToOuterDisposeScope();
+                }
+
+
+                private SKEncodedImageFormat TranslateImageFormat(ImageFormat format)
```
tarekgh, line 151: >TranslateImageFormat [](http://example.com/codeflow?start=45&length=20)

I see you have this, so you may ignore my other comment on the ImageFormat.

NiklasGustafsson:
Thanks for the feedback, everyone! I'm going to close this PR now and come back with one that separates TorchVision and TorchAudio into their own packages, and then we can take up the discussion about image backend choices again.

NiklasGustafsson:
To break out TV into its own package, there's a problem with the Tensor.Tensor(IntPtr handle) ctor. We either need to start using friend assemblies (yuck!), make the constructor public (there's all kinds of issues with that), or introduce a derived class 'ImageTensor' in TV and then make the Tensor ctor protected. It al feels very hacky.

tarekgh:
> there's a problem with the Tensor.Tensor(IntPtr handle) ctor

What are the problems with such a constructor? Is it possible to expose some static method like `UnsafeCreateTensor(IntPtr)`? which can tell it is unsafe and at the same time allow the functionality?

NiklasGustafsson:
> > there's a problem with the Tensor.Tensor(IntPtr handle) ctor
> 
> What are the problems with such a constructor? 

The risk of passing something in that wasn't created the right way in native code.

>> Is it possible to expose some static method like `UnsafeCreateTensor(IntPtr)`? which can tell it is unsafe and at the same time allow the functionality?

That's an excellent idea -- make it available but highlight the risk.

NiklasGustafsson:
FYI -- @tarekgh, @lostmsu, @kaiidams, @GeorgeS2019 

Here's some context on the decision to go with SkiaSharp instead of ImageSharp as the __default__, "out of the box" image reader:

https://dotnetfoundation.org/blog/2022/10/20/imagesharpupdate 

We did not want to create a transitive dependency on ImageSharp with TorchSharp. Therefore, the options were to have no have no default image reader, to go with one that is offered under an OSI-compliant license like SkiaSharp, or to fork a version of ImageSharp from before the license change. We felt that having no reader would provide a poor experience, and that forking ImageSharp is not something we want to do as good citizens of the .NET OSS community at large. Therefore, we decided that SkiaSharp was the best option given the circumstances.

An unsupported ImageSharp-based image reader is available in one of the examples.

PR #801, opened 2022-10-15
Add `torchvision.datasets.CelebA()`

kaiidams:
- Added implementation of `torchvision.datasets.CelebA()` [PyTorch document](https://pytorch.org/vision/main/generated/torchvision.datasets.CelebA.html)
- `torchvision.datasets.CelebA()` downloads one huge zip [img_align_celeba.zip](https://drive.google.com/file/d/0B7EVK8r0v71pZjFTYXZWM3FlRnM/view?usp=share_link&resourcekey=0-dYn9z10tMJOBAkviAcfdyQ) and 5 small text files from GDrive. 
- `torchvision.datasets.CelebA()` of both TorchSharp and PyTorch cannot download `img_align_celeba.zip` because of quota. But you can download it with a web browser and place it in `root` directory before calling `torchvision.datasets.CelebA()`.

#### Sample code

```csharp
torchvision.io.DefaultImager = new torchvision.io.SkiaImager();
var transform = torchvision.transforms.Resize(80, 80);
var data = torchvision.datasets.CelebA("./data", download: false, transform: transform);
var loader = new torch.utils.data.DataLoader(data, 4);
int i = 0;
foreach (var batch in loader) {
    var item = batch;
    var input = item["input"];
    var target = item["attr"];
    for (int j = 0; j < 4; j++) {
        torchvision.io.write_image(input[j], $"data/test_{i}.png", torchvision.ImageFormat.Png);
        StringBuilder sb = new StringBuilder();
        Console.WriteLine("{0}", target[j].ToString(TensorStringStyle.Default));
    }
    ++i;
}
```


kaiidams:
@NiklasGustafsson This is WIP. Can you take a look how `torchvision.datasets` are implemented? `torchvision` and `torchaudio` have different transform mechanism. This change is only for `torchvision.datasets` but not for `torchaudio.datasets`

NiklasGustafsson:
I'll take a look at this later today.

NiklasGustafsson:
Errors in build:

```
/__w/1/s/src/TorchVision/Datasets/CelebA.cs(86,25): error CS0815: Cannot assign method group to an implicitly-typed variable [/__w/1/s/src/TorchVision/TorchVision.csproj]
/__w/1/s/src/TorchVision/Datasets/CelebA.cs(86,25): error CS0815: Cannot assign method group to an implicitly-typed variable [/__w/1/s/src/TorchVision/TorchVision.csproj]
```

kaiidams:
Fixed the error.
VS 2022 is okay with the code before change.

NiklasGustafsson:
>Can you take a look how `torchvision.datasets` are implemented?

Can you elaborate on what you would like to see?

kaiidams:
@NiklasGustafsson 

> Can you elaborate on what you would like to see?

I am trying to replace current `ITransform` with generic based interface like `IModule<>`. Current `ITransform` is bound to `IDictionary<string, Tensor>`, but some of `Dataset` may return any types if we try to emulate interface of torchvision and torchaudio. This change follows `torchvision.datasets.VisionDataset` that assumes any types. The bad side of this change is too many type parameters. I'd like to ask you if you have any opinions, ideas.

kaiidams:
`ITransform` assumes that all values are `Tensor`s. This doesn't work sometimes. For example, an image caption dataset like `torchvision.datasets.CocoDetection`, returns targets that includes caption strings. Without knowing tokenizer to use, we cannot convert these into `Tensor`s. Also, non `Tensor` values may be efficient for classification datasets as you don't have to make a scalar `Tensor` for each sample and stack them into a batch.

```
loading annotations into memory...
Done (t=0.14s)
creating index...
index created!
(<PIL.Image.Image image mode=RGB size=586x640 at 0x275CE5D5250>,
 [{'image_id': 139,
   'id': 372891,
   'caption': 'A woman stands in the dining area at the table.'},
  {'image_id': 139,
   'id': 376968,
   'caption': 'A room with chairs, a table, and a woman in it.'},
  {'image_id': 139,
   'id': 379917,
   'caption': 'A woman standing in a kitchen by a window'},
  {'image_id': 139,
   'id': 382074,
   'caption': 'A person standing at a table in a room.'},
  {'image_id': 139,
   'id': 384831,
   'caption': 'A living area with a television and a table'}])
```

NiklasGustafsson:
> @NiklasGustafsson
> 
> `ITransform` assumes that all values are `Tensor`s. This doesn't work sometimes. For example, an image caption dataset like `torchvision.datasets.CocoDetection`, returns targets that includes caption strings. Without knowing tokenizer to use, we cannot convert these into `Tensor`s. Also, non `Tensor` values may be efficient for classification datasets as you don't have to make a scalar `Tensor` for each sample and stack them into a batch.

Right, but do you have to formalize the interface in an interface? The purpose of ITransform was to formalize it for Compose.

kaiidams:
> Right, but do you have to formalize the API in an interface? The purpose of ITransform was to formalize it for Compose.

You're right. If we can use `dynamic` or reflection in `Compose`, then `Compose` can handle non-`Tensor` types. Do you think it is acceptable?

NiklasGustafsson:
@kaiidams -- this PR was a WIP PR five months ago. Are you still interested in pursuing it?

kaiidams:
@NiklasGustafsson 
Thanks, I'll try to make a change without generic transform. CelebA of PyTorch supports multiple target tensors, but this PR will only support one target tensor.

NiklasGustafsson:
Errors in latest builds:

```
/__w/1/s/src/TorchVision/dsets/Utils.cs(62,44): error CS1998: This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread. [/__w/1/s/src/TorchVision/TorchVision.csproj]
```

kaiidams:
@NiklasGustafsson 
This PR is ready for review. Please take a look in your convenient time. No need to hurry.

NiklasGustafsson:
> @NiklasGustafsson This PR is ready for review. Please take a look in your convenient time. No need to hurry.

I should be able to get to it this week.

src/TorchSharp/Hub.cs
```diff
@@ -24,7 +24,8 @@ public static void set_create_progress_bar_func(Func<bool, IProgressBar> func)
                 createProgressBarFunc = func;
             }
 
-            internal static IProgressBar _create_progress_bar(bool hidden)
+            // Used internally by TorchVision
+            public static IProgressBar _create_progress_bar(bool hidden)
```
NiklasGustafsson, line 28: We can't have a public method start with a '_' -- if this is public only because it's needed in TV, maybe it can be a) duplicated or b) moved to some 'internal' or 'implementation' namespace somewhere.

src/TorchVision/dsets/CelebA.cs
```diff
@@ -0,0 +1,276 @@
...
+#nullable enable
+namespace TorchSharp
+{
+    namespace Modules
+    {
+        internal sealed class CelebA : Dataset
+        {
+            private static readonly (string, string, string)[] file_list = new (string, string, string)[] {
+                // File ID                                      MD5 Hash                            Filename
+                ("0B7EVK8r0v71pZjFTYXZWM3FlRnM", "00d2c5bc6d35e252742224ab0c1e8fcb", "img_align_celeba.zip"),
```
NiklasGustafsson, line 30: Is there any chance that these hashes will change if the dataset is updated?

src/TorchVision/dsets/CelebA.cs
```diff
@@ -0,0 +1,276 @@
...
+                }
+                result["input"] = X;
+                if (this.target_transform is not null) {
+                    string t = this.target_type[0];
+                    result[t] = this.target_transform.call(result[t]);
+                }
+                return result;
+            }
+
+            private bool _check_integrity()
```
NiklasGustafsson, line 187: For non-public APIs, and APIs that are not copies of PyTorch APIs, we usually try to stay with .NET naming conventions. I'm not sure that we've done it consistently, but it's been the intent.

src/TorchVision/dsets/CelebA.cs
```diff
@@ -0,0 +1,276 @@
...
+            /// <param name="transform">A function/transform that  takes in an PIL image
+            /// and returns a transformed version.</param>
+            /// <param name="target_transform">A function/transform that takes in the
+            /// target and transforms it.</param>
+            /// <param name="download">If true, downloads the dataset from the internet and
+            /// puts it in root directory. If dataset is already downloaded, it is not
+            /// downloaded again.</param>
+            public static Dataset CelebA(
+                string root,
+                string split = "train",
```
NiklasGustafsson, line 253: Could this be an enum instead of a string? That's one area where we usually diverge from PyTorch.

src/TorchVision/dsets/Utils.cs
```diff
@@ -0,0 +1,313 @@
...
+#nullable enable
+namespace TorchSharp
+{
+    public static partial class torchvision
+    {
+        public static partial class datasets
+        {
+            public static partial class utils
+            {
+                private static async Task _save_response_content_async(
```
NiklasGustafsson, line 35: See previous comment on non-public APIs and naming conventions.

src/TorchVision/dsets/Utils.cs
```diff
@@ -0,0 +1,313 @@
...
+                    try {
+                        Task.Run(async () => {
+                            await download_file_from_google_drive_async(file_id, root, filename, md5);
+                        }).Wait();
+                    } catch (AggregateException ex) {
+                        throw ex.InnerException ?? ex;
+                    }
+                }
+
+                internal static async Task download_file_from_google_drive_async(
```
NiklasGustafsson, line 127: This sounds broadly useful, so should probably be in an IO namespace, have a .NET-convention name, and be public...

NiklasGustafsson (changes requested):

src/TorchVision/dsets/CelebA.cs
kaiidams: Fixed by 0b49104c07c7b9e3e37b13e267a3c9e41bee9ca1

src/TorchVision/dsets/CelebA.cs
kaiidams: That will be changed when the content is updated, but I didn't see any changes for recent years.

src/TorchVision/dsets/CelebA.cs
kaiidams: Thanks. `CelebA._check_integrity()` is a non-public API but a copy of a PyTorch API (https://github.com/pytorch/vision/blob/b78d98bb152ffb9c0c0f5365f59f475c70b1784e/torchvision/datasets/celeba.py#LL132C9-L132C25). Keeping the same name makes it easier to track the changes in the original PyTorch code. Should we still stay with .NET naming conventions in this case?

kaiidams: Fixed by 377a802da22b963dc248b8248b3244ed2e6deadb

src/TorchVision/dsets/Utils.cs
kaiidams: Fixed by 377a802da22b963dc248b8248b3244ed2e6deadb

src/TorchSharp/Hub.cs
kaiidams: Renamed to `CreateProgressBar()`. It cannot be duplicated unless all dependencies are also duplicated.

src/TorchVision/dsets/Utils.cs
kaiidams: > be in an IO namespace

Moved the functionality to `torchvision.io` static class. I guess you didn't mean .NET `TorchSharp.IO` namespace.

kaiidams:
@NiklasGustafsson Thank you for reviewing. Updated the code.

NiklasGustafsson (approved):

NiklasGustafsson:
@kaiidams -- it took a while, but it's finally here! It'll be in the next release.

Issue #746, opened 2022-09-19
Move to .NET 6.0


NiklasGustafsson:
.NET 5.0 is no longer supported, so we should probably move TorchSharp to assume 6.0 and newer.

PR #808, opened 2022-10-20
Adding a language version property to all .csproj files.

NiklasGustafsson:
As per @MovGP0's suggestion.


NiklasGustafsson:
This also addresses issue #746. Since .NET 5.0 is no longer supported, the .NET Core build is moving to .NET 6.0.

PR #823, opened 2022-11-02
Incorrect and missing package references in .nupkgproj files.

NiklasGustafsson:


pkg/TorchVision/TorchVision.nupkgproj
```diff
@@ -17,5 +17,7 @@
     <Content Include="..\common\NormalPackage.targets" Pack="true" PackagePath="buildTransitive\netstandard2.0\$(MSBuildProjectName).targets" />
     <Content Include="..\empty.txt" Pack="true" PackagePath="lib\netstandard2.0\_._" />
     <Content Include="$(RepoRoot)\LICENSE.txt" Pack="true" PackagePath="" />
+    <PackageReference Include="SkiaSharp" Version="2.88.3" />
+    <PackageReference Include="SkiaSharp.NativeAssets.Linux" Version="2.88.3" />
```
ericstj, line 21: Is there a reason we choose Linux rather than [all](https://www.nuget.org/packages?q=SkiaSharp.NativeAssets) or nothing?

ericstj: cc @tarekgh

NiklasGustafsson: The base SkiaSharp package contains MacOS and Windows assets already.

michaelgsharp (dismissed):

pkg/TorchVision/TorchVision.nupkgproj
tarekgh: @NiklasGustafsson is right. We do the same in ML. I would recommend using [SkiaSharp.NativeAssets.Linux.NoDependencies](https://www.nuget.org/packages/SkiaSharp.NativeAssets.Linux.NoDependencies) for Linux though. It is smaller and will not lose anything already using.

tarekgh, line 21: I want to ensure seeing my comment https://github.com/dotnet/TorchSharp/pull/823/files#r1012295720

tarekgh (dismissed):
Added minor comment. LGTM otherwise.

pkg/TorchVision/TorchVision.nupkgproj
NiklasGustafsson: Yes, thank you for pointing that out. That's what we should be using.

PR #850, opened 2022-11-17
Fixes #845

### LINKED ISSUE #845: registered buffers are being ignored in move model to device
https://github.com/dotnet/TorchSharp/blob/a360c6d141ad18572586b16f4bccae5e02387539/src/TorchSharp/NN/Module.cs#L166

The named_children() method does not include registered buffers. 

If the registered buffer is not also a 'field' of the module object then its not moved.

The buffer may be registered with a module (even though it is not a field) (e.g. TorchSharp.Fun library does that).

Please add the move for buffers that are not fields

NiklasGustafsson:
Fixes #845

### DIFF
--- a/src/TorchSharp/NN/Module.cs
+++ b/src/TorchSharp/NN/Module.cs
@@ -165,6 +165,8 @@ namespace TorchSharp
                 {
                     foreach (var (_, sm) in named_children()) sm._to(device, dtype);
 
+                    var alreadyHandled = new HashSet<IntPtr>();
+
                     foreach (var field in GetType().GetFields(BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Instance)) {
 
                         var fieldName = field.Name;
@@ -181,6 +183,7 @@ namespace TorchSharp
                                 var p = new Parameter(t, param.requires_grad);
                                 field.SetValue(this, p);
                                 ConditionallyRegisterParameter(fieldName, p);
+                                alreadyHandled.Add(p.handle);
                                 break;
                             }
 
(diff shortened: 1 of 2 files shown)


NiklasGustafsson:
@fwaris -- could you take a look at this and do a review?

src/TorchSharp/NN/Module.cs
```diff
@@ -181,18 +183,32 @@ protected void _toEpilog(Device device, ScalarType dtype)
...
                                 var t = tensor.to(dtype, device);
                                 field.SetValue(this, t);
                                 ConditionallyRegisterBuffer(fieldName, t);
+                                alreadyHandled.Add(t.handle);
                                 break;
                             }
                         }
                     }
 
+                    foreach (var (name, param) in named_parameters().ToList()) {
```
tarekgh, line 200: >ToList [](http://example.com/codeflow?start=69&length=6)

why calling ToList()? this apply to other changes in this file.

NiklasGustafsson: Because the call to "ConditionallyRegister" will modify the collection underlying `named_parameters(),` so I need a clone. It may be faster to call `.Where((n,p) => !alreadyHandled.Contains(p.handle)).ToList()` and get a smaller list out, but I'm always worried about using LINQ in that way, not sure about the performance implications.

tarekgh: ok, make sense then.

NiklasGustafsson: Would ToArray() be faster? I've always assumed ToList() is faster.

tarekgh: I didn't measure it myself, but it may be worth measuring it with your scenario. this thread https://stackoverflow.com/questions/1105990/is-it-better-to-call-tolist-or-toarray-in-linq-queries is interesting to read. At the end measuring can give concrete answer.

Issue #842, opened 2022-11-12
How to use register_forward_hook?


tondangerza:
Hi,
I use resnet50 to classify images and I want use grad-cam to create heatmap 
to identify a defect but it must use a register_forward_hook and I not found this function
 What should I do?

NiklasGustafsson:
Not supported at the moment. Now that an issue has been raised, it'll be on the backlog. If you want to, please implement and send us a PR! :-)

NiklasGustafsson:
@tondangerza -- do you also need the pre-and full backward hooks?

tondangerza:
> @tondangerza -- do you also need the pre-and full backward hooks?

Yes, could you possibly add backward hooks? 
because I don't want to use 2 systems (python & C#)

NiklasGustafsson:
@tondangerza -- I haven't forgotten about this, but still trying to figure out the best way to hook methods calls in .NET. It's certainly possible to add pre-and-post-hooks manually to all forward() methods of built-in modules, and then ask users to add them also to their own forwards(). There is a precedent for this -- calling `RegisterComponents()` in the constructor of all custom modules, but I'd prefer something more transparent.

Also, it is necessary to also find an approach that works for ScriptModules, i.e TorchScript.

Any and all suggestions toward a solution are welcome...

NiklasGustafsson:
Somebody suggested a pretty simple (but it would be a very breaking change) to make `forward()` non-virtual. 

Instead of overriding 'forward()' in custom modules, we would have a virtual or abstract method `ForwardImpl()` (or something like that) which would be where the logic of your module would be found.

So, in C# "pseudo-code":

```C#
public abstract class Module<T, TResult> : Module, IModule<T, TResult>
{
    protected Module(string name) : base(name) { }
    protected Module(IntPtr handle, IntPtr boxedHandle) : base(handle, boxedHandle) { }
    internal Module(HType handle, IntPtr? boxedHandle) : base(handle, boxedHandle) { }

    public TResult forward(T input1)
    {
        // call pre-hook if there is one.

        var res = ForwardImpl(input);

        // call post-hook if there is one.

        return res;
    }

    public abstract TResult forward_impl(T input1);
}
```

A custom module would override `forward_impl()` and call `forward()` on all sub-modules.

It's similar to how you defined `forward()` in PyTorch, but usually use `__call__()` to pass data to the module.

[...]

PR #860, opened 2022-12-02
Re-structure modules in order to allow forward hooks.

NiklasGustafsson:
Meant to address https://github.com/dotnet/TorchSharp/issues/842

With this breaking change, `forward()` is used to implement the forward logic of each module, while a new API, `call()` replaces its role when invoking the forward logic. This allows us to easily implement forward pre-and-post hooks.


NiklasGustafsson:
@kaiidams, @MovGP0, @lostmsu, @dayo05 -- your thoughts would be appreciated.

NiklasGustafsson:
Note: implementing backward hooks will be much more complicated.

src/TorchSharp/NN/Module.cs
```diff
@@ -1171,12 +1271,38 @@ public abstract class Module<T, TResult> : Module, IModule<T, TResult>
...
             /// <typeparam name="T2">The second argument type of the module's forward() function.</typeparam>
             /// <typeparam name="TResult">The return type of the module's forward() function.</typeparam>
-            public abstract class Module<T1, T2, TResult> : Module, IModule<T1, T2, TResult>
+            public abstract class Module<T1, T2, TResult> : HookableModule<Func<Module<T1, T2, TResult>, T1, T2, (T1, T2)?>, Func<Module<T1, T2, TResult>, T1, T2, TResult, TResult>>, IModule<T1, T2, TResult>
             {
                 protected Module(string name) : base(name) { }
                 protected Module(IntPtr handle, IntPtr boxedHandle) : base(handle, boxedHandle) { }
                 internal Module(HType handle, IntPtr? boxedHandle) : base(handle, boxedHandle) { }
-                public abstract TResult forward(T1 input1, T2 input2);
+                protected abstract TResult forward (T1 input1, T2 input2);
```
kaiidams, line 1279: A space in `forward (`.

src/TorchSharp/NN/Module.cs
```diff
@@ -1148,21 +1149,120 @@ public interface IModule<T1, T2, T3, TResult>
...
             /// <typeparam name="T">The argument type of the module's forward() function.</typeparam>
             /// <typeparam name="TResult">The return type of the module's forward() function.</typeparam>
-            public abstract class Module<T, TResult> : Module, IModule<T, TResult>
+            public abstract class Module<T, TResult> : HookableModule<Func<Module<T,TResult>, T, T>, Func<Module<T, TResult>, T, TResult, TResult>>, IModule<T, TResult>
             {
                 protected Module(string name) : base(name) { }
                 protected Module(IntPtr handle, IntPtr boxedHandle) : base(handle, boxedHandle) { }
                 internal Module(HType handle, IntPtr? boxedHandle) : base(handle, boxedHandle) { }
-                public abstract TResult forward(T input1);
+                protected abstract TResult forward(T input1);
```
kaiidams, line 1242: I think this is okay, but the user can call `forward()` in PyTorch, you cannot with this change.

NiklasGustafsson: Yes, I was thinking about that. I see some pros and cons with having 'forward' be public:

1. It's confusing to have two ways to call the module -- only one of which is hookable.
2. Existing models wouldn't be forced to change, so hooking would only be possible for new models that are explicitly set up for it.
3. On the other hand, keeping forward as a public will avoid a breaking change.

So, I'm not sure which is better.

dayo05: Some pytorch tutorial calls forward()

MovGP0:
I am wondering how
```csharp
Moudle.register_forward_hook(....)
Module.register_forward_pre_hook(...)
Module.register_full_backward_hook(...)
``` 
would fit in this pattern.

NiklasGustafsson:
> I am wondering how
> 
> ```cs
> Moudle.register_forward_hook(....)
> Module.register_forward_pre_hook(...)
> Module.register_full_backward_hook(...)
> ```
> 
> would fit in this pattern.

It's in order to implement those that this draft PR was created. Not the backward hook, though, that needs to be implemented in native code.

src/TorchSharp/NN/Module.cs
JustinWick: If you're worried about confusion, why not simply document the differences between the two ways you can call the module and call it a day? Breaking changes without a strict technical requirement are the worst.

NiklasGustafsson:
@kaiidams, @MovGP0, @lostmsu, @xhuan8?

Coming back to this PR -- any new thoughts on this after 2.5 months?

src/TorchSharp/NN/Module.cs
NiklasGustafsson: @kaiidams, @JustinWick, @dayo05 -- I'm persuaded. The Python behavior is actually that the hooks don't get called when you call 'm.forward(x)` instead of `m(x)`. So, this will avoid a breaking change and works more like Pytorch.

I'll make `forward()` public.

NiklasGustafsson:
Converting this to a reviewable PR -- no longer a work in progress. I made the `forward()` method public, which was the main point of discussion.

kaiidams:
LGTM. Does TorchSharp run fast enough with this change?

kaiidams (approved):

NiklasGustafsson:
Okay, back to "WIP" status.

Way back when I introduced TorchVision transforms, I should have gone and looked at the Python source code. If I had, I would have realized that there's no need for the ITransform interface. TV transforms inherit from Module<Tensor,Tensor> and don't need the interface to separate them from other kinds of modules.

I'm thinking that we should make a change now, while the version number is still< 1.0? If it's important, we can still keep the interface to avoid a breaking change, but I'd rather just get rid of it.

@kaiidams, @lostmsu, @JustinWick, @dayo05, @xhuan8 -- Thoughts?

PR #868, opened 2022-12-09
Adding ToNDArray() to 'TensorAccessor'

NiklasGustafsson:
This will allow users to get a .NET array with the same number of dimensions as the source tensor.

### DIFF
--- a/src/Examples/AlexNet.cs
+++ b/src/Examples/AlexNet.cs
@@ -58,6 +58,17 @@ namespace TorchSharp.Examples
             using (var x = avg.reshape(new long[] { avg.shape[0], 256 * 2 * 2 }))
                 return classifier.forward(x);
         }
+
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing) {
+                features.Dispose();
+                avgPool.Dispose();
+                classifier.Dispose();
+                ClearModules();
+            }
+            base.Dispose(disposing);
+        }
     }
 
 }
--- a/src/TorchSharp/Utils/TensorAccessor.cs
+++ b/src/TorchSharp/Utils/TensorAccessor.cs
@@ -44,6 +44,9 @@ namespace TorchSharp.Utils
 
         public T[] ToArray()
         {
+            if (_tensor.ndim < 2)
+                return (T[])ToNDArray();
+
             var result = new T[Count];
             CopyTo(result);
             return result;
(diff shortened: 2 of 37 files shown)


xhuan8:
TranslateIndex() is slow, a simple index++ should work

```
case 2:
                unsafe {
                    var result = new T[shape[0], shape[1]];
                    T* ptr = (T*)_tensor_data_ptr;
                    int index = 0;
                    for (long i0 = 0; i0 < shape[0]; i0++) {
                        for (long i1 = 0; i1 < shape[1]; i1++) {
                            result[i0, i1] = ptr[index++];
                        }
                    }
                    return result;
                }
```

NiklasGustafsson:
@xhuan8, unfortunately, it's not that simple -- only on contiguous tensors will this be the case. You always have to take the strides into account, since the tensor may be a (non-contigous) view over another tensor. 

That said, there's no need to call TranslateIndex for every element -- each nested loop can increment by the index * stride for that dimension, including the innermost loop (right-most dimension). That should be straightforward to do.

NiklasGustafsson:
I've updated ToNDArray() with a better-performing way to calculate the indices.

src/Examples/AlexNet.cs
```diff
@@ -58,6 +58,17 @@ public override Tensor forward(Tensor input)
             using (var x = avg.reshape(new long[] { avg.shape[0], 256 * 2 * 2 }))
                 return classifier.forward(x);
         }
+
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing) {
```
NiklasGustafsson, line 64: Newer versions of the C# compiler will complain if you never call Dispose() on submodules. Even if most of these instances will live for as long as the process, this is a code cleanliness matter.

src/TorchSharp/Utils/TensorAccessor.cs
```diff
@@ -44,11 +44,119 @@ public TensorAccessor(torch.Tensor tensor)
...
             CopyTo(result);
             return result;
         }
 
+        /// <summary>
+        /// Extract tensor data as a multi-dimensional .NET array, with the same number of dimensions as the tensor.
+        /// </summary>
+        /// <returns>An array object, which should be cast to the concrete array type.</returns>
+        /// <exception cref="NotImplementedException">Thrown if the tensor has more than 6 dimensions.</exception>
+        public System.Array ToNDArray()
```
NiklasGustafsson, line 60: The nested loops in this method are necessary in order to apply the right strides. A tensor may be a view of another, and it is not correct to assume that storage is contiguous.

ChengYen-Tang: So, is there something wrong with my PR?

ChengYen-Tang: So, is there something wrong with my PR?

NiklasGustafsson: I added this comment for reviewers. Before accepting your PR, I would like to see the relative performance -- the use reflection `array.SetValue()` and the repeated calls to 'TranslateIndex' are likely to have some impact on performance, possibly significant. 

As @xhuan8 pointed out, a simple index++ should work, except that the strides need to be taken in to account. Even if the reflection-based approach is slow, it could potentially be used to cover additional, less common dimensionality. Rather than calling TranslateIndex(), a solution based on recursion could be used (SetValue() would still have to be used).

NiklasGustafsson: And on a procedural note -- your PR will have to be against this repo, not mine, since any submission will have to be subject to the TorchSharp CLA, which would be missed if I take your contribution into my personal repo.

ChengYen-Tang: OK

NiklasGustafsson: I've merged this PR now -- if you want to generalize the ToNDArray() to N dimensions, please submit a PR against this repo, so that we can handle the CLA properly.

Thanks for your interest and help!

PR #884, opened 2023-01-13
Add support for slicing a Sequential object.

NiklasGustafsson:


NiklasGustafsson:
@tarekgh -- a quick review? Not a lot of code here.

src/TorchSharp/NN/Sequential.cs
```diff
@@ -176,6 +181,41 @@ public override nn.Module apply(Action<nn.Module> fn)
...
 
+            /// <summary>
+            /// Module indexer.
+            /// </summary>
+            [IndexerName("SequentialItems")]
+            public Sequential this[(int? start, int? end) index] {
+                get {
+                    var result = new Sequential(Array.Empty<torch.nn.Module<Tensor, Tensor>>());
+                    var start = index.start.HasValue ? index.start.Value : 0;
+                    var end = index.end.HasValue ? index.end.Value : _modules.Count;
```
tarekgh, line 192: >index.end.Value [](http://example.com/codeflow?start=51&length=15)

You may make this `Math.Min(index.end.Value, _modules.Count)` and then you can remove the condition `i < _modules.Count ` from the loop.

src/TorchSharp/NN/Sequential.cs
```diff
@@ -176,6 +181,41 @@ public override nn.Module apply(Action<nn.Module> fn)
                 }
             }
 
+            /// <summary>
+            /// Module indexer.
+            /// </summary>
+            [IndexerName("SequentialItems")]
+            public Sequential this[(int? start, int? end) index] {
```
tarekgh, line 188: >(int? start, int? end [](http://example.com/codeflow?start=35&length=21)

do we need to protect against negative values?

src/TorchSharp/NN/Sequential.cs
```diff
@@ -176,6 +181,41 @@ public override nn.Module apply(Action<nn.Module> fn)
...
+                    var end = index.end.HasValue ? index.end.Value : _modules.Count;
+
+                    for (var i = start; i < _modules.Count && i < end; i++) {
+                        result.Add(_names[i], _modules[i]);
+                    }
+                    return result;
+                }
+            }
+
+#if !NETSTANDARD2_0_OR_GREATER
```
tarekgh, line 201: I guess this condition should be 

`#if NETSTANDARD2_1_OR_GREATER`

NiklasGustafsson: Hmm... We've been using the 2_0 for a while now (you introduced it, I believe). What's changed?

NiklasGustafsson: And, by the way, it didn't work in the test project -- I had to use 'NETFRAMEWORK' instead.

tarekgh: Looking more I think `#if  NETCOREAPP3_1_OR_GREATER` would be better.

>  We've been using the 2_0 for a while now (you introduced it, I believe). What's changed?

Nothing changed. original condition `!NETSTANDARD2_0_OR_GREATER` can cause issue in the future if start support like ns2.1 which exclude these new methods.

src/TorchSharp/NN/Sequential.cs
```diff
@@ -176,6 +181,41 @@ public override nn.Module apply(Action<nn.Module> fn)
...
+            /// <summary>
+            /// Module indexer.
+            /// </summary>
+            [IndexerName("SequentialItems")]
+            public Sequential this[System.Range index] {
```
tarekgh, line 206: > public Sequential this[System.Range index] [](http://example.com/codeflow?start=11&length=43)

can we implement this by just calling the other indexer `[start, end]`? just to avoid duplicating the code.

src/TorchSharp/NN/Sequential.cs
NiklasGustafsson: So, then I should replace it elsewhere, too, then?

src/TorchSharp/NN/Sequential.cs
NiklasGustafsson: Except that the other indexer doesn't have the 'FromEnd' capability that System.Range has.

NiklasGustafsson: Although I guess the (start,end) tuple would yield the right answer.

src/TorchSharp/NN/Sequential.cs
tarekgh: You may address this in another PR. I think changing it would be good in general.

tarekgh (approved):

Issue #889, opened 2023-01-17
Enhance performance with CUDA by exposing allow_tf32 properties 


dje-dev:
In April of 2022 it was announced:  "Update: in consultation with our colleagues at NVIDIA we will be changing the default value of torch.backends.matmul.allow_tf32 to False." https://dev-discuss.pytorch.org/t/pytorch-and-tensorfloat32/504''

The result is much slower TorchSharp performance with CUDA for many common GPUs. 

It would be desirable to be able to enable these in code:
  torch.backends.cuda.matmul.allow_tf32 = True
  torch.backends.cudnn.allow_tf32 = True
but this does not currently appear possible in Torchsharp.

As a workaround, I have recovered the speedup (of 2.15x on a medium sized transformer model) by using this line at the beginning of the program:
  Environment.SetEnvironmentVariable("NVIDIA_TF32_OVERRIDE", "1");

NiklasGustafsson:
@dje-dev -- that you for noticing this announcement. I will look into how to do this in native code, since that's all we have to play with here. TorchSharp is just a thin layer on top of libtorch.

NiklasGustafsson:
I found these in the `at` namespace:
```C++
  bool allowTF32CuDNN() const;
  void setAllowTF32CuDNN(bool);
  bool allowTF32CuBLAS() const;
  void setAllowTF32CuBLAS(bool);
```

Presumably, `setAllowTF32CuDNN` corresponds to `torch.backends.cudnn.allow_tf32 = ...` and `setAllowCuBLAS()` corresponds to `torch.backends.cuda.allow_tf32 = ...`.

PR #890, opened 2023-01-18
Added some CUDA and CUDNN backend properties.

NiklasGustafsson:
Adding some backend properties, per issue #889.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -2,6 +2,14 @@
 
 Releases, starting with 9/2/2021, are listed with the most recent release at the top.
 
+## NuGet Version 0.99.3
+
+__API Changes__:
+
+Fixing misspelling of 'DetachFromDisposeScope,' deprecating the old spelling.<br/>
+Adding allow_tf32<br/>
+
+
 ## NuGet Version 0.99.2
 
 __API Changes__:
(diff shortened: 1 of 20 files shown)


NiklasGustafsson:
@dje-dev, please take a look at this PR.

dje-dev:
Thank you for the quick enhancement. Looks good. The only suggestion is to consider carrying forward the distinction of using "cublas" in some of the names (as you already do with "cudnn") rather than referring to them generically as "cuda".

NiklasGustafsson:
> Thank you for the quick enhancement. Looks good. The only suggestion is to consider carrying forward the distinction of using "cublas" in some of the names (as you already do with "cudnn") rather than referring to them generically as "cuda".

The intent was to follow the Pytorch namespace hierarchy.

dje-dev:
In numerous places, including the DLLExport method signatures but also in the C++ shown here, the underlying name has "CuBLAS" rather than the more generic "cuda" so it seemed a more directly translation to use "cublas."
void THSBackend_cuda_set_allow_tf32(const bool flag)
 {
     CATCH(at::globalContext().setAllowTF32CuBLAS(flag););
 }

NiklasGustafsson:
> In numerous places, including the DLLExport method signatures but also in the C++ shown here, the underlying name has "CuBLAS" rather than the more generic "cuda" so it seemed a more directly translation to use "cublas." void THSBackend_cuda_set_allow_tf32(const bool flag) { CATCH(at::globalContext().setAllowTF32CuBLAS(flag);); }

Got it. I don't generally spend a lot of time on the interop names, but it's a fair point.

Issue #904, opened 2023-02-03
Missing add_module api


ChengYen-Tang:
Hi, I can't find add_module in TorchSharp.
https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.add_module

ChengYen-Tang:
@NiklasGustafsson,

From the source code I found the register_module.
![image](https://user-images.githubusercontent.com/32535494/216864626-d1c8f325-afa9-415d-914a-6c2bba971ffe.png)

PR #906, opened 2023-02-06
add_module API

ChengYen-Tang:
#904

### DIFF
--- a/src/TorchSharp/NN/Module.cs
+++ b/src/TorchSharp/NN/Module.cs
@@ -6,10 +6,9 @@ using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using TorchSharp.Modules;
+using static TorchSharp.PInvoke.LibTorchSharp;
 using static TorchSharp.torch;
 using static TorchSharp.Utils.LEB128Codec;
-using static TorchSharp.PInvoke.LibTorchSharp;
-using static Google.Protobuf.Reflection.SourceCodeInfo.Types;
 
 namespace TorchSharp
 {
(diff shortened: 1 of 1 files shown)


src/TorchSharp/NN/Module.cs
```diff
@@ -757,28 +756,51 @@ public virtual void register_parameter(string name, Parameter param)
...
-                /// <param name="submodule">The module to register.</param>
+                /// <param name="name">
+                /// name of the child module.
+                /// The child module can be accessed from this module using the given name
+                /// </param>
+                /// <param name="module">child module to be added to the module.</param>
+                /// <exception cref="ArgumentException"></exception>
                 /// <exception cref="InvalidOperationException"></exception>
-                public virtual void register_module(string name, Module submodule)
+                public virtual void add_module(string name, Module module)
```
NiklasGustafsson, line 769: Since they are aliases, I would suggest leaving `register_module()` in place and virtual, and then define `add_module()` as non-virtual and have it call `register_module()`. There's no reason for both of them to be virtual, and in case anyone has already overriden `register_module`, you would have a breaking code change if it's no longer virtual.

src/TorchSharp/NN/Module.cs
```diff
@@ -757,28 +756,51 @@ public virtual void register_parameter(string name, Parameter param)
...
-                /// <param name="name">Name of the submodule.</param>
-                /// <param name="submodule">The module to register.</param>
+                /// <param name="name">
+                /// name of the child module.
+                /// The child module can be accessed from this module using the given name
+                /// </param>
+                /// <param name="module">child module to be added to the module.</param>
+                /// <exception cref="ArgumentException"></exception>
                 /// <exception cref="InvalidOperationException"></exception>
-                public virtual void register_module(string name, Module submodule)
```
NiklasGustafsson, line 765: Changing the name of a parameter of a public method is a breaking change, so unless it's absolutely necessary, it should be avoided.

NiklasGustafsson (changes requested):

NiklasGustafsson:
Hi @ChengYen-Tang -- I usually ask all contributors to introduce themselves in this thread:

https://github.com/dotnet/TorchSharp/discussions/333

It is voluntary, but it's nice for everyone to know who's helping with TorchSharp.

ChengYen-Tang:
@NiklasGustafsson,

Is this okay.

NiklasGustafsson:
> @NiklasGustafsson,
> 
> Is this okay.

It looks okay to me. As soon as I can get a clean test run (they intermittently fail due to concurrency), I'll merge.

NiklasGustafsson (approved):

PR #919, opened 2023-02-15
Detect 32-bit process and throw an exception.

NiklasGustafsson:


tarekgh (dismissed):

src/Native/LibTorchSharp/THSTensor.h
```diff
@@ -1010,6 +1010,8 @@ EXPORT_API(Tensor) THSTensor_randperm(const Generator gen, const int64_t n, cons
 
 EXPORT_API(Tensor) THSTensor_randperm_out(const Generator gen, const int64_t n, const Tensor out);
 
+EXPORT_API(Tensor) THSTensor_from_file(const char* filename, const int8_t shared, const int64_t size, const int8_t scalar_type, const int device_type, const int device_index, const bool requires_grad);
```
lostmsu, line 1013: My only concern with this would be localization. Does it work on Unicode file names?

NiklasGustafsson: That's the character type the underlying c10::from_file() function takes. There's no documentation (that I can find) suggesting whether it supports any encodings.

NiklasGustafsson: Let me see what happens in the test if I change it!

lostmsu: My bet would be on UTF-8, but who knows.

NiklasGustafsson: It does not work.

NiklasGustafsson: This is a problem that I think we have across the board where we are passing strings for filenames. I'll open a separate issue to track fixing that.

lostmsu: Maybe try `UnmanagedType.LPUTF8Str`?

NiklasGustafsson: I did, but it's not available on all versions of .NET that we support.

tarekgh: Using ChatGPT it says the following :-) 

image

lostmsu: @NiklasGustafsson if it worked, can pass `byte[]` instead of string and explicitly obtain it using `Encoding.UTF8.GetBytes`

NiklasGustafsson: I'll try that.

lostmsu: @tarekgh this does not sound promising. Will probably require explicit conversion with a known encoding such as UTF8 on both sides.

tarekgh: Right, if we are sure, it always uses UTF-8 then your suggestion is good. At least we can manually convert to UTF-8 when compiling to netstandards and use `UnmanagedType.LPUTF8Str` when compiling for .NET.

NiklasGustafsson: Okay, so this ugly hack worked:

```C#
        [DllImport("LibTorchSharp")]
        internal static extern IntPtr THSTensor_from_file(IntPtr filename, sbyte shared, long size, sbyte scalarType, int deviceType, int deviceIndex, [MarshalAs(UnmanagedType.U1)] bool requires_grad);

[...]

var bytes = Encoding.UTF8.GetBytes(filename);
var newBytes = new byte[bytes.Length+1];
Array.Copy(bytes, newBytes, bytes.Length);
newBytes[bytes.Length] = 0;

fixed (byte* ptr = newBytes) {
    var handle = THSTensor_from_file((IntPtr)ptr, (sbyte)(!shared.HasValue ? -1 : shared.Value ? 1 : 0), size.HasValue ? size.Value : -1, (sbyte)dtype, (int)device.type, device.index, requires_grad);
    if (handle == IntPtr.Zero) { CheckForErrors(); }
        return new Tensor(handle);
}

```

NiklasGustafsson: From `dir *.dat`:

```
02/15/2023  16:06            16,384 tensor_åöä_ασδφεες_አስድፋስድፍ.dat
```

Latin, Swedish, Greek, and Amharic...

tarekgh: can't you use `GetBytes(String, Int32, Int32, Byte[], Int32)` instead of `GetBytes(filename);` and save the other allocation and copying?

NiklasGustafsson: I don't know how long the byte array needs to be until it's been encoded.

tarekgh: You can use Encoding.GetMaxByteCount(Int32) to know the enough buffer length to use for allocation.

lostmsu: Considering this function is opening a file, I'd say a readable implementation is better than superoptimized one.

NiklasGustafsson:
```C#
var bytes = new byte[Encoding.UTF8.GetMaxByteCount(filename.Length)];
var len = Encoding.UTF8.GetBytes(filename, 0, filename.Length, bytes, 0);
bytes[len] = 0;
```

lostmsu: @NiklasGustafsson you seem to forgot `+ 1` in `GetMaxByteCount() + 1`.

Also, I am unsure you actually need the `fixed` bit. Could probably delegate marshaling of the byte array to PInvoke.

NiklasGustafsson: Thank you!

tarekgh: One last comment for me :-) 

Using `Encoding.UTF8` will make the encoding produce BOM in the beginning of the buffer. I am not sure if this can cause any issue in the native side as it looks like you already tested it. If need to be in safe side, you may define a static field like 

```C#
private static readonly Encoding s_utfEncoding = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false, throwOnInvalidBytes: false);
```

and then use `s_utfEncoding` instead of `Encoding.UTF8`

NiklasGustafsson: @tarekgh, @lostmsu -- thanks for all the help! This will help to support Unicode in other file name APIs, too.

PR #926, opened 2023-02-17
Missing methods implemented

NiklasGustafsson:
Not all, but most...

This is fairly big, but straight-forward. Just a bunch of native interop added.

@MovGP0, @lostmsu -- if you have the time to take a look, that would be appreciated.


src/TorchSharp/PInvoke/LibTorchSharp.THSTensor.cs
```diff
@@ -992,6 +1020,9 @@ internal static partial class LibTorchSharp
         [DllImport("LibTorchSharp")]
         internal static extern IntPtr THSTensor_outer(IntPtr input, IntPtr vec2);
 
+        [DllImport("LibTorchSharp")]
+        internal static extern IntPtr THSTensor_ormqr(IntPtr input, IntPtr tau, IntPtr other, bool left, bool transpose);
```
tarekgh, line 1024: do we need to add the attribute `[MarshalAs(UnmanagedType.U1)]` to the bool parameters?

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -5654,6 +5771,25 @@ public Tensor roll((long, long) shifts, (long, long) dims)
...
+            /// </summary>
+            /// <param name="k">The number of times to rotate.</param>
+            /// <param name="dims">Axes to rotate</param>
+            public Tensor rot90(long k = 1, (long, long)? dims = null)
+            {
+                if (!dims.HasValue) {
+                    dims = (0, 1);
+                }
+
+                var res =
```
tarekgh, line 5787: nit: have the whole assignment in the same line.

src/TorchSharp/PInvoke/LibTorchSharp.THSTensor.cs
NiklasGustafsson: Yes! Thank you.

tarekgh (dismissed):

src/TorchSharp/PInvoke/LibTorchSharp.THSTensor.cs
NiklasGustafsson: Do you know if it needs to go on the 'bool' return types, too, or just arguments?

tarekgh: Yes, need to apply it to the return value too. Here is a good [example](https://learn.microsoft.com/en-us/visualstudio/code-quality/ca1414?view=vs-2022&tabs=csharp#example).

PR #946, opened 2023-03-06
WIP: Re-implement Linear in managed code.

NiklasGustafsson:
There are a couple of things in this WIP/draft PR:

1. Overriding `_to()` implementations in modules that are known not to have any parameters or buffers. This will save a small amount of runtime overhead.

2. An alternative implementation of Linear to more closely align with how it works in PyTorch. Rather than creating native module instances and managing their lifetimes, this approach only involves a .NET instance, and then calls into the torch.nn.functional APIs to perform the forward pass. It's simpler and gets us out of the business of managing native module instances. The downside is that we need to do a whole bunch of work without any new functionality. It's just getting rid of some technical debt and aligning better with the Python implementation, which could be enough of a reason.

@MovGP0, @lostmsu, @kaiidams, @ChengYen-Tang, @dayo05 -- your thoughts on #2?


MovGP0:
```csharp
internal Linear(long inputSize, long outputSize, bool hasBias = true, Device device = null, ScalarType? dtype = null) : base(nameof(Linear))
```
It should be `Device? device = null`

MovGP0:
```csharp
init.kaiming_uniform_(weight, a: Math.Sqrt(5));
```
I think we could cache the result of `Math.Sqrt(5)` in a `private static readonly` variable. But probably won't matter in the grand scheme of things.

NiklasGustafsson:
> It should be `Device? device = null`. Same problem with the other constructors.

Right. I took the `#nullable enable` out. Putting it back.

NiklasGustafsson:
Well, caching it probably won't make much of a difference, but it also won't hurt.

src/TorchSharp/NN/Linear.cs
```diff
@@ -13,48 +12,53 @@ namespace Modules
...
 
-            public override Tensor forward(Tensor tensor)
+            protected override void Dispose(bool disposing)
             {
-                var res = THSNN_Linear_forward(handle, tensor.Handle);
-                if (res == IntPtr.Zero) { torch.CheckForErrors(); }
-                return new Tensor(res);
+                if (disposing) {
+                    weight.Dispose();
+                    if (bias is not null) bias.Dispose();
```
lostmsu, line 38: NIT: `bias?.Dispose`

src/TorchSharp/NN/Linear.cs
```diff
@@ -13,48 +12,53 @@ namespace Modules
...
-                    if (res == IntPtr.Zero) { torch.CheckForErrors(); }
-                    return ((res == IntPtr.Zero) ? null : new Parameter(res));
-                }
+            public Parameter bias {
                 set {
-                    THSNN_Linear_set_bias(handle, value?.Handle ?? IntPtr.Zero);
-                    torch.CheckForErrors();
-                    ConditionallyRegisterParameter("bias", value);
+                    _bias = value;
+                    ConditionallyRegisterParameter("bias", bias);
```
lostmsu, line 45: Will this properly re-register parameter if I manually replace it?

src/TorchSharp/NN/Linear.cs
```diff
@@ -13,48 +12,53 @@ namespace Modules
...
-                    if (res == IntPtr.Zero) { torch.CheckForErrors(); }
-                    return (res == IntPtr.Zero) ? null : new Parameter(res);
-                }
+            public Parameter weight {
+                get => _weight;
                 set {
-                    THSNN_Linear_set_weight(handle, value!.Handle);
-                    torch.CheckForErrors();
-                    ConditionallyRegisterParameter("weight", value);
+                    if (value is null) throw new ArgumentNullException("weight");
```
lostmsu, line 54: NIT: `nameof(weight)`

src/TorchSharp/NN/Linear.cs
```diff
@@ -13,48 +12,53 @@ namespace Modules
     {
         public sealed class Linear : torch.nn.Module<Tensor, Tensor>
         {
-            internal Linear(IntPtr handle, IntPtr boxedHandle) : base(handle, boxedHandle)
+            internal Linear(long inputSize, long outputSize, bool hasBias = true, Device device = null, ScalarType? dtype = null) : base(nameof(Linear))
```
lostmsu, line 15: Is there any reason to not have this constructor to be `public`? It does not take handles anymore.

src/TorchSharp/NN/Linear.cs
```diff
@@ -87,7 +88,7 @@ public static partial class functional
                 /// <param name="weights">Weights of shape (Hout,Hin) or (Hin)</param>
                 /// <param name="bias">Bias of shape (Hout) or ()</param>
                 /// <returns>A tensor of shape (*,Hout) where '*' is the same as the subshape of the input.</returns>
-                public static Tensor linear(Tensor input, Tensor weights, Tensor? bias = null)
+                public static Tensor linear(Tensor input, Tensor weights, Tensor bias = null)
```
lostmsu, line 91: Wrong change.

lostmsu (commented):
Mostly NITs on the code itself, however I am unsure about the idea of manually reimplemented `Linear`. What is the plan for ensuring we actually track the upstream changes?

As far as PRs go, I think the part that overrides `_to` should be separate. OCD 🙃

src/TorchSharp/NN/Linear.cs
NiklasGustafsson: Yes. @MovGP0 pointed that out, too.

src/TorchSharp/NN/Linear.cs
NiklasGustafsson: Only because the Python-likeness we're striving for would prefer that users use the factories in `torch.nn` instead of the constructors.

src/TorchSharp/NN/Linear.cs
NiklasGustafsson: Yes, but I found some complications with DisposeScopes that aren't handled. If you overwrite, the bias (and weight, of course) can (and should) be immediately disposed.

src/TorchSharp/NN/Linear.cs
NiklasGustafsson: Yup. Already fixed.

src/TorchSharp/NN/Linear.cs
lostmsu: Might be worth putting that in the `DisposeScope` related section of documentation.

NiklasGustafsson:
> Mostly NITs on the code itself, however I am unsure about the idea of manually reimplemented `Linear`. What is the plan for ensuring we actually track the upstream changes?

We would have to get to all of the modules, so it's not just Linear. In some cases, it may be easier to track. There have been times when we've moved from one version to another, and the native module API didn't have some new feature, but the functional API did.

src/TorchSharp/NN/Linear.cs
```diff
@@ -70,12 +74,9 @@ public static partial class nn
             /// <param name="hasBias">If set to false, the layer will not learn an additive bias.</param>
             /// <param name="device">The desired device of the parameters and buffers in this module</param>
             /// <param name="dtype">The desired floating point or complex dtype of the parameters and buffers in this module</param>
-            public static Linear Linear(long inputSize, long outputSize, bool hasBias = true, Device? device = null, ScalarType? dtype = null)
+            public static Linear Linear(long inputSize, long outputSize, bool hasBias = true, Device device = null, ScalarType? dtype = null)
```
kaiidams, line 77: Should this receive `in_features` and `out_features` to be compatible with PyTorch?

src/TorchSharp/NN/Linear.cs
```diff
@@ -70,12 +74,9 @@ public static partial class nn
...
             {
-                var res = THSNN_Linear_ctor(inputSize, outputSize, hasBias, out var boxedHandle);
```
kaiidams, line 75: Is this going to be deleted from https://github.com/dotnet/TorchSharp/blob/main/src/Native/LibTorchSharp/THSNN.cpp ?

kaiidams:
I hope these temporary objects are gone with this change.
```cs
                public static Tensor relu(Tensor x, bool inplace = false)
                {
                    using (var m = nn.ReLU(inplace)) {
                        return m.call(x);
                    }
                }
```

src/TorchSharp/NN/Bilinear.cs
```diff
@@ -13,46 +13,57 @@ namespace Modules
...
+                var bound = 1 / Math.Sqrt(weight!.shape[1]);
+
+                init.uniform_(_weight, -bound, bound);
+
+                if (hasBias) {
+                    bias = torch.empty(out_features, device: device, dtype: dtype).AsParameter();
+                    var (fanIn, _) = init.CalculateFanInAndFanOut(weight);
+                    init.uniform_(_bias, -bound, bound);
+                }
+                //NOTE: it's important not to call 'RegisterComponents' here.
```
lostmsu, line 28: Adding why would be helpful for future readers. I assume it is called in `base`?

src/TorchSharp/NN/Bilinear.cs
```diff
@@ -13,46 +13,57 @@ namespace Modules
...
+                    if (value is null) throw new ArgumentNullException(nameof(weight));
+                    if (value.Handle != _weight?.Handle) {
+                        _weight?.Dispose();
+                        _weight = (value.DetachFromDisposeScope() as Parameter)!;
+                        ConditionallyRegisterParameter(nameof(weight), _weight);
+                    }
                 }
             }
+
+            private Parameter? _weight;
```
lostmsu, line 66: Why not make this `readonly` and non-nullable?

src/TorchSharp/NN/Dropout.cs
```diff
@@ -12,9 +12,13 @@ namespace Modules
...
         /// This class is used to represent a dropout module.
         /// </summary>
-        public sealed class Dropout : torch.nn.Module<Tensor, Tensor>
+        public sealed class Dropout : ParamLessModule<Tensor, Tensor>
         {
-            internal Dropout(IntPtr handle, IntPtr boxedHandle) : base(handle, boxedHandle) { }
+            internal Dropout(double p = 0.5, bool inplace = false) : base(nameof(Dropout))
+            {
+                this._p = p;
+                this._inplace = inplace;
```
lostmsu, line 20: Should these be part of the state dict?

src/TorchSharp/NN/Parameter.cs
```diff
@@ -26,6 +26,12 @@ public class Parameter : Tensor
             public Parameter(Tensor data, bool requires_grad = true) :
                 base(data.with_requires_grad(requires_grad).MoveHandle())
             {
+                var scope = data.OwningDisposeScope;
+                if (scope is not null) {
+                    this.OwningDisposeScope = scope;
+                    scope.Include(this);
+                    scope.Detach(data);
```
lostmsu, line 33: This needs some explanation.

test/TorchSharpTest/NN.cs
```diff
@@ -325,6 +325,8 @@ public void TestIdentity()
                 var input = torch.randn(new long[] { 1, 1000 }, device: device);
                 var output = lin.call(input);
 
+                output[0, 511] = 10; // When we modify the copy, the original should be altered, too.
+
                 Assert.Equal(device.type, output.device_type);
```
lostmsu, line 330: Is this how `Identity` behaves in PyTorch? o-O

NiklasGustafsson: Yup:

```Python
>>> id = torch.nn.Identity()
>>> input = torch.zeros(10,10)
>>> output = id(input)
>>> output[0,0] = 13
>>> input
tensor([[13.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.],
        [ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.],
        [ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.],
        [ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.],
        [ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.],
        [ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.],
        [ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.],
        [ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.],
        [ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.],
        [ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.]])
```

NiklasGustafsson: From the PyTorch source code:

```Python
    def forward(self, input: Tensor) -> Tensor:
        return input
```

lostmsu:
Is this PR still valid?

NiklasGustafsson:
Depends on what you mean. I submitted it with partial implementation of modules, just to get feedback. It would require a full implementation of all modules in order to be merged. So far, the "WIP" tag still applies.

PR #964, opened 2023-03-24
Export optimizer state from PyTorch, import it into TorchSharp

NiklasGustafsson:
This PR enables saving and loading optimizer state going from PyTorch to TorchSharp (but not the other way around). All optimizers except LBFGS are supported.


TheBlackPlague (approved):
This PR LGTM. Sorry for the late review, been busy.

PR #968, opened 2023-03-27
Free up native strings passed back to .NET

NiklasGustafsson:


NiklasGustafsson:
This was meant to fix the string interop memory leak, but I haven't quite figured out why things are crashing when I implemented it, and then the changes shouldn't have been on 'main' to begin with. Therefore, what remains are some trivial cleanup fixes.

PR #960, opened 2023-03-16
Add torch.histogram

ChengYen-Tang:
#939

PR #975, opened 2023-04-05
Numpy histogram

ChengYen-Tang:
https://github.com/dotnet/TorchSharp/pull/960#issuecomment-1472322015


NiklasGustafsson:
@ChengYen-Tang -- lots of good work here!! Could you please add a little bit more in terms of what this PR is for, and what scenarios it supports?

src/TorchSharp/Utils/Histogram.cs
```diff
@@ -0,0 +1,288 @@
+// Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using static TorchSharp.torch;
+
+namespace TorchSharp.Utils
+{
+    // https://github.com/numpy/numpy/blob/v1.24.0/numpy/lib/histograms.py
```
NiklasGustafsson, line 10: This is derived from Numpy and should therefore add an attribution notice, along with the copyright notice, to LICENSE in the root folder.

ChengYen-Tang: Can you give me an example?

ChengYen-Tang:
> @ChengYen-Tang -- lots of good work here!! Could you please add a little bit more in terms of what this PR is for, and what scenarios it supports? I ask, because I don't see anything that's in the `torch` or `torch.tensor` scopes here.

https://pytorch.org/docs/stable/tensorboard.html#torch.utils.tensorboard.writer.SummaryWriter.add_histogram
Required by tensorboard add_histogram method.

src/TorchSharp/Utils/Histogram.cs
GeorgeS2019: @NiklasGustafsson 
For future TorchSharp users, many would use numpy-related codes, links, references,  this advice needs to be easily found

> This is derived from Numpy and should therefore add an attribution notice, along with the copyright notice, to LICENSE in the root folder

@ChengYen-Tang 

Are you using any Numpy for .NET libraries? If not, what advice would u give to new TorchSharp users who have numpy integration challenge with TorchSHarp?

NiklasGustafsson: > Can you give me an example?

I was incorrect. It's in another file in the root folder: https://github.com/dotnet/TorchSharp/blob/main/THIRD-PARTY-NOTICES.TXT. The various sections of 3rd-party notices are separated by '==============' lines

ChengYen-Tang:
> > Required by the bins parameter in the tensorboard add_histogram method.
> 
> Should it be internal rather than public, then?

I think maybe others need this method too, this can be featured in torchsharp

src/TorchSharp/Utils/Histogram.cs
ChengYen-Tang: > @NiklasGustafsson For future TorchSharp users, many would use numpy-related codes, links, references, this advice needs to be easily found
> 
> > This is derived from Numpy and should therefore add an attribution notice, along with the copyright notice, to LICENSE in the root folder
> 
> @ChengYen-Tang
> 
> Are you using any Numpy for .NET libraries? If not, what advice would u give to new TorchSharp users who have numpy integration challenge with TorchSHarp?

@GeorgeS2019 
https://github.com/Quansight-Labs/numpy.net

ChengYen-Tang: > > Can you give me an example?
> 
> I was incorrect. It's in another file in the root folder: https://github.com/dotnet/TorchSharp/blob/main/THIRD-PARTY-NOTICES.TXT. The various sections of 3rd-party notices are separated by '==============' lines.
> 
> Add another section, make a note that it's NumPy, and then copy the NumPy license text into the file.

Done.

NiklasGustafsson:
> I think maybe others need this method too, this can be featured in torchsharp

Yes, I agree. In that case, I would start it out as internal. Changing it to public if there is demand won't be a breaking change.

src/TorchSharp/Tensor/torch.ComparisonOps.cs
```diff
@@ -276,6 +276,21 @@ public static Tensor searchsorted(Tensor sorted_sequence, Scalar values, bool ou
...
+        /// If bins is a 1D tensor, it specifies the sequence of bin edges including the rightmost edge. It should contain at least 2 elements and its elements should be increasing.
+        /// </summary>
+        /// <param name="input"> the input tensor. </param>
+        /// <param name="bins"> int or 1D Tensor. If int, defines the number of equal-width bins. If tensor, defines the sequence of bin edges including the rightmost edge. </param>
+        /// <param name="range"> Defines the range of the bins. </param>
+        /// <param name="density"> If False, the result will contain the count (or total weight) in each bin. If True, the result is the value of the probability density function over the bins, normalized such that the integral over the range of the bins is 1. </param>
+        /// <returns></returns>
+        internal static (Tensor hist, Tensor bin_edges) histogram(Tensor input, HistogramBinSelector bins, (double min, double max)? range = null, bool density = false)
+            => Utils.Histogram.histogram(input, bins, range, density);
+
```
ChengYen-Tang, line 293: @NiklasGustafsson ,
Is this okay.

ChengYen-Tang: But this cannot be unit tested

NiklasGustafsson: Yeah, you're right.

NiklasGustafsson: Unit testing is important.

ChengYen-Tang: > Unit testing is important.

I mean, if it is internal, it cannot be unit tested, because the unit test project cannot find this method
so the build fails

NiklasGustafsson: We could make them friend assemblies, but it's complicated. I'm okay with public, especially since it's not in the `torch` namespace.

ChengYen-Tang: So, you mean I can change it back to public?

NiklasGustafsson: Yes

GeorgeS2019:
@ChengYen-Tang 
Currently, I am occupied with other urgent tasks. Please kindly remove me from the reviewer list.

src/TorchSharp/Tensor/torch.ComparisonOps.cs
```diff
@@ -276,6 +276,21 @@ public static Tensor searchsorted(Tensor sorted_sequence, Scalar values, bool ou
             return new Tensor(res);
         }
 
+        /// https://github.com/numpy/numpy/blob/v1.24.0/numpy/lib/histograms.py#L679
```
NiklasGustafsson, line 279: I know this is a little bit of a nit-pick, but does this method belong in a file called 'ComparisonOps'?

GeorgeS2019:
@ChengYen-Tang 

Thank you very much.

Not being greedy, if you have time, a small polyglot notebook sample showing how to use the features you have written :-)

GeorgeS2019:
The aim is to show the community that What you do in PyTorch, the same can be done in TorchSharp.

Try to be as close as possible to this and turn that into dotnet PolyGlot notebook experience
https://pytorch.org/tutorials/recipes/recipes/tensorboard_with_pytorch.html

@NiklasGustafsson 

What do you think to put the polyglot notebook here?
https://github.com/dotnet/TorchSharpExamples/tree/main/tutorials/CSharp

NiklasGustafsson:
Yes, I suggest adding another tutorial to the CSharp folder (I can translate to F# if you aren't familiar with it). Maybe use a more useful name than what I have for the others? :-)

Make sure to follow the structure of the other tutorials -- the first cell is the same in all of them and sets up for automatic formatting of tensors as the output of a cell.

PR #1025, opened 2023-06-24
C#-style formatting of tensor strings.

NiklasGustafsson:


GeorgeS2019:
What is the motivation and use case of csharp formating? Will we see example in csharp polyglot notebook in TorchSharp example repos?

PR #1045, opened 2023-07-20
Fixed bug in named_parameters and named_buffers

NiklasGustafsson:


lostmsu:
Would be nice to see what bug was or what the fix changes in the history.

NiklasGustafsson:
Good point -- the implementations of both methods used 'Handle' (which validates and throws) in a situation where 'handle' was appropriate. It is not illegal to use the same module twice in a Sequential or in a custom module, which means that when the `Dispose(bool)` method is invoked and disposes the parameters and buffers, it is possible to encounter already disposed tensors, which will have an invalid handle. Rather than validating and throwing an exception, the fix simply ignores tensors with invalid handles.

PR #1048, opened 2023-07-28
Fix exception in Grayscale.cs

GilesBathgate:
TorchSharp `Greyscale()` transform assumes batch dimension when calculating the channel dim `channel = image.shape[1]` but this is not correct, as for a non-batched tensor `channel = image.shape[0]` The channel dim should be calculated using something such as  `channel = image.shape[image.ndim - 3]`

Example working pytorch:

```python
img = torch.rand((1, 32, 32))
result = torchvision.transforms.functional.rgb_to_grayscale(img)
result = torchvision.transforms.Grayscale()(img)
```

Example failing Torchsharp
```c#
var img = torch.rand(1, 32, 32);
var res = torchvision.transforms.functional.rgb_to_grayscale(img);
var res2 = torchvision.transforms.Grayscale().call(img);
```

### DIFF
--- a/src/TorchVision/Grayscale.cs
+++ b/src/TorchVision/Grayscale.cs
@@ -17,8 +17,9 @@ namespace TorchSharp
 
             public Tensor call(Tensor input)
             {
-                if (input.shape[1] == 1) return input.alias();
-                if (input.shape[1] != 3)
+                int cDim = (int)input.Dimensions - 3;
+                if (input.shape[cDim] == 1) return input.alias();
+                if (input.shape[cDim] != 3)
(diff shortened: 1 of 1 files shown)


NiklasGustafsson (approved):

NiklasGustafsson:
Thank you @GilesBathgate, for starting to contribute to TorchSharp! Much appreciated.

If you have time, please add a brief intro about yourself to this thread: https://github.com/dotnet/TorchSharp/discussions/333. It's nice for everyone (users and other contributors) to know who's been working on the code.

Issue #1049, opened 2023-07-28
Warning emitted from RandomRotation


GilesBathgate:
I see the following warning when using RandomRotation

Example code:

```csharp
var img = torch.rand(3, 32, 32);
var t = transforms.RandomRotation(30).call(img);
```

Output:
```
[W vision.h:87] Warning: Default grid_sample and affine_grid behavior has changed to align_corners=False since 1.3.0. Please specify align_corners=True if the old behavior is desired. See the documentation of grid_sample for details. (function grid_sample)
```

PR #1051, opened 2023-07-31
Suppress warning by setting align_corners false

GilesBathgate:
This should suppress the warning in #1049


GeorgeS2019:
@GilesBathgate 

Thx for identifying!

GilesBathgate:
@NiklasGustafsson I won't add a unit test for this, since it suppresses a warning from libtorch, not sure if it can be unit tested.

NiklasGustafsson (approved):

PR #1052, opened 2023-07-31
Port RandAugment from pytorch

GilesBathgate:
I wanted to use RandAugment from the pytorch autoaugment module. https://pytorch.org/vision/main/generated/torchvision.transforms.RandAugment.html

I think this is line for line equivalent to the pytorch code, but maybe needs additional improvements specific to TorchSharp. It works as expected, but depends on my previous PRs

### DIFF
--- a/src/TorchVision/AutoAugment.cs
+++ b/src/TorchVision/AutoAugment.cs
@@ -32,7 +32,7 @@ namespace TorchSharp
             protected static Tensor apply_op(
                                     Tensor img,
                                     opType op_name,
-                                    float magnitude,
+                                    double magnitude,
                                     InterpolationMode interpolation,
                                     IList<float>? fill)
             {
@@ -43,7 +43,7 @@ namespace TorchSharp
                             angle: 0.0f,
                             translate: new[] { 0, 0 },
                             scale: 1.0f,
-                            shear: new[] { (float)((180.0f / Math.PI) * Math.Atan(magnitude)), 0.0f },
+                            shear: new[] { (float)((180.0 / Math.PI) * Math.Atan(magnitude)), 0.0f },
                             interpolation: interpolation,
                             fill: fill?.FirstOrDefault());
                     case opType.ShearY:
@@ -52,7 +52,7 @@ namespace TorchSharp
                             angle: 0.0f,
                             translate: new[] { 0, 0 },
                             scale: 1.0f,
-                            shear: new[] { 0.0f, (float)((180.0f / Math.PI) * Math.Atan(magnitude)) },
+                            shear: new[] { 0.0f, (float)((180.0 / Math.PI) * Math.Atan(magnitude)) },
                             interpolation: interpolation,
                             fill: fill?.FirstOrDefault());
                     case opType.TranslateX:
@@ -74,7 +74,7 @@ namespace TorchSharp
                             shear: new[] { 0.0f, 0.0f },
                             fill: fill?.FirstOrDefault());
                     case opType.Rotate:
-                        return F.rotate(img, magnitude, interpolation, fill: fill);
+                        return F.rotate(img, (float)magnitude, interpolation, fill: fill);
                     case opType.Brightness:
                         return F.adjust_brightness(img, 1.0 + magnitude);
                     case opType.Color:
@@ -119,7 +119,6 @@ namespace TorchSharp
 
             public Tensor call(Tensor img)
             {
-                var fill = this.fill;
                 var cDim = img.Dimensions - 3;
                 var height = img.shape[cDim + 1];
                 var width = img.shape[cDim + 2];
@@ -128,12 +127,12 @@ namespace TorchSharp
                     var op_index = torch.randint(op_meta.Count, new[] { 1 }).ToInt32();
                     var op_name = op_meta.Keys.ToList()[op_index];
                     var (magnitudes, signed) = op_meta[op_name];
-                    var magnitude = magnitudes.Dimensions > 0 ? magnitudes[this.magnitude] : 0.0f;
+                    var magnitude = magnitudes.Dimensions > 0 ? magnitudes[this.magnitude].ToDouble() : 0.0;
 
                     if (signed && torch.randint(2, new[] { 1 }).ToBoolean())
                         magnitude *= -1.0;
 
-                    img = apply_op(img, op_name, magnitude.ToSingle(), interpolation, fill);
+                    img = apply_op(img, op_name, magnitude, interpolation, this.fill);
                 }
                 return img;
             }
--- a/src/TorchVision/AutoAugment.cs
+++ b/src/TorchVision/AutoAugment.cs
@@ -9,80 +9,82 @@ namespace TorchSharp
 {
     public static partial class torchvision
     {
-        private static Tensor apply_op(
-                                Tensor img,
-                                string op_name,
-                                float magnitude,
-                                InterpolationMode interpolation,
-                                IList<float>? fill)
+        internal abstract class AutoAugmentBase
         {
-            if (op_name == "ShearX") {
-                img = F.affine(
-                    img,
-                    angle: 0.0f,
-                    translate: new[] { 0, 0 },
-                    scale: 1.0f,
-                    shear: new[] { (float)((180.0f / Math.PI) * Math.Atan(magnitude)), 0.0f },
-                    interpolation: interpolation,
-                    fill: fill?.FirstOrDefault());
-            } else if (op_name == "ShearY") {
-                img = F.affine(
-                    img,
-                    angle: 0.0f,
-                    translate: new[] { 0, 0 },
-                    scale: 1.0f,
-                    shear: new[] { 0.0f, (float)((180.0f / Math.PI) * Math.Atan(magnitude)) },
-                    interpolation: interpolation,
-                    fill: fill?.FirstOrDefault());
-            } else if (op_name == "TranslateX") {
-                img = F.affine(
-                    img,
-                    angle: 0.0f,
-                    translate: new[] { (int)(magnitude), 0 },
-                    scale: 1.0f,
-                    interpolation: interpolation,
-                    shear: new[] { 0.0f, 0.0f },
-                    fill: fill?.FirstOrDefault());
-            } else if (op_name == "TranslateY") {
-                img = F.affine(
-                    img,
-                    angle: 0.0f,
-                    translate: new[] { 0, (int)(magnitude) },
-                    scale: 1.0f,
-                    interpolation: interpolation,
-                    shear: new[] { 0.0f, 0.0f },
-                    fill: fill?.FirstOrDefault());
-            } else if (op_name == "Rotate") {
-                img = F.rotate(img, magnitude, interpolation, fill: fill);
-            } else if (op_name == "Brightness") {
-                img = F.adjust_brightness(img, 1.0 + magnitude);
-            } else if (op_name == "Color") {
-                img = F.adjust_saturation(img, 1.0 + magnitude);
-            } else if (op_name == "Contrast") {
-                img = F.adjust_contrast(img, 1.0 + magnitude);
-            } else if (op_name == "Sharpness") {
-                img = F.adjust_sharpness(img, 1.0 + magnitude);
-            } else if (op_name == "Posterize") {
-                img = F.posterize(img, (int)magnitude);
-            } else if (op_name == "Solarize") {
-                img = F.solarize(img, magnitude);
-            } else if (op_name == "AutoContrast") {
-                img = F.autocontrast(img);
-            } else if (op_name == "Equalize") {
-                img = F.equalize(img);
-            } else if (op_name == "Invert") {
-                img = F.invert(img);
-            } else if (op_name == "Identity") {
-                // Pass
-            } else {
-                throw new ArgumentException($"The provided operator {op_name} is not recognized.");
-            }
+            protected static Tensor apply_op(
+                                    Tensor img,
+                                    string op_name,
+                                    float magnitude,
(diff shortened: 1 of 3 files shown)


NiklasGustafsson:
Three thoughts: 

1) Thank you very much for contributing! I'm impressed by the number of PRs that I returned to after my vacation. :-)

2) When you're adding new functionality, try to see if there's a unit test or two that you can add. When you fix a bug, add a unit test that would have failed with the bug, but passes after the fix.

3) When you add new functionality or fix a bug, add an entry to the RELEASENOTEs.md file, at the top under the next release heading. It makes it easier to highlight the changes when we make a release.

Thanks,

Niklas

GilesBathgate:
@NiklasGustafsson Hope you had a nice vacation. I was holding off doing more until I had feedback on the PR, and also haven't found out how to add unit tests at this stage, so I could use some guidance. I implemented these changes for a work demonstration involving a multi-label classifier using vgg16. Since I needed RandAugment, for that demo, I had already implemented the code before contributing.

NiklasGustafsson:
As long as you don't add new files, adding unit tests is very straight-forward. Just add a new `public void` function with the `[Fact]` attribute to the end of an existing file (try finding one with related tests) in the 'TestTorchSharp' project, call it something reasonable, and you're off to the races.

Also, for regression tests, the file `TestTensorBugs` is where I have collected most of them, even though they are not all about bugs in the `Tensor` code.

NiklasGustafsson:
I wonder if it's as easy as checking (in `Tensor.Dispose()`) whether the associated `grad` Tensor is non-null and disposing it, too...

src/TorchVision/AutoAugment.cs
```diff
@@ -0,0 +1,162 @@
...
+using F = TorchSharp.torchvision.transforms.functional;
+
+#nullable enable
+namespace TorchSharp
+{
+    public static partial class torchvision
+    {
+        private static Tensor apply_op(
+                                Tensor img,
+                                string op_name,
```
NiklasGustafsson, line 14: TorchSharp convention -- we use enums in situations like this.

src/TorchVision/AutoAugment.cs
```diff
@@ -0,0 +1,162 @@
...
+                                float magnitude,
+                                InterpolationMode interpolation,
+                                IList<float>? fill)
+        {
+            if (op_name == "ShearX") {
```
NiklasGustafsson, line 19: Once you have switched to enums, a switch statement may be faster and easier to read.

src/TorchVision/AutoAugment.cs
```diff
@@ -0,0 +1,162 @@
...
+                // Pass
+            } else {
+                throw new ArgumentException($"The provided operator {op_name} is not recognized.");
+            }
+
+            return img;
+        }
+
+
+        internal class RandAugment : ITransform
```
NiklasGustafsson, line 85: I would place the class in the 'Modules' namespace under 'TorchSharp,' which is the pattern we (have tried to) use elsewhere.

src/TorchVision/AutoAugment.cs
```diff
@@ -0,0 +1,162 @@
...
+using System.Linq;
+using static TorchSharp.torch;
+using F = TorchSharp.torchvision.transforms.functional;
+
+#nullable enable
+namespace TorchSharp
+{
+    public static partial class torchvision
+    {
+        private static Tensor apply_op(
```
NiklasGustafsson, line 12: Should this just be a private method in the class? Where else is it used?

src/TorchVision/AutoAugment.cs
```diff
@@ -0,0 +1,162 @@
...
+            } else if (op_name == "AutoContrast") {
+                img = F.autocontrast(img);
+            } else if (op_name == "Equalize") {
+                img = F.equalize(img);
+            } else if (op_name == "Invert") {
+                img = F.invert(img);
+            } else if (op_name == "Identity") {
+                // Pass
+            } else {
+                throw new ArgumentException($"The provided operator {op_name} is not recognized.");
```
NiklasGustafsson, line 78: With an enum, you won't have to throw. Just sayin'... :-)

NiklasGustafsson (changes requested):

src/TorchVision/AutoAugment.cs
GilesBathgate: You don't need to convince me. I much prefer enums, or better still class based dispatch. 
the idea was to copy RandAugment, then update it: https://pytorch.org/vision/main/_modules/torchvision/transforms/autoaugment.html#_apply_op

src/TorchVision/AutoAugment.cs
GilesBathgate: It will be used by `AutoAugment`, `TrivialAugmentWide`, etc...

NiklasGustafsson: Time for a base class with a protected member?

src/TorchVision/AutoAugment.cs
GilesBathgate: I don't think it needs to be exposed under the modules namespace, its `internal`. Other transforms see [Rotate.cs ](https://github.com/dotnet/TorchSharp/blob/main/src/TorchVision/Rotate.cs) follow the pattern I've used here.

NiklasGustafsson: It's easier to find when you're cross-referencing code in VS. 

Not a biggie, but consistency (if you find places where the current code is not) is helpful across a code base.

GilesBathgate: @NiklasGustafsson I think all files under `TorchVision/*.cs` currently! I feel its a separate refactoring issue.

src/TorchVision/AutoAugment.cs
```diff
@@ -0,0 +1,172 @@
...
+                        return F.solarize(img, magnitude);
+                    case opType.AutoContrast:
+                        return F.autocontrast(img);
+                    case opType.Equalize:
+                        return F.equalize(img);
+                    case opType.Invert:
+                        return F.invert(img);
+                    case opType.Identity:
+                        return img; // Pass
+                    default:
```
NiklasGustafsson, line 89: I'm surprised the C# compiler doesn't complain about the fact that you have already covered all the possible cases. The body of the default case is, in effect, unreachable code.

NiklasGustafsson: I take that back -- ShearY is not covered. Is that intentional?

src/TorchVision/AutoAugment.cs
```diff
@@ -0,0 +1,172 @@
...
+                this.num_ops = num_ops;
+                this.magnitude = magnitude;
+                this.num_magnitude_bins = num_magnitude_bins;
+                this.interpolation = interpolation;
+                this.fill = fill;
+            }
+
+            public Tensor call(Tensor img)
+            {
+                var fill = this.fill;
```
NiklasGustafsson, line 113: I would put a dispose scope declaration in this method, and move the result to the outer scope on return.

NiklasGustafsson, line 113: Why the local?

src/TorchVision/AutoAugment.cs
```diff
@@ -0,0 +1,172 @@
...
+                var height = img.shape[cDim + 1];
+                var width = img.shape[cDim + 2];
+                var op_meta = augmentation_space(num_magnitude_bins, (height, width));
+                for (int i = 0; i < num_ops; ++i) {
+                    var op_index = torch.randint(op_meta.Count, new[] { 1 }).ToInt32();
+                    var op_name = op_meta.Keys.ToList()[op_index];
+                    var (magnitudes, signed) = op_meta[op_name];
+                    var magnitude = magnitudes.Dimensions > 0 ? magnitudes[this.magnitude] : 0.0f;
+
+                    if (signed && torch.randint(2, new[] { 1 }).ToBoolean())
```
NiklasGustafsson, line 124: Is it possible to use `stackalloc` instead of `new` here? Probably not.

src/TorchVision/AutoAugment.cs
GilesBathgate: hmm, No, did it never have, or did I remove it in the refactoring I wonder.

src/TorchVision/AutoAugment.cs
```diff
@@ -0,0 +1,172 @@
...
+                        magnitude *= -1.0;
+
+                    img = apply_op(img, op_name, magnitude.ToSingle(), interpolation, fill);
```
NiklasGustafsson, line 127: Since you have to have it as a float, anyway (should it be a double??), isn't it better to do the `ToSingle()` call in the true-branch of the conditional expression on line 122, and leave the local variable a float instead of a Tensor?

src/TorchVision/AutoAugment.cs
NiklasGustafsson, line 133: Hmmm... I wonder if there isn't a cheaper way for us to get a random scalar than this. I'll have to look into that.

NiklasGustafsson: Created https://github.com/dotnet/TorchSharp/issues/1068 to track

src/TorchVision/AutoAugment.cs
GilesBathgate: @NiklasGustafsson Good catch, I've fixed that now. 

BTW I think c# allows you to cast any int to an enum even when that enum doesn't contain that value, maybe that was an old quirk that would emit a warning now.

src/TorchVision/AutoAugment.cs
GilesBathgate: I've no idea. Pytorch has the same. will remove.

BTW I think its a good approach to copy as close to the original implementation line for line and then refactor, even if the original didn't make sense.

src/TorchVision/AutoAugment.cs
```diff
@@ -0,0 +1,181 @@
...
+            private readonly int num_ops;
+            private readonly int magnitude;
+            private readonly int num_magnitude_bins;
+            private readonly InterpolationMode interpolation;
+            private readonly IList<float>? fill;
+        }
+
+        public static partial class transforms
+        {
+            static public ITransform RandAugment(
```
NiklasGustafsson, line 170: Doc comments, of course.

src/TorchVision/AutoAugment.cs
NiklasGustafsson: Definitely. I do the same.

src/TorchVision/AutoAugment.cs
GilesBathgate: Ah, good call. Never came across `stackalloc` before it must be because I am :older_man:

src/TorchVision/AutoAugment.cs
NiklasGustafsson: Oh -- need to say this: for anything with any kind of algorithm in it, I'm always try to be careful to add a link to the original Python file at the top of the file, for attribution as well as reference. 

See TorchVision/Ops/Permute.cs, for example.

src/TorchVision/AutoAugment.cs
GilesBathgate: Yeah, I'll make it double too.

src/TorchVision/AutoAugment.cs
GilesBathgate: Just the wrong parameters here. Fixed.

NiklasGustafsson:
@GilesBathgate. All my comments have been addressed, except the need for some unit tests.

GilesBathgate:
> @GilesBathgate. All my comments have been addressed, except the need for some unit tests.

Yep. Will make some progress on those over the w/end.

NiklasGustafsson:
> > @GilesBathgate. All my comments have been addressed, except the need for some unit tests.
> 
> Yep. Will make some progress on those over the w/end.

Great. While you're at it, add some text at the beginning of the file to point at the Python source code.

Also -- every time you push, a new build starts, but the last one keeps going (it looks like it's been canceled, but it hasn't). It will save some compute resources if you batch up a number of commits into one push.

NiklasGustafsson:
And add a line for this PR and the make_grid PR in the RELASENOTES.md file, under the 0.100.4 release notes at the top.

GilesBathgate:
@NiklasGustafsson If the you think the unit tests are sufficient, then this PR is completed.

NiklasGustafsson:
> @NiklasGustafsson If the you think the unit tests are sufficient, then this PR is completed.

Yup. I'll try to get a clean build, then merge.

NiklasGustafsson (approved):

src/TorchVision/AutoAugment.cs
```diff
@@ -0,0 +1,197 @@
...
+                return img.MoveToOuterDisposeScope();
+            }
+
+            private Dictionary<opType, (Tensor, bool)> augmentation_space(int num_bins, (long height, long width) image_size)
+            {
+                return new Dictionary<opType, (Tensor, bool)> {
+                    { opType.Identity, (torch.tensor(0.0), false) },
+                    { opType.ShearX, (torch.linspace(0.0, 0.3, num_bins), true) },
+                    { opType.ShearY, (torch.linspace(0.0, 0.3, num_bins), true) },
+                    { opType.TranslateX, (torch.linspace(0.0, 150.0 / 331.0 * image_size.height, num_bins), true) },
```
GilesBathgate, line 149: @NiklasGustafsson I think I spotted a bug here, TranslateX should not be image_size.height, it should be image_size.width, TranslateY vice versa.

GilesBathgate: Fixed as part of #1077

PR #1053, opened 2023-08-02
Port torchvision.utils.make_grid and torchvision.utils.save_image from pytorch.

GilesBathgate:
I find the make_grid/save_image functions in torchvision.utils a quick and easy way to save an image batch to png/jpg...etc.

The code might need some additional TorchSharp specific changes.

### DIFF
--- a/src/TorchVision/Utils.cs
+++ b/src/TorchVision/Utils.cs
@@ -121,11 +121,26 @@ namespace TorchSharp
                 double pad_value = 0.0f,
                 Imager imager = null)
             {
+                using var filestream = new FileStream(filename, FileMode.OpenOrCreate);
+                save_image(tensor, filestream, format, nrow, padding, normalize, value_range, scale_each, pad_value, imager);
+            }
+
+            public static void save_image(
+                Tensor tensor,
+                Stream filestream,
+                ImageFormat format,
+                long nrow = 8,
+                int padding = 2,
+                bool normalize = false,
+                (double low, double high)? value_range = null,
+                bool scale_each = false,
+                double pad_value = 0.0f,
+                Imager imager = null)
+            {
                 var grid = make_grid(tensor, nrow, padding, normalize, value_range, scale_each, pad_value);
                 // Add 0.5 after unnormalizing to [0, 255] to round to nearest integer
                 var narr = grid.mul(255).add_(0.5).clamp_(0, 255).to(uint8, CPU);
-                using (var filestream = new FileStream(filename, FileMode.OpenOrCreate))
-                    (imager ?? DefaultImager).EncodeImage(narr, format, filestream);
+                (imager ?? DefaultImager).EncodeImage(narr, format, filestream);
             }
         }
     }
--- a/src/TorchVision/Utils.cs
+++ b/src/TorchVision/Utils.cs
@@ -85,8 +85,10 @@ namespace TorchSharp
                     }
                 }
 
-                if (tensor.size(0) == 1)
-                    return tensor.unsqueeze(0);
+                if (tensor.size(0) == 1) {
+                    tensor = tensor.unsqueeze(0);
+                    return tensor.MoveToOuterDisposeScope();
+                }
 
                 var nmaps = tensor.size(0);
                 var xmaps = Math.Min(nrow, nmaps);
--- a/src/TorchVision/Utils.cs
+++ b/src/TorchVision/Utils.cs
@@ -1,7 +1,6 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
-using TorchSharp;
 using static TorchSharp.torch;
 using static TorchSharp.torchvision.io;
 
@@ -16,9 +15,9 @@ namespace TorchSharp
                 long nrow = 8,
                 int padding = 2,
                 bool normalize = false,
-                (float low, float high)? value_range = null,
+                (double low, double high)? value_range = null,
                 bool scale_each = false,
-                float pad_value = 0.0f)
+                double pad_value = 0.0f)
             {
                 return make_grid(torch.stack(tensors, dim: 0), nrow, padding, normalize, value_range, scale_each, pad_value);
             }
@@ -28,9 +27,9 @@ namespace TorchSharp
                 long nrow = 8,
                 int padding = 2,
                 bool normalize = false,
-                (float low, float high)? value_range = null,
+                (double low, double high)? value_range = null,
                 bool scale_each = false,
-                float pad_value = 0.0f)
+                double pad_value = 0.0f)
             {
                 if (tensor.Dimensions == 2) // Single image H x W
                 {
@@ -51,13 +50,13 @@ namespace TorchSharp
 
                 if (normalize == true)
                 {
-                    void norm_ip(Tensor img, float low, float high)
+                    void norm_ip(Tensor img, double low, double high)
                     {
                         img.clamp_(min: low, max: high);
                         img.sub_(low).div_(Math.Max(high - low, 1e-5));
                     }
 
-                    void norm_range(Tensor t, (float low, float high)? range)
+                    void norm_range(Tensor t, (double low, double high)? range)
                     {
                         if (range.HasValue)
                         {
@@ -88,7 +87,7 @@ namespace TorchSharp
 
                 var nmaps = tensor.size(0);
                 var xmaps = Math.Min(nrow, nmaps);
-                var ymaps = (long)Math.Ceiling((float)nmaps / xmaps);
+                var ymaps = (long)Math.Ceiling((double)nmaps / xmaps);
                 var width = tensor.size(3) + padding;
                 var height = tensor.size(2) + padding;
                 var num_channels = tensor.size(1);
@@ -113,13 +112,13 @@ namespace TorchSharp
             public static void save_image(
                 Tensor tensor,
                 string filename,
-                TorchSharp.torchvision.ImageFormat format,
+                ImageFormat format,
                 long nrow = 8,
                 int padding = 2,
                 bool normalize = false,
-                (float low, float high)? value_range = null,
+                (double low, double high)? value_range = null,
                 bool scale_each = false,
-                float pad_value = 0.0f,
+                double pad_value = 0.0f,
                 Imager imager = null)
             {
                 var grid = make_grid(tensor, nrow, padding, normalize, value_range, scale_each, pad_value);
--- a/src/TorchVision/Utils.cs
+++ b/src/TorchVision/Utils.cs
@@ -31,6 +31,8 @@ namespace TorchSharp
                 bool scale_each = false,
                 double pad_value = 0.0f)
             {
+                using var _ = torch.NewDisposeScope();
+
                 if (tensor.Dimensions == 2) // Single image H x W
                 {
                     tensor = tensor.unsqueeze(0);
@@ -107,7 +109,7 @@ namespace TorchSharp
                     }
                 }
 
-                return grid;
+                return grid.MoveToOuterDisposeScope();
             }
 
             public static void save_image(
@@ -140,8 +142,9 @@ namespace TorchSharp
             {
                 using var grid = make_grid(tensor, nrow, padding, normalize, value_range, scale_each, pad_value);
                 // Add 0.5 after unnormalizing to [0, 255] to round to nearest integer
-                using var narr = grid.mul_(255).add_(0.5).clamp_(0, 255).to(uint8, CPU);
-                (imager ?? DefaultImager).EncodeImage(narr, format, filestream);
+                using var narr0 = grid.mul(255).add_(0.5).clamp_(0, 255);
+                using var narr1 = narr0.to(uint8, CPU);
+                (imager ?? DefaultImager).EncodeImage(narr1, format, filestream);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


src/TorchVision/Utils.cs
```diff
@@ -0,0 +1,133 @@
...
+namespace TorchSharp
+{
+    public static partial class torchvision
+    {
+        public static class utils
+        {
+            public static Tensor make_grid(
+                IEnumerable<Tensor> tensors,
+                long nrow = 8,
+                int padding = 2,
```
GilesBathgate, line 17: @NiklasGustafsson Do you prefer 'long' for most int parameters, and 'double' for most float parameters. I am going by the pytorch type hints.

NiklasGustafsson: I have to confess that the code is not entirely consistent. Double, yes, but the integer size hasn't been very consistent.

src/TorchVision/Utils.cs
```diff
@@ -0,0 +1,132 @@
...
+                ImageFormat format,
+                long nrow = 8,
+                int padding = 2,
+                bool normalize = false,
+                (double low, double high)? value_range = null,
+                bool scale_each = false,
+                double pad_value = 0.0f,
+                Imager imager = null)
+            {
+                var grid = make_grid(tensor, nrow, padding, normalize, value_range, scale_each, pad_value);
```
GilesBathgate, line 124: @NiklasGustafsson If I understand TorchSharp memory management correctly these should be `using var grid` and `using var narr` ?

NiklasGustafsson: Actually, there should be a dispose scope around the function body. There are temporaries being created that will not be disposed, even with a 'using'. Good catch!

src/TorchVision/Utils.cs
```diff
@@ -0,0 +1,132 @@
...
+                long nrow = 8,
+                int padding = 2,
+                bool normalize = false,
+                (double low, double high)? value_range = null,
+                bool scale_each = false,
+                double pad_value = 0.0f)
+            {
+                if (tensor.Dimensions == 2) // Single image H x W
+                {
+                    tensor = tensor.unsqueeze(0);
```
NiklasGustafsson, line 36: put the old tensor in a temp, do the unsqueeze, and then dispose the old tensor:

```C#
using var t = tensor;
tensor = tensor.unsqueeze(0);
```

NiklasGustafsson: Unless, of course, the old tensor is being used outside the function. In that case, create an alias of the input tensor at the beginning, use that throughout, and then have a dispose scope guarding the whole thing. That's probably the easiest way. Make sure to have the call to `.alias()` after you create the dispose scope.

Then, just make sure to call `MoveToOuterDisposeScope()` in the return statements.

NiklasGustafsson: It's safe to ignore my advice, too, since I would expect most calls to be within a dynamic dispose scope, anyway. However, placing on inside algorithmic functions like this has the advantage of giving back native memory eagerly, which can help at the margins.

src/TorchVision/Utils.cs
```diff
@@ -0,0 +1,132 @@
...
+                }
+                if (tensor.Dimensions == 3) // Single image
+                {
+                    if (tensor.size(0) == 1) // Convert single channel to 3 channel
+                    {
+                        tensor = torch.cat(new[] { tensor, tensor, tensor }, 0);
+                    }
+                    tensor = tensor.unsqueeze(0);
+                }
+                if (tensor.Dimensions == 4 && tensor.size(1) == 1) // Single channel images
```
NiklasGustafsson, line 46: What if there's more than one batch dimension?

src/TorchVision/Utils.cs
GilesBathgate: This is why I said 
> The code might need some additional TorchSharp specific changes.

src/TorchVision/Utils.cs
NiklasGustafsson: You're right. The docs require it to be 4D, actually, or a list of 4D tensors.

src/TorchVision/Utils.cs
GilesBathgate: Isn't it just `mul()` that creates a tempory? `add_`, `clamp_` and `to` are inplace ops?

NiklasGustafsson: Yeah, I guess so. Maybe mul_() would be useful here? I can't think of any reason why it couldn't be used.

GilesBathgate: I think I will add a memory unit test:

``c#
using var image = torch.randn(32,3,32,32);
using (var d = torch.NewDisposeScope()) {
    torchvision.utils.save_image(image, ...
    Assert.Equal(d.DisposablesCount, 0);
}
```

That is to say I expect the save_image function to clean up all of its temporaries.

I also expect make_grid to only return one temporary.

Comment (NiklasGustafsson) [reply src/TorchVision/Utils.cs:124]: Yes, that makes sense.

File: src/TorchVision/Utils.cs (hunk 3 above)
Comment (GeorgeS2019) [reply src/TorchVision/Utils.cs:36]: @GilesBathgate 
Thank you and looking forwards to your TorchSharp specific change recommendations

**Comment by NiklasGustafsson** [comment]
@GilesBathgate -- does this one also depend on other PRs of yours?

**Comment by GilesBathgate** [comment]
> @GilesBathgate -- does this one also depend on other PRs of yours?

No I don't think so.

**Comment by NiklasGustafsson** [comment]
Okay, I'll merge this one before the RandomAugment, then. There were changes requested for that one.

**Comment by GilesBathgate** [comment]
Zoiks! I missed a line: `tensor = tensor.clone()  # avoid modifying tensor in-place`

File: test/TorchSharpTest/TestTorchVisionUtils.cs [hunk 5]
```diff
@@ -0,0 +1,20 @@
...
+using Xunit;
+
+namespace TorchSharp
+{
+    public class TestTorchVisionUtils
+    {
+        [Fact]
+        public void Save_Image_TestMemoryUsage()
+        {
+            var imager = new torchvision.io.SkiaImager();
```
GilesBathgate, line 11: @NiklasGustafsson Is there a mock for Imager, or is it ok to use SkiaImager here?

NiklasGustafsson:
From the MacOS build:

```
  Failed TorchSharp.TestTorchVisionUtils.Save_Image_TestMemoryUsage [24 ms]
  Error Message:
   Assert.Equal() Failure
Expected: 0
Actual:   2
  Stack Trace:
     at TorchSharp.TestTorchVisionUtils.Save_Image_TestMemoryUsage() in /Users/runner/work/1/s/test/TorchSharpTest/TestTorchVisionUtils.cs:line 16
Results File: /Users/runner/work/1/s/bin/AnyCPU.Debug/TorchSharpTest/_Mac-1691663692931_2023-08-10_20_10_46.trx
```

GilesBathgate:
> From the MacOS build:

Yeah I am seeing that here too. But it worked when I committed. I think this could be intermittent. Could you take a look @NiklasGustafsson ?

NiklasGustafsson:
> > From the MacOS build:
> 
> Yeah I am seeing that here too. But it worked when I committed. I think this could be intermittent. Could you take a look @NiklasGustafsson ?

There are certainly lots of intermittent errors in testing, which I believe is because tests are run concurrently. This one, as I've seen more build coming through, can't be explained by concurrency and happens in all the builds, now. I'll see if I can pull a branch from your repo and check.

NiklasGustafsson:
@GilesBathgate -- so, I debugged a bit. Don't know how this could have passed.

In this line:

```C#
using var narr0 = grid.mul(255).add_(0.5).clamp_(0, 255);
```

The in-place operators create return an alias for the tensor, which is the safe thing to do, generally speaking, but maybe not what most would consider intuitive. That explains why the pending disposables is '2'.

I need to think about how much of a breaking change it would be to change the behavior of all inplace operators -- it's certainly a big change to make, just mechanically speaking, and it's similar to the requires_grad change that was made earlier this week.

It needs to be done in the native code, as well as the managed code, in order to avoid increasing the reference counter.

NiklasGustafsson:
For now, so that this can be checked in, I suggest filing a bug and adding a 'Skip="<reference the bug number>"' to the [Fact] attribute on the test.

GilesBathgate:
> The in-place operators create return an alias for the tensor, which is a safe thing to do, generally speaking, but maybe not what most would consider intuitive. That explains why the pending disposables is '2'.
> 
> I need to think about how much of a breaking change it would be to change the behavior of all inplace operators -- it's certainly a big change to make, just mechanically speaking, and it's similar to the requires_grad change that was made earlier this week.

Yes intuitively in-place operations shouldn't allocate. I am sure there is a performance benefit also. I think inplace is discouraged in user code, but its used more extensively in the pytorch internals.

NiklasGustafsson:
> For now, so that this can be checked in, I suggest filing a bug and adding a 'Skip="<reference the bug number>"' to the [Fact] attribute on the test.

Alternatively, place a dispose scope in the save_image method and get rid of the usings. That may be the right thing to do in the interim.

NiklasGustafsson:
Just filed issue #1064 to track.

NiklasGustafsson:
One of the return statements in make_grid is missing the 'MoveToOuterDisposeScope()' call, which causes some tests to fail, I believe.

NiklasGustafsson:
```
  Failed TorchSharp.TestTorchVisionUtils.Make_Grid_ColorImageInput [15 ms]
  Error Message:
   Assert.Equal() Failure
           ↓ (pos 0)
Expected: [3, 2, 2]
Actual:   [1, 1, 3, 2, 2]
           ↑ (pos 0)
  Stack Trace:
     at TorchSharp.TestTorchVisionUtils.Make_Grid_ColorImageInput() in /Users/runner/work/1/s/test/TorchSharpTest/TestTorchVisionUtils.cs:line 46
  Failed TorchSharp.TestTorchVisionUtils.Make_Grid_ImageInput [23 ms]
  Error Message:
   Assert.Equal() Failure
           ↓ (pos 0)
Expected: [3, 2, 2]
Actual:   [1, 1, 3, 2, 2]
           ↑ (pos 0)
  Stack Trace:
     at TorchSharp.TestTorchVisionUtils.Make_Grid_ImageInput() in /Users/runner/work/1/s/test/TorchSharpTest/TestTorchVisionUtils.cs:line 34
```

GilesBathgate:
I do: `dotnet build && dotnet test`, I get a pass, then I push. Then later I run it again, and get a fail.... anyway The code is wrong and I don't know why. the shape should be 3,2,2 in both tests. Confirmed against pytorch.

NiklasGustafsson:
Looks like some unsqueeze(0) calls got added somewhere.

GilesBathgate:
Should be `return tensor.squeeze(0)` doh!

NiklasGustafsson:
When I look at the code, I think you're hitting tensor.unsqueeze(0) on line 46, which then means that you also hit line 89 and unsqueeze again.

src/TorchVision/Utils.cs
```diff
@@ -0,0 +1,153 @@
...
+using static TorchSharp.torch;
+using static TorchSharp.torchvision.io;
+
+namespace TorchSharp
+{
+    public static partial class torchvision
+    {
+        public static class utils
+        {
+            public static Tensor make_grid(
```
NiklasGustafsson, line 13: Should have mentioned this earlier -- doc comments, please!

I usually cut-and-paste from the PyTorch docs, capitalizing the first word of the text (since it seems to be PT convention not to).

NiklasGustafsson (changes requested):

test/TorchSharpTest/TestTorchVisionUtils.cs
```diff
@@ -0,0 +1,20 @@
+using System.IO;
+using Xunit;
+
+namespace TorchSharp
```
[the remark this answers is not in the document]
GilesBathgate: Please elaborate, or suggest changes.

src/TorchVision/Utils.cs
GilesBathgate: > By the way, any unit test that writes to a file should make sure to delete it afterwards.

Or use mocks, like MemoryStream.

test/TorchSharpTest/TestTorchVisionUtils.cs
GilesBathgate: No mock framework like NSubstitite or Moq?. It would be quite easy to add a Mock implementation of the interface, but I'll leave that for another time.

NiklasGustafsson: No, not Moq. It's overkill here, I think.

https://www.bleepingcomputer.com/news/security/popular-open-source-project-moq-criticized-for-quietly-collecting-data/

GilesBathgate: Yeah, 'we' heard! NSubstitute seems like a good alternative.

NiklasGustafsson (approved):
Approved. I'll trust that you know better than I whether the algorithm is correct, but all my other concerns have been addressed. I'll merge when I get a clean build.

GilesBathgate:
@NiklasGustafsson It works for me, and the results of the unit tests can be replicated by pytorch.

NiklasGustafsson:
@GilesBathgate -- don't know if you saw it, but this has been merged now.

PR #1055, opened 2023-08-04
Fix cuda synchronize method

cryocz:
The native function binding for `torch.cuda.synchronize` had a mismatched name, which resulted in a link error at runtime when the symbol could not be found. This commit fixes that by renaming the native binding in C# (from `THSCuda_synchronize`) to `THSTorchCuda_synchronize` as defined in the associated header file.

### DIFF
--- a/src/Native/LibTorchSharp/THSTorch.cpp
+++ b/src/Native/LibTorchSharp/THSTorch.cpp
@@ -140,7 +140,7 @@ int THSTorchCuda_device_count()
     return (int)torch::cuda::device_count();
 }
 
-EXPORT_API(void) THSTorchCuda_synchronize(const int64_t device_index)
+void THSTorchCuda_synchronize(const int64_t device_index)
 {
     CATCH(torch::cuda::synchronize(device_index);)
 }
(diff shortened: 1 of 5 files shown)


cryocz:
@dotnet-policy-service agree

NiklasGustafsson (changes requested):
It would make sense to add a unit test that would fail before this fix, and succeed after it.

cryocz:
I've added an act-only unit test for the `torch.cuda.synchronize` method. I'm not entirely sure whether it should be passed over if `cuda.is_available() == false`, and I can't think of a clean way of testing the underlying behavior (except a very hack-y timing check). 

I should also mention that I've bumped into a compilation issue with the native library. Specifically, I've tried to build the project in an official `dotnet/sdk:6.0-alpine` docker image (as I do not have .NET 6 installed on my machine), and I've ran into a compilation issue: in `src/Native/LibTorchSharp/THSTensor.cpp` at lines 389 and 394, the compiler was unable to coerce a `nullptr` into a `float`. I believe those lines use `NULL` by mistake and should probably be changed to use NaN.

NiklasGustafsson:
Thanks! Since this was a linkage error, a regression test need only make sure that the native method can be called.

NiklasGustafsson (approved):

Issue #1046, opened 2023-07-21
Requesting support for Half data type (e.g. creating Tensor)


dje-dev:
Commonly Float16 (represented as Half in .NET) is used as the data type for neural network computations, and the inputs/outputs are in this type. Currently it seems there may not be an efficient/convenient way to provide these inputs to the net (which typically have to be transferred to the GPU). Instead, float must be used (requiring data type conversion overhead on the CPU followed by 2x transfer time to GPU as the 2x wider float type).

Suppose we have prepared our input array
  Half [] nnInput;  
and we want to create a Tensor on the GPU with this data to then run inference. It seems we need a method analogous to this one that is already available, but using the Half type instead of float:
    public static Tensor tensor(float[] rawArray, ScalarType? dtype = null, Device? device = null, bool requires_grad = false, string[]? names = null)

Thank you.

dje-dev:
Yes indeed, many thanks! This will address my most serious pain point. I don't think it addresses the data movement in opposite direction (GPU --> CPU) in half (except for maybe singletons). But we can tackle that in some separate future effort.

PR #1060, opened 2023-08-09
[Enhanced function] Issue #1046

ChengYen-Tang:
https://github.com/dotnet/TorchSharp/issues/1046

### DIFF
--- a/src/TorchSharp/Scalar.cs
+++ b/src/TorchSharp/Scalar.cs
@@ -69,7 +69,7 @@ namespace TorchSharp
             return value.ToScalar();
         }
 
-#if NET5_0_OR_GREATER
+#if NET6_0_OR_GREATER
         /// <summary>
         /// Implicitly convert a .NET scalar value to Scalar
         /// </summary>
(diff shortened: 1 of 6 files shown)


src/TorchSharp/Scalar.cs
```diff
@@ -248,17 +259,18 @@ public static Scalar ToScalar(this bool value)
             return new Scalar(THSTorch_bool_to_scalar(value));
         }
 
-#if NET6_0_OR_GREATER
+#if NET5_0_OR_GREATER
```
NiklasGustafsson, line 262: We no longer support .NET 5

src/TorchSharp/Tensor/Tensor.Factories.cs
```diff
@@ -2104,6 +2103,112 @@ public static Tensor tensor(long[,,,] rawArray, ScalarType? dtype = null, Device
             return _tensor_i64(rawArray, stackalloc long[] { rawArray.GetLongLength(0), rawArray.GetLongLength(1), rawArray.GetLongLength(2), rawArray.GetLongLength(3) }, dtype, device, requires_grad, names: names);
         }
 
+#if NET5_0_OR_GREATER
```
NiklasGustafsson, line 2106: See previous comment about .NET 5

NiklasGustafsson (changes requested):

NiklasGustafsson:
Awesome! Thanks for the help. Only comment is that we no longer support .NET 5, so is should probably be `if NET5_OR_GREATER`

ChengYen-Tang:
Ok, I will fix it tomorrow, I am a little tired today, good night.

NiklasGustafsson (approved):

PR #1062, opened 2023-08-10
Remove apt install dotnet

ChengYen-Tang:
Since Ubuntu CI always fails, script is used UseDotNet@2, there should be no need to use apt install dotnet.🤔

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -34,7 +34,7 @@ jobs:
         sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys DE19EB17684BA42D &&
         sudo apt-get -y update &&
         sudo apt-get -y install clang-6.0 git cmake libunwind8 curl libomp-dev libomp5 &&
-        wget https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb &&
+        wget https://packages.microsoft.com/config/ubuntu/22.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb &&
         sudo dpkg --purge packages-microsoft-prod && sudo dpkg -i packages-microsoft-prod.deb &&
         sudo apt-get update; sudo apt-get install -y apt-transport-https && sudo apt-get update &&
         ldd --version && (/sbin/ldconfig -p | grep stdc++) && (strings /usr/lib/x86_64-linux-gnu/libstdc++.so.6 | grep LIBCXX)
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -36,7 +36,7 @@ jobs:
         sudo apt-get -y install clang-6.0 git cmake libunwind8 curl libomp-dev libomp5 &&
         wget https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb &&
         sudo dpkg --purge packages-microsoft-prod && sudo dpkg -i packages-microsoft-prod.deb &&
-        sudo apt-get update; sudo apt-get install -y apt-transport-https && sudo apt-get update && sudo apt-get install -y dotnet-sdk-6.0 &&
+        sudo apt-get update; sudo apt-get install -y apt-transport-https && sudo apt-get update &&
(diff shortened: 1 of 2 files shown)


ChengYen-Tang:
Why did the build fail 😂
https://dev.azure.com/dotnet/TorchSharp/_build/results?buildId=96964&view=logs&j=80b813b5-9a08-5859-11a8-dc0e5b556e52

NiklasGustafsson:
I'm not sure I would say that Ubuntu never works, it's just become unreliable because of the dotnet install. Oddly, it fails even with this change, but differently.

NiklasGustafsson:
> Why did the build fail 😂 https://dev.azure.com/dotnet/TorchSharp/_build/results?buildId=96964&view=logs&j=80b813b5-9a08-5859-11a8-dc0e5b556e52

From the log:

  Unable to find Clang Compiler
  Install clang-6.0, clang-5.0, or clang-4.0

It just seems like the Ubuntu dependency installation has become pretty unreliable lately. The machines that we build on are pretty underpowered, I believe, and don't have a lot of disk, but Ubuntu shouldn't be a problem, it doesn't require a lot.

ChengYen-Tang:
> > Why did the build fail 😂 https://dev.azure.com/dotnet/TorchSharp/_build/results?buildId=96964&view=logs&j=80b813b5-9a08-5859-11a8-dc0e5b556e52
> 
> From the log:
> 
> ```
>   Unable to find Clang Compiler
>   Install clang-6.0, clang-5.0, or clang-4.0
> ```
> 
> It just seems like the Ubuntu dependency installation has become pretty unreliable lately. The machines that we build on are pretty underpowered, I believe, and don't have a lot of disk, but Ubuntu shouldn't be a problem, it doesn't require a lot.

image

Or can you apply for an agent with this condition😂
https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/scale-set-agents?view=azure-devops

NiklasGustafsson:
It seems like the Ubuntu 22.04 change was removed in the last merge from 'main.' Should it be 22.04 or 20.04 for backward compat? I think there is likely a lot of 20.04 still being used.

NiklasGustafsson (approved):

NiklasGustafsson:
I'm merging this now, since it should marginally help with build stability, but before releasing, we should consider whether '20.04' is a better target than '22.04' -- I think there's a lot of '20.04' still out there, and we don't want to create backward compat issues.

NiklasGustafsson:
After I merge, the build jobs that include the modified Ubuntu tasks fail with:

```
  Unable to find Clang Compiler
  Install clang-6.0, clang-5.0, or clang-4.0
```

It does not apply to the Linux_Native_Build_For_Packages job, which was not modified.

Here's something different from the dependency logs:

Working
```
GLIBCXX_DEBUG_MESSAGE_LENGTH
Warning: apt-key output should not be parsed (stdout is not a terminal)
Executing: /tmp/apt-key-gpghome.twLpNwTseX/gpg.1.sh --keyserver keyserver.ubuntu.com --recv-keys DE19EB17684BA42D
gpg: key FCEE74BB7F3C88C8: public key "Kitware Apt Archive Automatic Signing Key (2021) <debian@kitware.com>" imported
gpg: Total number processed: 1
gpg:               imported: 1
```

Not working:

```
GLIBCXX_DEBUG_MESSAGE_LENGTH
Warning: apt-key output should not be parsed (stdout is not a terminal)
Executing: /tmp/apt-key-gpghome.eCFyVS9aoA/gpg.1.sh --keyserver keyserver.ubuntu.com --recv-keys DE19EB17684BA42D
gpg: keyserver receive failed: Cannot assign requested address
```

azure-pipelines.yml
```diff
@@ -34,9 +34,9 @@ jobs:
         sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys DE19EB17684BA42D &&
         sudo apt-get -y update &&
         sudo apt-get -y install clang-6.0 git cmake libunwind8 curl libomp-dev libomp5 &&
-        wget https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb &&
+        wget https://packages.microsoft.com/config/ubuntu/22.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb &&
```
ChengYen-Tang, line 37: Looks like I shouldn't be modifying this.

PR #1070, opened 2023-08-12
Fix semantic error in THSTensor.cpp

GilesBathgate:
While NULL is defined as 0 which is a valid default value for the float, the default value should be 0.0f

This fixes the compiler warning:
```
TorchSharp/src/Native/LibTorchSharp/THSTensor.cpp(389,5): error G13A67890: converting to non-pointer type ‘float’ from NULL [-Werror=conversion-null] [TorchSharp/src/Native/build.proj]
    389 |     CATCH_RETURN(float, NULL, (float)(tensor->data_ptr<c10::Half>())[i]);
        |     ^~~~~~~~~~~~
```

### DIFF
--- a/src/Native/LibTorchSharp/THSJIT.cpp
+++ b/src/Native/LibTorchSharp/THSJIT.cpp
@@ -514,7 +514,7 @@ void* THSJIT_Type_cast(const JITType type)
     case c10::TypeKind::TensorType:
         return new std::shared_ptr<torch::jit::TensorType>((*type)->cast<c10::TensorType>());
     default:
-        return NULL;
+        return nullptr;
     }
 }
 
--- a/src/Native/LibTorchSharp/THSLoss.cpp
+++ b/src/Native/LibTorchSharp/THSLoss.cpp
@@ -20,7 +20,7 @@ Tensor THSNN_binary_cross_entropy(const Tensor input, const Tensor target, const
     CATCH_RETURN_Tensor(
         auto opts = torch::nn::functional::BinaryCrossEntropyFuncOptions();
         ApplyReduction(opts, reduction);
-        if (weight != NULL)
+        if (weight != nullptr)
             opts = opts.weight(*weight);
         res = ResultTensor(torch::nn::functional::binary_cross_entropy(*input, *target, opts));
     )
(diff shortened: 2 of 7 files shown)


GilesBathgate:
There are some other uses of `NULL` which generally are replaced with nullptr since c++11:

LibTorchSharp/THSNN.cpp:        if (outAsAnyModule != NULL)
LibTorchSharp/THSNN.cpp:        if (outAsAnyModule != NULL)
LibTorchSharp/THSModule.cpp:    if (outAsAnyModule != NULL)
LibTorchSharp/Utils.cpp:thread_local char * torch_last_err = NULL;
LibTorchSharp/THSTensor.cpp:        res = grad.defined() ? new torch::Tensor(grad) : NULL;
LibTorchSharp/THSTensor.cpp:            auto end = (indexEnds == NULL || indexEnds[i] == INT64_MIN) ? c10::optional<c10::SymInt>() : c10::optional<c10::SymInt>(indexEnds[i]);
LibTorchSharp/THSTensor.cpp:            auto step = (indexSteps == NULL || indexSteps[i] == INT64_MIN) ? c10::optional<c10::SymInt>() : c10::optional<c10::SymInt>(indexSteps[i]);
LibTorchSharp/THSTensor.cpp:    CATCH_RETURN(Scalar, NULL, new torch::Scalar(tensor->item()));
LibTorchSharp/THSJIT.cpp:        return NULL;
LibTorchSharp/Utils.h:#define CATCH_RETURN_NNModule(stmt) CATCH_RETURN_RES(NNModule, NULL, stmt)
LibTorchSharp/Utils.h:#define CATCH_RETURN_Tensor(stmt) CATCH_RETURN_RES(Tensor, NULL, stmt)
LibTorchSharp/Utils.h:// Return undefined tensors as NULL to C#
LibTorchSharp/Utils.h:        return NULL;
LibTorchSharp/Utils.h:    if (outAsAnyModule != NULL)
LibTorchSharp/Utils.h:    if (outAsAnyModule != NULL)
LibTorchSharp/THSLoss.cpp:        if (weight != NULL)
LibTorchSharp/THSLoss.cpp:        if (weight != NULL)
LibTorchSharp/THSLoss.cpp:        if (weight != NULL)
LibTorchSharp/THSLoss.cpp:        if (weight != NULL)
LibTorchSharp/THSLoss.cpp:        if (weight != NULL)

These might be good to clean up in this PR.

NiklasGustafsson (approved):

PR #1071, opened 2023-08-12
Use Mock Imager in TorchVisionUtils unit tests

GilesBathgate:
This should make the `Save_Image_TestMemoryUsage` unit test more stable, the following error was observed in the CI build:

```
TorchSharp.TestTorchVisionUtils.Save_Image_TestMemoryUsage [FAIL]
  Failed TorchSharp.TestTorchVisionUtils.Save_Image_TestMemoryUsage [47 ms]
  Error Message:
   System.AccessViolationException : Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
  Stack Trace:
     at SkiaSharp.SKBitmap..ctor()
   at TorchSharp.torchvision.io.SkiaImager.EncodeImage(Tensor image, ImageFormat format, Stream stream) in D:\a\1\s\src\TorchVision\IO\SkiaSharpImager.cs:line 40
   at TorchSharp.torchvision.utils.save_image(Tensor tensor, Stream filestream, ImageFormat format, Int64 nrow, Int32 padding, Boolean normalize, Nullable`1 value_range, Boolean scale_each, Double pad_value, Imager imager) in D:\a\1\s\src\TorchVision\Utils.cs:line 221
   at TorchSharp.TestTorchVisionUtils.Save_Image_TestMemoryUsage() in D:\a\1\s\test\TorchSharpTest\TestTorchVisionUtils.cs:line 15
```

### DIFF
--- a/test/TorchSharpTest.WithCudaBinaries/TorchSharpTest.WithCudaBinaries.csproj
+++ b/test/TorchSharpTest.WithCudaBinaries/TorchSharpTest.WithCudaBinaries.csproj
@@ -55,6 +55,12 @@
   </ItemGroup>
 
   <ItemGroup>
+      <AssemblyAttribute Include="Xunit.CollectionBehavior">
+          <_Parameter1>DisableTestParallelization=true</_Parameter1>
+      </AssemblyAttribute>
+  </ItemGroup>
+
+  <ItemGroup>
     <Content Include="..\TorchSharpTest\linrelu.script.dat" Link="linrelu.script.dat">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
--- a/test/TorchSharpTest.WithCudaBinaries/TorchSharpTest.WithCudaBinaries.csproj
+++ b/test/TorchSharpTest.WithCudaBinaries/TorchSharpTest.WithCudaBinaries.csproj
@@ -57,6 +57,7 @@
   <ItemGroup>
       <AssemblyAttribute Include="Xunit.CollectionBehavior">
           <_Parameter1>DisableTestParallelization=true</_Parameter1>
+          <_Parameter1_IsLiteral>true</_Parameter1_IsLiteral>
       </AssemblyAttribute>
   </ItemGroup>
 
--- a/test/TorchSharpTest/TestTorchVisionUtils.cs
+++ b/test/TorchSharpTest/TestTorchVisionUtils.cs
@@ -1,3 +1,4 @@
+using System;
 using System.IO;
 using Xunit;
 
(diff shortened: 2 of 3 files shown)


test/TorchSharpTest/TestTorchVisionUtils.cs
```diff
@@ -1,18 +1,42 @@
...
         [Fact]
         public void Save_Image_TestMemoryUsage()
         {
-            var imager = new torchvision.io.SkiaImager();
-            using var ms = new MemoryStream();
+            var imager = new MockImager();
             using var image = torch.randn(32, 3, 32, 32);
             using (var d = torch.NewDisposeScope()) {
-                torchvision.utils.save_image(image, ms, torchvision.ImageFormat.Png, imager: imager);
+                torchvision.utils.save_image(image, (Stream)null, torchvision.ImageFormat.Png, imager: imager);
```
NiklasGustafsson, line 39: Explain, please. How does this address the intermittent errors?

GilesBathgate: Well I was going to mock the memory stream too, as the test doesn't require it, but mocking the Imager means the memorystream isn't used.

I can put the MemoryStream back, the real fix is mocking the Imager.

NiklasGustafsson: I was more thinking about the intermittency (is that a word??) of the failures -- they happen all over the place, and fairly randomly. I believe that it's because unit tests are run concurrently, and there are probably non-reentrancy issues in the native code that have yet to be identified. Therefore, the fact that it failed in your new unit test is probably a pure coincidence.

GilesBathgate: Either way, my unit test is not intended to check whether SkiaImager works reliably. I use Mocks to limit the scope of the test, and I find it gives better granularity that way.

GilesBathgate:
> I believe that it's because unit tests are run concurrently, and there are probably non-reentrancy issues in the native code that have yet to be identified.

Do the think the re-entrancy issues are in SkiaImager? Would a ConcurrentSkiaImager be worth implementing since its set globally as io.DefaultImager

NiklasGustafsson:
> > I believe that it's because unit tests are run concurrently, and there are probably non-reentrancy issues in the native code that have yet to be identified.
> 
> Do the think the re-entrancy issues are in SkiaImager? Would a ConcurrentSkiaImager be worth implementing since its set globally as io.DefaultImager

No, there's no reason to believe it has anything to do with SkiaImager.

GilesBathgate:
> No, there's no reason to believe it has anything to do with SkiaImager.

Is there any reason why the SkiaImager SKBitmap.ctor appears in the memory corruption exception then?

NiklasGustafsson:
Coincidence. 

Unless you consistently see the same traceback, there's no reason to think otherwise.

Something, somewhere, corrupts memory because of concurrency issues... then, whatever is running the same time is affected in some random way.

Python is single-threaded, so that environment doesn't put a lot of stress on native code from a concurrency perspective, in my analysis.

test/TorchSharpTest.WithCudaBinaries/TorchSharpTest.WithCudaBinaries.csproj
```diff
@@ -54,6 +54,13 @@
     <Compile Include="..\..\src\TorchSharp\netstandard.cs" />
   </ItemGroup>
 
+  <ItemGroup>
+      <AssemblyAttribute Include="Xunit.CollectionBehavior">
```
GilesBathgate, line 58: Assuming this directive is correct, the tests still fail despite being run sequentially.

NiklasGustafsson: Did you add it now?

NiklasGustafsson: If a test fails consistently in the same way, it's not due to concurrency.

NiklasGustafsson: BTW, I think all of the tests should already be in a collection called 'Sequential'

NiklasGustafsson: Take a look at the failures in the Windows builds in the latest pipeline instance:

https://dev.azure.com/dotnet/TorchSharp/_build/results?buildId=97107&view=results

They're all random failures, none of them involving your unit test.

GilesBathgate: Maybe it runs two assemblies in parallel: https://xunit.net/docs/running-tests-in-parallel#:~:text=Console%20Runner

GilesBathgate:
@NiklasGustafsson This is beyond my understanding. I was hoping to improve unit test stability, but it seems like a wider issue, I've marked this PR as draft.

PR #1077, opened 2023-08-15
Port AutoAugment/TrivialAugmentWide/AugMix from pytorch

GilesBathgate:
Port the remaining AutoAugment classes.


src/TorchVision/Functional.cs
```diff
@@ -24,6 +24,19 @@ public static partial class transforms
...
             {
+                /// <summary>
+                /// Get the image dimensions
+                /// </summary>
+                /// <param name="img">An image tensor.</param>
+                /// <returns>The number of channels, the height and width of the image</returns>
+                public static (long channels, long height, long width) get_dimensions(Tensor img)
+                {
+                    var d = img.Dimensions;
+                    var channels = d == 2 ? 1 : img.shape[d - 3];
```
NiklasGustafsson, line 35: Can it ever be NxHxW, or is it always Nx1xHxW?

GilesBathgate: In PyTorch, this convention is NxCxHxW. so NxHxW, does not fit that convention but Nx1xHxW does.
Anyway implementation is from here:
https://github.com/pytorch/vision/blob/498b9c8662e2322615748aafc321ad4a5bc02afb/torchvision/transforms/v2/functional/_meta.py#L14

NiklasGustafsson:
@GilesBathgate -- just FYI: I'm going to take another week off from work, so if you wonder why I'm ignoring you, that's why.

NiklasGustafsson:
@GilesBathgate -- I'll take a look at this today.

GilesBathgate:
@NiklasGustafsson I am going on holiday next week.

src/TorchVision/AutoAugment.cs
```diff
@@ -9,7 +9,7 @@ namespace TorchSharp
 {
     public static partial class torchvision
     {
-        internal abstract class AutoAugmentBase
+        public abstract class autoaugment
```
NiklasGustafsson, line 12: Why the lower-case class name? Even PyTorch uses capitalization on class names.

src/TorchVision/AutoAugment.cs
```diff
@@ -29,6 +29,12 @@ protected enum opType {
                 Identity
             }
 
+            public enum AutoAugmentPolicy {
```
NiklasGustafsson, line 32: Explain what this is for in a doc comment, please.

src/TorchVision/AutoAugment.cs
```diff
@@ -101,38 +107,206 @@ protected enum opType {
...
+                        { opType.AutoContrast, (torch.tensor(0.0), false)},
+                        { opType.Equalize, (torch.tensor(0.0), false)},
+                        { opType.Invert, (torch.tensor(0.0), false)}
+                    };
+            }
+
+            private IList<((opType, double, int?), (opType, double, int?))> get_policies(AutoAugmentPolicy policy)
+            {
+                switch(policy)
+                {
```
NiklasGustafsson, line 181: Could these lists be made static or memoized, so that there's not an instance created every time the method is called?

src/TorchVision/AutoAugment.cs
```diff
@@ -101,38 +107,206 @@ protected enum opType {
...
+                        { opType.Sharpness, (torch.linspace(0.0, 0.9, num_bins), true)},
+                        { opType.Posterize, (8 - (torch.arange(num_bins) / ((num_bins - 1) / 4)).round().@int(), false)},
+                        { opType.Solarize, (torch.linspace(255.0, 0.0, num_bins), false)},
+                        { opType.AutoContrast, (torch.tensor(0.0), false)},
+                        { opType.Equalize, (torch.tensor(0.0), false)},
+                        { opType.Invert, (torch.tensor(0.0), false)}
+                    };
+            }
+
+            private IList<((opType, double, int?), (opType, double, int?))> get_policies(AutoAugmentPolicy policy)
```
NiklasGustafsson, line 178: For methods that are not a 1:1 correspondence to something in the PyTorch APIs, I try (but I sometimes forget myself) to stay with .NET naming conventions. The only reason for not using .NET conventions is to make it super-easy to move code from Python to C#, so when that's not a motivating reason, better to stick with .NET-style names.

src/TorchVision/AutoAugment.cs
```diff
@@ -101,38 +107,206 @@ protected enum opType {
...
+        internal class RandAugment : autoaugment, ITransform
         {
             public RandAugment(
                 int num_ops = 2,
                 int magnitude = 9,
                 int num_magnitude_bins = 31,
                 InterpolationMode interpolation = InterpolationMode.Nearest,
-                IList<float>? fill = null)
+                IList<float>? fill = null,
+                Generator? generator = null)
```
NiklasGustafsson, line 288: Perfect!

NiklasGustafsson (changes requested):

src/TorchVision/AutoAugment.cs
GilesBathgate: Its a namespace hack, because this class is exposed publicly to give: `torchvision.autoaugment.AutoAugmentPolicy` which matches the convention of PyTorch. There is no AutoAugmentBase class in pytorch. Maybe I can make the base internal again, and use a namespace instead.

src/TorchVision/AutoAugment.cs
NiklasGustafsson: Yes.

NiklasGustafsson:
And, of course -- doc comments on all public APIs.

NiklasGustafsson:
@GilesBathgate -- do you want to keep this PR open?

NiklasGustafsson:
@GilesBathgate -- just reminding you about this PR. It's probably going to be a doozy to merge at this point...

src/TorchVision/AutoAugment.cs
GilesBathgate, line 12:
``` suggestion
// Namespace hack to make use of the baseclass as a namespace for torchvision.autoaugment.AutoAugmentPolicy 
public abstract class autoaugment
```

src/TorchVision/AutoAugment.cs
GilesBathgate, line 32:
```suggestion
            /// <summary>AutoAugment policies learned on different datasets.
            /// Available policies are IMAGENET, CIFAR10 and SVHN.
            /// </summary>
            public enum AutoAugmentPolicy {
```

src/TorchVision/AutoAugment.cs
GilesBathgate: I'll leave this if thats ok. They are all private methods, would the `augmentation_space` functions need to be changed as well?

src/TorchVision/AutoAugment.cs
GilesBathgate: It's called in the constructor. I don't see this as a real issue.

NiklasGustafsson (approved):

PR #1078, opened 2023-08-16
Inplace operators should not create an alias.

NiklasGustafsson:
Also, renamed some unit tests.

This is a work in progress -- the most important operators have been converted, but there are many left to work on.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -4,11 +4,15 @@ Releases, starting with 9/2/2021, are listed with the most recent release at the
 
 ## NuGet Version 0.100.5
 
-__API Changes__:
+__Breaking Changes__:
 
+Inplace operators no longer create an alias, but instead return 'this'. This change will impact any code that explicitly calls `Dispose` on a tensor after the operation.
+
+__API Changes__:
 
 __Bug Fixes__:
 
+#1064 Inplace operators create an alias
 
 ## NuGet Version 0.100.4
 
--- a/src/Native/LibTorchSharp/THSTensor.cpp
+++ b/src/Native/LibTorchSharp/THSTensor.cpp
@@ -450,9 +450,9 @@ Tensor THSTensor_digamma(const Tensor tensor)
     CATCH_TENSOR(tensor->digamma());
 }
 
-Tensor THSTensor_digamma_(const Tensor tensor)
+void THSTensor_digamma_(const Tensor tensor)
 {
-    CATCH_TENSOR(tensor->digamma_());
+    CATCH(tensor->digamma_(););
 }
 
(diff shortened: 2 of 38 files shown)


src/Native/LibTorchSharp/THSTensor.cpp
```diff
@@ -1157,19 +1157,19 @@ Tensor THSTensor_relu(const Tensor tensor)
     CATCH_TENSOR(torch::relu(*tensor));
 }
 
-Tensor THSTensor_relu_(const Tensor tensor)
+void THSTensor_relu_(const Tensor tensor)
 {
-    CATCH_TENSOR(torch::relu_(*tensor));
+    CATCH(torch::relu_(*tensor););
```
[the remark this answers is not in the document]
NiklasGustafsson: Indeed, but not something I'm going to worry about fixing... It was easier to just add 100+ semicolons than to also remove just as many... :-)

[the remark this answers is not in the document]
NiklasGustafsson: Actually, I just took care of it. In most places, at least.

RELEASENOTES.md
```diff
@@ -4,11 +4,15 @@ Releases, starting with 9/2/2021, are listed with the most recent release at the
 
 ## NuGet Version 0.100.5
 
-__API Changes__:
+__Breaking Changes__:
 
+Inplace operators no longer create an alias, but instead return 'this'. This change will impact any code that explicitly calls `Dispose` on a tensor after the operation.
```
GilesBathgate, line 9: Is the expected impact a noop, or exception?

NiklasGustafsson: The expected impact is that if you use dispose scopes, then you don't worry about anything. If you explicitly clean up with calls to `Dispose()` you have to worry about this:

Before: 
```C#
Tensor x = ...'
Tensor y = x.mul_(10);
x.Dispose();
-- y can now be used, the Dispose() call just decremented the native reference counter --
```

After:
```C#
Tensor x = ...'
Tensor y = x.mul_(10);
x.Dispose();
-- y has been disposed, because it's the same managed code instance as X --
```

NiklasGustafsson: You can address this by introducing explicit calls to `.alias()`, as I had to do in a couple of places in the code.

GilesBathgate: It might be worth adding this to [Memory.md](https://github.com/dotnet/TorchSharp/blob/main/docfx/articles/memory.md) and referencing it in the releasenotes?

NiklasGustafsson: Something a little longer will be added to the relevant wiki page once released.

PR #1080, opened 2023-08-17
don't init trace StringBuilder if backend already loaded

### LINKED ISSUE #1079: Unnecessary allocation of StringBuilder
When running an object allocation profiler on my app, I noticed that torch sharp was allocating thousands of `char[16]` and `StringBuilder`. 
allocations

Digging in, I found that they are both being caused by the `LoadNativeBackend` method creating a new `StringBuilder` for every call, even when the backend has already been loaded.

https://github.com/dotnet/TorchSharp/blob/c343981a849b872cae5f6d2c5621e446b9d4c724/src/TorchSharp/Torch.cs#L84-L86

In my fork, I've moved the `new` call inside the if, initialized the out as null and added a null check to the call site. This seems to work without issue, and removes all the allocations I was seeing, resulting in a 30% reduction in GC pauses in my app.
allocations-gone

Happy to submit a PR with the fix.

CodedBeard:
fixes #1079

### DIFF
--- a/src/TorchSharp/Torch.cs
+++ b/src/TorchSharp/Torch.cs
@@ -75,17 +75,18 @@ namespace TorchSharp
             return ok;
         }
 
-        private static void LoadNativeBackend(bool useCudaBackend, out StringBuilder trace)
+        private static void LoadNativeBackend(bool useCudaBackend, out StringBuilder? trace)
         {
             if (!System.Environment.Is64BitProcess) {
                 throw new NotSupportedException("TorchSharp only supports 64-bit processes.");
             }
 
             var alreadyLoaded = useCudaBackend ? nativeBackendCudaLoaded : nativeBackendLoaded;
-            trace = new StringBuilder();
+            trace = null;
+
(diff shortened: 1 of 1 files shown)


src/TorchSharp/Torch.cs
```diff
@@ -272,7 +274,7 @@ public static void InitializeDeviceType(DeviceType deviceType)
                 // If it doesn't we report the entire load trace.
                 var result = cuda.CallTorchCudaIsAvailable();
                 if (!result)
-                    throw new InvalidOperationException($"Torch device type {deviceType} did not initialise on the current machine. Trace from LoadNativeBackend:\n{trace}");
+                    throw new InvalidOperationException($"Torch device type {deviceType} did not initialise on the current machine. Trace from LoadNativeBackend:\n{trace?.ToString() ?? string.Empty}");
```
NiklasGustafsson, line 277: Would it make more sense to simply not have the second sentence in the message if `trace` is null?

NiklasGustafsson (approved):

Issue #1116, opened 2023-10-18
Gradient null for simple calculation


roelofsmn:
I wrote an example using TorchSharp to compute the gradient of a simple function as follows:
```csharp
var device = torch.cuda_is_available() ? torch.CUDA : torch.CPU;

torch.Tensor x1 = torch.tensor(new double[] { 1.0, 3.0 }, device: device, requires_grad: true);

torch.Tensor intercepts = torch.tensor(new double[] { 1.0, 5.0 }, device: device, requires_grad: true);

var x2 = x1 * x1;
var example = x2.sum();
example.backward();
var gradients = x1.grad()?.data<double>().ToArray();
Console.WriteLine($"Gradients at minimum: dy/x1={gradients?[0]}, dy/dx2={gradients?[1]}");
```

However, the gradients variable is `null`.

When I create two variables x1 and x2 as scalar tensors, I can get the expected gradients for each.

Does someone know why the gradients in above code are `null`?

P.S. I used TorchSharp 0.100.5 and TorchSharp-cuda-windows 0.100.5

roelofsmn:
Update: when setting `device = torch.CPU`, everything works as expected.

NiklasGustafsson:
So, when it's on CUDA, it fails, but not on CPU? What happens if you move x1 to CPU before calling `grad()`?

To clarify -- it's the tensor coming out of the call to `grad()` that is null, not just the `gradients` array.

roelofsmn:
Correct. If I do
```
var example = objectiveFunction(x1, intercepts);
example.backward();
x1 = x1.to(torch.CPU);
var gradients = x1.grad()?.data<double>().ToArray();
```
also evaluates `gradients` to null. It is indeed the tensor coming out of `grad()` that is null.

roelofsmn:
I'm now getting the following warning:
```
Warning: The .grad attribute of a Tensor that is not a leaf Tensor is being accessed. Its .grad attribute won't be populated during autograd.backward(). If you indeed want the .grad field to be populated for a non-leaf Tensor, use .retain_grad() on the non-leaf Tensor. If you access the non-leaf Tensor by mistake, make sure you access the leaf Tensor instead. See github.com/pytorch/pytorch/pull/30531 for more informations. (function grad)
```
But what makes a tensor a Leaf tensor? Or, rather, why would `x1` not be a leaf tensor in my example?

NiklasGustafsson:
That doesn't make any sense. 

[...]

PR #1117, opened 2023-10-19
Fixed issue #1116, where gradients were not tracked after moving a tensor to CUDA.

NiklasGustafsson:
The problem was that if a tensor was first created with 'requires_grad' on CPU, the moved to GPU, the gradient tensor would remain on CPU as far as the C++ object was concerned but remained uninitialized and therefore appear as 'null' in C#. 

The fix is not to set `requires_grad` on the tensor until it has already been moved to GPU.

### DIFF
--- a/src/Native/LibTorchSharp/THSTensor.cpp
+++ b/src/Native/LibTorchSharp/THSTensor.cpp
@@ -1796,7 +1796,7 @@ Tensor THSTensor_to_device(const Tensor tensor, const int device_type, const int
 {
     CATCH_RETURN_Tensor(
         auto device = c10::Device((c10::DeviceType)device_type, (c10::DeviceIndex)device_index);
-    res = ResultTensor(tensor->to(device, false, copy));
+        res = ResultTensor(tensor->to(device, false, copy));
     );
 }
 
--- a/src/Native/LibTorchSharp/THSTensorFactories.cpp
+++ b/src/Native/LibTorchSharp/THSTensorFactories.cpp
@@ -153,7 +153,14 @@ Tensor THSTensor_logspace(const double start, const double end, const int64_t st
 }
 
 
-Tensor THSTensor_from_file(const char* filename, const int8_t shared, const int64_t size, const int8_t scalar_type, const int device_type, const int device_index, const bool requires_grad)
+Tensor THSTensor_from_file(
+    const char* filename,
+    const int8_t shared,
+    const int64_t size,
+    const int8_t scalar_type,
+    const int device_type,
+    const int device_index,
+    const bool requires_grad)
 {
     auto options = at::TensorOptions()
         .dtype(at::ScalarType(scalar_type))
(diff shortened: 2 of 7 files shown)


src/Native/LibTorchSharp/THSTensor.cpp
```diff
@@ -1771,9 +1771,9 @@ void THSTensor_topk(const Tensor tensor, Tensor* (*allocator)(size_t length), co
 {
     CATCH(
         auto topk = tensor->topk(k, dim, largest, sorted);
-    Tensor * result = allocator(2);
-    result[0] = new torch::Tensor(std::get<0>(topk));
-    result[1] = new torch::Tensor(std::get<1>(topk));
+        Tensor * result = allocator(2);
```
stephentoub, line 1774: Style nit: just looking at pointer use above, the style seems to be to not have a space before the asterisk, e.g.
```suggestion
        Tensor* result = allocator(2);
```

src/Native/LibTorchSharp/THSTensor.cpp
```diff
@@ -1879,10 +1879,10 @@ void THSTensor_unbind(const Tensor tensor, Tensor* (*allocator)(size_t length),
 {
     CATCH(
         auto res = tensor->unbind(dim);
-    const size_t sz = res.size();
-    Tensor * result = allocator(sz);
-    for (size_t i = 0; i < sz; i++)
-        result[i] = new torch::Tensor(res[i]);
+        const size_t sz = res.size();
+        Tensor * result = allocator(sz);
```
stephentoub, line 1883: Nit: same here?
```suggestion
        Tensor* result = allocator(sz);
```

src/Native/LibTorchSharp/THSTensorFactories.cpp
```diff
@@ -170,43 +177,100 @@ Tensor THSTensor_new(
...
+        .requires_grad(requires_grad && !move);
+
+    CATCH(
+        auto res = torch::from_blob(data, at::ArrayRef<int64_t>(sizes, szlength), deleter, options);
+        if (move) // Not CPU
+        {
+            auto device = c10::Device((c10::DeviceType)device_type, (c10::DeviceIndex)device_index);
+            res = convert
+                ? res.to(device, at::ScalarType(dtype), false, false).set_requires_grad(requires_grad)
+                : res.to(device, false, false).set_requires_grad(requires_grad);
```
stephentoub, line 200: Just a style question. We prefer this over e.g.
```suggestion
            res = convert
                ? res.to(device, at::ScalarType(dtype), false, false)
                : res.to(device, false, false);
            res = res.set_requires_grad(requires_grad);
```

stephentoub (approved):
New to the codebase, but LGTM

src/Native/LibTorchSharp/THSTensorFactories.cpp
NiklasGustafsson: Well, to a certain point (line length being reasonable), I personally like it, feels more functional, but I won't pretend that I'm consistent about it.

src/Native/LibTorchSharp/THSTensor.cpp
NiklasGustafsson: K. Will fix.

PR #1118, opened 2023-10-19
API fixes

NiklasGustafsson:
DataLoader and torch.sparse_coo_tensor()

### DIFF
--- a/src/Examples/CIFAR10.cs
+++ b/src/Examples/CIFAR10.cs
@@ -1,6 +1,8 @@
 // Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
 using System;
 using System.Diagnostics;
+
+using TorchSharp.Modules;
 using static TorchSharp.torch.nn;
 using static TorchSharp.torch.nn.functional;
 using static TorchSharp.torch.utils.data;
@@ -103,8 +105,8 @@ namespace TorchSharp.Examples
             using (Dataset train_data = torchvision.datasets.CIFAR100(datasetPath, true, download: true),
                            test_data = torchvision.datasets.CIFAR100(datasetPath, false, download: true))
             {
-                using var train = new DataLoader(train_data, _trainBatchSize, device: device, shuffle: true);
-                using var test = new DataLoader(test_data, _testBatchSize, device: device, shuffle: false);
+                using var train = DataLoader(train_data, _trainBatchSize, device: device, shuffle: true);
+                using var test = DataLoader(test_data, _testBatchSize, device: device, shuffle: false);
 
                 using (var optimizer = torch.optim.Adam(model.parameters(), 0.001)) {
 
--- a/src/TorchSharp/DataLoader.cs
+++ b/src/TorchSharp/DataLoader.cs
@@ -17,276 +17,352 @@ namespace TorchSharp
         {
             public static partial class data
             {
-                /// <summary>
-                /// This class makes easier to create batch. Data set must implement Dataset interface
-                /// </summary>
-                public class DataLoader : DataLoader<Dictionary<string, torch.Tensor>, Dictionary<string, torch.Tensor>>
+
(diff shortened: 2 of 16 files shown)


src/Examples/CIFAR10.cs
```diff
@@ -103,8 +105,8 @@ internal static void Main(string[] args)
             using (Dataset train_data = torchvision.datasets.CIFAR100(datasetPath, true, download: true),
                            test_data = torchvision.datasets.CIFAR100(datasetPath, false, download: true))
             {
-                using var train = new DataLoader(train_data, _trainBatchSize, device: device, shuffle: true);
-                using var test = new DataLoader(test_data, _testBatchSize, device: device, shuffle: false);
+                using var train = DataLoader(train_data, _trainBatchSize, device: device, shuffle: true);
```
stephentoub, line 108: Trying to follow along with what's going on here, as this doesn't follow typical .NET conventions...

Is DataLoader here a static method defined on one of the types that's already being imported with a using static (I guess TorchSharp.torch.utils.data)? And then the `using TorchSharp.Modules;` is because somewhere in this file `DataLoader` is being used for its type name rather than for this static method name?

src/TorchSharp/DataLoader.cs
```diff
@@ -17,276 +17,352 @@ public static partial class utils
...
-                    /// </param>
-                    public DataLoader(Dataset dataset, int batchSize, bool shuffle = false, Device device = null, int? seed = null, int num_worker = 1, bool drop_last = false)
-                        : base(dataset, batchSize, Collate, shuffle, device, seed, num_worker, drop_last)
-                    {
-                    }
+                public static Modules.DataLoader DataLoader(Dataset dataset, int batchSize, bool shuffle = false, Device device = null, int? seed = null, int num_worker = 1, bool drop_last = false)
+                {
+                    return new Modules.DataLoader(dataset,batchSize,shuffle, device, seed, num_worker,drop_last);
+                }
+                public static Modules.IterableDataLoader DataLoader(IterableDataset dataset, int batchSize, IEnumerable<long> shuffler, Device device = null, int num_worker = 1, bool drop_last = false)
```
stephentoub, line 30: Nit:
```suggestion
                }

                public static Modules.IterableDataLoader DataLoader(IterableDataset dataset, int batchSize, IEnumerable<long> shuffler, Device device = null, int num_worker = 1, bool drop_last = false)
```

src/TorchSharp/DataLoader.cs
```diff
@@ -17,276 +17,352 @@ public static partial class utils
...
+                    var t = cat(dic.Select(k => k[x].unsqueeze(0)).ToArray(), 0);
+                    if (t.device_type != device.type || t.device_index != device.index)
+                        t = t.to(device);
+                    batch[x] = t;
+                }
+                return batch;
+            }
+        }
+
+        public class IterableDataLoader : DataLoader<IList<torch.Tensor>, IList<torch.Tensor>>
```
stephentoub, line 101: Do we generally try to have XML comments on types?

stephentoub (approved):

NiklasGustafsson:
Yes, this pattern is admittedly insane, but it works to make it look more like Python, which is what Don Syme and I decided to do once we realized there was no way to get resources to write tutorials, etc., so staying as close to the Python-based assets was important. We're following the example of the SciSharp org here.

Anyway, there are three specific aspects of C# that we're overcoming in a hacky way:


  1.  You can't put methods in namespaces, so we extensively use nested static classes as modules.

  2.  Constructors need a 'new,' unlike F# or Python, where the type name itself is a function.

  3.  You can't place a method and a class with the same name in the same scope.

So, we move all the types to 'TorchSharp.Modules' and maintain concrete factory methods in the static classes that parallel the PyTorch module hierarchy.

The change here was because the contributor who added DataLoader (that came from the community) didn't follow the otherwise established pattern.


FWIW,

Niklas


Sent from Outlook<http://aka.ms/weboutlook>

src/TorchSharp/DataLoader.cs
NiklasGustafsson: We should. I'll add some.

PR #1128, opened 2023-11-01
Added attribute for custom component names in modules

shaltielshmid:
Added an attribute to allow the users to specify a custom component name instead of the default behavior of using the field name. 
This can be very helpful when using models from PyTorch, and wanting to maintain the naming conventions in each language whilst sharing the state_dict parameter names. 

For example, in python I would write:

```python
self.word_embeddings = nn.Embeddings(30522, 768);
```

And in C# I would write:
```cs
[ComponentName(Name = "word_embeddings")]
private readonly Embedding _wordEmbeddings;
```

This way I can use the C# naming convention but also not have to worry about renaming the parameters when I save the python state_dict.


shaltielshmid:
@dotnet-policy-service agree

shaltielshmid:
I see that the CI failed, on two items:
`Windows_x64_NetFX Release_Build`: All the tests it ran passed, but it seems to have timed out on a network connection, the error is:
```Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host..```
How can I rerun the CI? 

`Ubuntu_x64 Debug_Build`: Azure seemed to indicate that this test has been failing a lot recently, is this possibly a different issue unrelated to my changes?

File: src/TorchSharp/Utils/ComponentNameAttribute.cs [hunk 1]
```diff
@@ -0,0 +1,13 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace TorchSharp.Utils
+{
+    public class ComponentNameAttribute : Attribute
```
NiklasGustafsson, line 9: Doc comments, please!

shaltielshmid: Added!

src/TorchSharp/NN/Module.cs
```diff
@@ -1083,7 +1084,7 @@ protected virtual void RegisterComponents()
 
                     foreach (var field in GetType().GetFields(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance)) {
 
-                        var fieldName = field.Name;
+                        var fieldName = field.GetCustomAttribute<ComponentNameAttribute>()?.Name ?? field.Name;
```
NiklasGustafsson, line 1087: There are 3 other places where the code goes through the fields and does stuff, specifically, when moving a module to/from GPU, or converting to another element type. They need the same treatment.

NiklasGustafsson (changes requested):
Please make sure that all the places where the explicit name is needed are covered.

NiklasGustafsson (dismissed):

NiklasGustafsson:
I will merge this after the 2.1.0 work that I've been doing, but it will go in before the next release later this week.

shaltielshmid:
Great! Thanks!

NiklasGustafsson:
@shaltielshmid -- unit tests, too...

NiklasGustafsson:
Add a line to the release notes under the v0.101.0 heading / API Changes.

shaltielshmid:
Sure, done. 
Added unit tests & release notes.

NiklasGustafsson (approved):

PR #1131, opened 2023-11-01
Added parameter to Module.load to help with non-strict mode

shaltielshmid:
When loading in a `Module` in non-strict mode, there is no indication to how many of the components were loaded and how many were skipped because they weren't found.
Similar to `Module.load_state_dict` which returns two lists of missing/unexpected keys, except here the return value is the Module so I added it as a dictionary parameter, where if it's not null it will be populated with the list of parameters loaded from file, and which of them were matched.


NiklasGustafsson (approved):

NiklasGustafsson:
@shaltielshmid -- I'd like to see some unit tests added for this.

NiklasGustafsson:
Add a line to the release notes under the v0.101.0 heading / API Changes.

shaltielshmid:
Sure. Regarding unit tests - where do you think it would be appropriate to add a test for this?
I think it would be good to add it in the TestLoadSave.cs file, but that file is disabled with a ```#if false.```

NiklasGustafsson:
> Sure. Regarding unit tests - where do you think it would be appropriate to add a test for this? I think it would be good to add it in the TestLoadSave.cs file, but that file is disabled with a `#if false.`

That's where it belongs. Just put it last in the test class, after the `#endif`. The component name tests probably belong in NN.cs

shaltielshmid:
Sounds good, done!

PR #1137, opened 2023-11-05
BugFix: When loading non-strict, make sure to read in the tensor bytes

shaltielshmid:
Added a bug fix where when you load in a model in non-strict mode, if it skips a parameter load it still reads it in so that the BinaryReader can seek forwards to the next tensor. 

NOTE: I also modified the behavior so that if you specify a parameter in the "skip" list, then it no longer requires the dimensions to match. 

I haven't updated the release notes either, what version should I list this bugfix under? 

Also, as part of this feature I wrote a unit test which covers the unit test from the previous PR, so I removed the old one.

### DIFF
--- a/src/TorchSharp/Tensor/TensorExtensionMethods.cs
+++ b/src/TorchSharp/Tensor/TensorExtensionMethods.cs
@@ -377,7 +377,12 @@ namespace TorchSharp
         /// </summary>
         /// <param name="tensor">The tensor into which to load serialized data.</param>
         /// <param name="reader">A BinaryReader instance</param>
-        public static void Load(this Tensor tensor, System.IO.BinaryReader reader)
+        /// <param name="skip">If true, the data will be read from the stream, but not copied to the target tensor.</param>
+        /// <remarks>
+        /// Using a skip list only prevents tensors in the target module from being modified, it
+        /// does not alter the logic related to checking for matching tensor element types.
+        /// </remarks>
+        public static void Load(this Tensor tensor, System.IO.BinaryReader reader, bool skip = false)
         {
             // First, read the type
             var type = (ScalarType)reader.Decode();
@@ -395,7 +400,7 @@ namespace TorchSharp
                 totalSize *= loadedShape[i];
             }
 
-            if (!loadedShape.SequenceEqual(tensor.shape))
+            if (!skip && !loadedShape.SequenceEqual(tensor.shape))
                 // We only care about this if the bytes will be written to the tensor.
                 throw new ArgumentException("Mismatched tensor shape while loading. Make sure that the model you are loading into is exactly the same as the origin.");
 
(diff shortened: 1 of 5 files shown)


GeorgeS2019:
@shaltielshmid : thx you for helping us and your valuable contribution ☺️

NiklasGustafsson:
> I haven't updated the release notes either, what version should I list this bugfix under?

It'll be 0.101.2 -- please add a header for that, as well as updating https://github.com/dotnet/TorchSharp/blob/main/build/BranchInfo.props with the new patch release number.

src/TorchSharp/Tensor/TensorExtensionMethods.cs
```diff
@@ -400,7 +395,7 @@ public static void Load(this Tensor tensor, System.IO.BinaryReader reader, bool
                 totalSize *= loadedShape[i];
             }
 
-            if (!skip && !loadedShape.SequenceEqual(tensor.shape))
+            if (!loadedShape.SequenceEqual(tensor.shape))
                 // We only care about this if the bytes will be written to the tensor.
```
NiklasGustafsson, line 399: I don't understand why the shapes have to match for a skipped tensor. I get that we have to read past it, but that is just about advancing the stream pointer, so the saved size/shape information is sufficient.

shaltielshmid: The reason is because the skip flag is no longer there.

shaltielshmid:
Sounds good, I'll update the version.
One note: My other PR for loading/saving in python format is ready, just waiting for a different package to merge my PR there. Should we release these two items together, and make it a minor version?

src/TorchSharp/Tensor/TensorExtensionMethods.cs
NiklasGustafsson: Okay, but by removing an argument, you just introduced a breaking change, right?

shaltielshmid: Correct. 
Should I revert this part of the change?

NiklasGustafsson:
That one will require a more thorough review, and we usually don't do a release for just one of two PRs.

The 0.101.1 release was an exception, because it addressed a significant enough bug that had blocked several users for quite a while to make it worthwhile.


Sent from Outlook<http://aka.ms/weboutlook>

src/TorchSharp/Tensor/TensorExtensionMethods.cs
NiklasGustafsson: Unless there's a very compelling reason for breaking the API, or there is evidence that the old API could not possibly have been working correctly, we try to avoid doing so.

shaltielshmid: Ok, I'll fix it :)

shaltielshmid:
Ok! I'll update with the changes you mentioned and the version.

shaltielshmid:
Done. Undid breaking change + increased version.

Issue #930, opened 2023-02-20
Revisit: torch. save and torch. load


GeorgeS2019:
> - [ ] [https://github.com/dotnet/TorchSharp/issues/862](https://github.com/dotnet/TorchSharp/issues/862)
> - [ ] [https://github.com/dotnet/TorchSharp/issues/785](https://github.com/dotnet/TorchSharp/issues/785)
> - [ ] [https://github.com/dotnet/TorchSharp/issues/551](https://github.com/dotnet/TorchSharp/issues/551)

The dotnet team: .NET for spark used **Pyrolite** to address [**pickle protocol**](https://github.com/irmen/Pyrolite) to interoperate with python

## Pyrolite

![image](https://user-images.githubusercontent.com/49812372/220086742-0bc9ce3d-ac53-416b-a338-0e6511720d8e.png)

GeorgeS2019:
> Will it be supported in the future?
Python has a [pickle function library](https://docs.python.org/3/library/pickle.html), which can easily save the state of the object
However, this seems to require support from the dotnet team. I think this feature is useful in the field of machine learning

@ChengYen-Tang

ChengYen-Tang:
@GeorgeS2019 
I did a quick read of the code for this library and I don't think it fits our needs. This library is only suitable for data objects.

GeorgeS2019:
@ChengYen-Tang  thx for checking this out. :-)

GeorgeS2019:
@ChengYen-Tang 
@NiklasGustafsson 


https://github.com/irmen/pickle

### Python to .NET (unpickling)

The unpickler simply returns an object. In the case of C#, that is a statically typed
language so you will have to cast that to the appropriate type. Refer to this
table to see what you can expect to receive. Tip: you can use the 'dynamic' type 
in some places to avoid excessive type casting.


PYTHON              | .NET
--------------------|-------------
None                | null
bool                | bool
int                 | int
long                | long (c# doesn't have BigInteger so there's a limit on the size)
string              | string
unicode             | string
complex             | Razorvine.Pickle.Objects.ComplexNumber
datetime.date       | DateTime
datetime.datetime   | DateTime
datetime.time       | TimeSpan
datetime.timedelta  | TimeSpan
float               | double
array.array         | array (all kinds of element types supported)
list                | ArrayList (of objects)
tuple               | object[]
set                 | HashSet<object>
dict                | Hashtable (key=object, value=object)
bytes               | ubyte[]
[...]

PR #1138, opened 2023-11-05
Support loading state_dict from Python saved pickle file

shaltielshmid:
== Please see updated PR summary below ==

Currently as draft until [https://github.com/irmen/pickle/pull/15](https://github.com/irmen/pickle/pull/15) is merged, until then it won't compile. 

Fixes: [https://github.com/dotnet/TorchSharp/issues/930](https://github.com/dotnet/TorchSharp/issues/930)

A few notes  on the loading:
- Only support loading, not saving (for now)
- Added it as separate function `load_py` instead of flag in the load function
- I ignore the device that it was saved from, and I always re-load it to the CPU device. 
- Currently only support loading in the model if it the state_dict was saved using `torch.save(model.state_dict()... )` as opposed to `torch.save(model, ...)` which saves the whole model. 
   => Potentially for the future, we should consider loading in the full model including the requires_grad flag and the rest of the properties. 
- Currently, only support models that were saved using the default `torch.save` (with a zip file), as opposed to models that were saved using the legacy method (with a tar file). 
- Worth testing more on different edge cases - if you have any ideas, please let me know!
- When I reconstruct the tensor, I assign the loaded `bytes` into the tensor. There's an additional parameter there for stride - I assumed that there wasn't a need to look at it, since the tensors are stored the same way. But it's worth noting. 

Question:  In `PyTorchUnpickler.cs` line 99 - I have a `ScalarType` and I want to figure out how many bytes per element. I create an empty tensor using `torch.empty` with that `dtype`, and then call `.element_size()`. Is there a better way to do it?


GeorgeS2019:
@shaltielshmid 

If I understand correctly, this allow loading weights from a standard Pytorch.Save function.

We still need to write the model in c# to receive these weights.

What do you think of e.g. onnx2Torchsharp ( borrowing the idea from onnx2pytorch), reconstructing the model file in c# from an onnx exporter from Pytorch.

Do you see feasibility of reconstruction of Model files in C# from Pytorch.Save .
The other version currently not supported yet by your effort.

GeorgeS2019:
Your proposal is the by pass the need of using 
https://github.com/dotnet/TorchSharp/blob/main/src/Python/exportsd.py

When sharing State Dict from Pytorch to TorchSharp

Currently to share back from TorchSharp back to PyTorch, we need a C# and a python scripts

shaltielshmid:
@GeorgeS2019 Correct - this eliminates the need for the [exportsd.py](https://github.com/dotnet/TorchSharp/blob/main/src/Python/exportsd.py) file. 

i.e., if you have a model in Python:
```python
model = ...
torch.save(model.state_dict(), path)
```

You could then load it into C#:
```cs
model = ...;
model.load_py(path);
```

I am working on also adding the `save` functionality, but it's not as simple. 

Regarding your first message -  I didn't understand the differentiation, what do you mean by `We still need to write the model in c# to receive these weights.`? The weights are automatically loaded into the module.

GeorgeS2019:
The OnnxToTorch does not require one to have model written in pyTorch.

It reconstructw the Pytorch model based on a given onnx with the model structure in addition to weights.

shaltielshmid:
I see. Perhaps in a future PR :)
I wrote this code since I'm developing models in parallel in Python and C#, and wanted the convenience of being able to skip the script.

GeorgeS2019:
@shaltielshmid 

FYI: https://github.com/fel88/Dendrite/issues/11

shaltielshmid:
Updated PR summary:

Currently as draft until [https://github.com/irmen/pickle/pull/15](https://github.com/irmen/pickle/pull/15) is merged, until then it won't compile. 

Fixes: [https://github.com/dotnet/TorchSharp/issues/930](https://github.com/dotnet/TorchSharp/issues/930)

A few notes on the feature:
- Both loading and saving are supported.
- Added it as separate function `load_py` and `save_py` instead of flag in the load/save function
- I ignore the device that it was saved from, and I always re-load / re-save it to the CPU device. 
- Currently only support loading in the model if it the state_dict was saved using `torch.save(model.state_dict()... )` as opposed to `torch.save(model, ...)` which saves the whole model. 
   => Potentially for the future, we should consider loading in the full model including the requires_grad flag and the rest of the properties. 
- Currently, only support models that were saved using the default `torch.save` (with a zip file), as opposed to models that were saved using the legacy method (with a tar file). 
- Worth testing more on different edge cases - if you have any ideas, please let me know!

Question 1:  In `PyTorchUnpickler.cs` line 99 - I have a `ScalarType` and I want to figure out how many bytes per element. I create an empty tensor using `torch.empty` with that `dtype`, and then call `.element_size()`. Is there a better way to do it?

Question 2: I want to add a unit test to make sure the `save_py` function works, but that requires making sure the pytorch package is installed in the system and running a python script to check that it's loaded. Do you have an alternative suggestion?

docfx/articles/saveload.md
```diff
@@ -46,24 +42,32 @@ For efficient memory management, the model should be created on the CPU before l
 
 ><br/>It is __critical__ that all submodules and buffers in a custom module or composed by a Sequential object have exactly the same name in the original and target models, since that is how persisted tensors are associated with the model into which they are loaded.<br/><br/>The CustomModule 'RegisterComponents' will automatically find all fields that are either modules or tensors, register the former as modules, and the latter as buffers. It registers all of these using the name of the field, just like the PyTorch Module base class does.<br/><br/>
 
-### Saving a TorchSharp format model in Python
+### How to use the PyTorch format
```
NiklasGustafsson, line 45: It looks like you're replacing the TorchSharp-specific instructions pickling-based instructions. I would advice you to add rather than replace.

src/TorchSharp/TorchSharp.csproj
```diff
@@ -29,6 +29,7 @@
 
   <ItemGroup>
     <PackageReference Include="Google.Protobuf" Version="3.21.9" />
+    <PackageReference Include="Razorvine.Pickle" Version="1.4.0" />
```
NiklasGustafsson, line 32: If we're introducing a new dependency, we're going to have to take a look at its license, etc.

docfx/articles/saveload.md
NiklasGustafsson: We also want to make sure that we know it's something that has at least the same level of support that TorchSharp does.

src/TorchSharp/Utils/PyTorchPickler.cs
```diff
@@ -0,0 +1,217 @@
...
+                    torch.ScalarType.Byte => "ByteStorage",
+                    torch.ScalarType.Bool => "BoolStorage",
+                    torch.ScalarType.BFloat16 => "BFloat16Storage",
+                    torch.ScalarType.ComplexFloat64 => "ComplexDoubleStorage",
+                    torch.ScalarType.ComplexFloat32 => "ComplexFloatStorage",
+                    _ => throw new NotImplementedException()
+                };
+            }
+        }
+        class Storage
```
NiklasGustafsson, line 138: Even though many of these classes are private, it would be useful to have comments explaining their role. We don't systematically do that in this repo, but it would be useful.

test/TorchSharpTest/TorchSharpTest.csproj
```diff
@@ -21,6 +21,7 @@
     <None Remove="adamw1.dat" />
     <None Remove="asgd1.dat" />
     <None Remove="bug510.dat" />
+    <None Remove="pyload_test.model" />
```
NiklasGustafsson, line 24: It also needs to be added to the CUDA version of the test project (by link, not copy).

docfx/articles/saveload.md
shaltielshmid: Sure, of course. 
This feature is meant to bypass the need for the `importsd.py` and `exportsd.py`, so I had replaced the instructions that used those scripts. 
I'll rewrite the changes so that I'm adding and leaving in instructions for using the script for conversion and saving using the (new) built-in `module.save_py` and `module.load_py` methods.

src/TorchSharp/TorchSharp.csproj
NiklasGustafsson: Definitely not -- it's not nice to the author of the package, who deserves to have dependents noted, and it's essentially means forking, which has software engineering implications.

src/TorchSharp/Utils/PyTorchPickler.cs
shaltielshmid: Of course, I was working on that now :)

test/TorchSharpTest/TorchSharpTest.csproj
shaltielshmid: Thanks! Will do!

docfx/articles/saveload.md
NiklasGustafsson: > We also want to make sure that we know it's something that has at least the same level of support that TorchSharp does.

This was meant as a response to my question about the new dependency, not the article..

shaltielshmid: > > We also want to make sure that we know it's something that has at least the same level of support that TorchSharp does.
> 
> This was meant as a response to my question about the new dependency, not the article..

Ah! Now I understand.
It seems to be up to date - supporting the latest pickle protocol, but I'll leave that judgment up to you.

src/TorchSharp/TorchSharp.csproj
shaltielshmid: Agreed :)

shaltielshmid:
@NiklasGustafsson I wanted to hear your thoughts on these two questions, I wasn't sure how to implement them:

Question 1: In PyTorchUnpickler.cs line 99 - I have a ScalarType and I want to figure out how many bytes per element. I create an empty tensor using torch.empty with that dtype, and then call .element_size(). Is there a better way to do it?

Question 2: I want to add a unit test to make sure the save_py function works, but that requires making sure the pytorch package is installed in the system and running a python script to check that it's loaded. Do you have a suggestion on how to test it out?
Right now, I'm thinking of just testing whether the saved file can be loaded using the added `load_py` method.

NiklasGustafsson:
>> Is there a better way to do it?

No, but it seems like a useful addition to 'ScalarExtensionMethods' in Scalar.cs -- a SizeOf() function that takes the ScalarType as input and then returns the size from a big switch-statement.

NiklasGustafsson:
>> Do you have a suggestion on how to test it out?

I would add a file that's been verified by you as the correct output, and then do a binary comparison between what the unit test produces and the stored reference file, since the output should remain deterministic as long as the unit test doesn't change.

If you add more than one test, make sure to avoid file name collisions -- multiple test projects may run in parallel, and individual tests may run concurrently, too. I've added a function to generate random file names for that purpose.

NiklasGustafsson:
BTW, @shaltielshmid, would you mind adding an email to your public profile, so I can message you privately?

NiklasGustafsson:
All builds fail with:

```
D:\a\1\s\src\TorchSharp\Utils\PyTorchPickler.cs(225,47): error CS0246: The type or namespace name 'IObjectDeconstructor' could not be found (are you missing a using directive or an assembly reference?) [D:\a\1\s\src\TorchSharp\TorchSharp.csproj]
D:\a\1\s\src\TorchSharp\Utils\PyTorchPickler.cs(250,44): error CS0246: The type or namespace name 'IObjectDeconstructor' could not be found (are you missing a using directive or an assembly reference?) [D:\a\1\s\src\TorchSharp\TorchSharp.csproj]
D:\a\1\s\src\TorchSharp\Utils\PyTorchPickler.cs(85,37): error CS0115: 'PyTorchPickler.CustomPickler.persistentId(object, out object)': no suitable method found to override [D:\a\1\s\src\TorchSharp\TorchSharp.csproj]
D:\a\1\s\src\TorchSharp\Utils\PyTorchUnpickler.cs(87,39): error CS0115: 'PyTorchUnpickler.CustomUnpickler.persistentLoad(object)': no suitable method found to override [D:\a\1\s\src\TorchSharp\TorchSharp.csproj]
D:\a\1\s\src\TorchSharp\Utils\PyTorchPickler.cs(225,47): error CS0246: The type or namespace name 'IObjectDeconstructor' could not be found (are you missing a using directive or an assembly reference?) [D:\a\1\s\src\TorchSharp\TorchSharp.csproj]
D:\a\1\s\src\TorchSharp\Utils\PyTorchPickler.cs(250,44): error CS0246: The type or namespace name 'IObjectDeconstructor' could not be found (are you missing a using directive or an assembly reference?) [D:\a\1\s\src\TorchSharp\TorchSharp.csproj]
D:\a\1\s\src\TorchSharp\Utils\PyTorchPickler.cs(85,37): error CS0115: 'PyTorchPickler.CustomPickler.persistentId(object, out object)': no suitable method found to override [D:\a\1\s\src\TorchSharp\TorchSharp.csproj]
D:\a\1\s\src\TorchSharp\Utils\PyTorchUnpickler.cs(87,39): error CS0115: 'PyTorchUnpickler.CustomUnpickler.persistentLoad(object)': no suitable method found to override [D:\a\1\s\src\TorchSharp\TorchSharp.csproj]
```

shaltielshmid:
Yes, it won't compile until my [PR](https://github.com/irmen/pickle/pull/15) is merged and a new version is released for the pickle library.

NiklasGustafsson:
> Yes, it won't compile until my [PR](https://github.com/irmen/pickle/pull/15) for that library is merged and a new version is released for the pickle library.

I see. In that case, it would be good to hold off pushing commits go GH, or create another branch that's not associated with the PR, until it's in a shape to compile. Even though your PR is a Draft, the CI/CD system will be triggered and leave a lot of failing builds in Azure Pipelines.

shaltielshmid:
Sounds good - I'll continue committing on a separate branch and then fast forward this branch when it's ready.

shaltielshmid:
@NiklasGustafsson Following our discussion, an extension methods package for TorchSharp was now published on NuGet allowing loading & saving PyTorch format models from TorchSharp: https://www.nuget.org/packages/TorchSharp.PyBridge.

I'm closing this pull request as it is not relevant anymore. 

Should I create a new PR with a change to the saveload doc pointing to the TorchSharp.PyBridge library?

GeorgeS2019:
Additional discussion

[TorchSharp.PyBridge](https://www.nuget.org/packages/TorchSharp.PyBridge)

[Reference](https://github.com/irmen/pickle/issues/12#issuecomment-1817889286)

PR #1150, opened 2023-11-21
Added section in Save/Load readme for TorchSharp.PyBridge

shaltielshmid:
I added a paragraph in the Wiki referencing the package. 
Please let me know if you want me to revise it in any way.


NiklasGustafsson (approved):

NiklasGustafsson:
Does PyBridge also support saving/loading optimizer state, or just modules?

shaltielshmid:
Right now only Modules. I'm looking into loading Optimizer states. Would you like me to revisit this PR after I get it working with Optimizer states?

NiklasGustafsson:
You should come back with a new PR for that, but please do so.

shaltielshmid:
@NiklasGustafsson Thanks for adding it into the wiki! I think there was a slight mishap in the wiki and the markdown for the added section and the one after it only got partially rendered. See here:

image

NiklasGustafsson:
Fixed

PR #1151, opened 2023-11-22
Added state index ref to optimizer state dict

shaltielshmid:
While adding the functionality for saving the optimizer state_dict in the pytorch format, I encountered an issue that I couldn't resolve externally. 

When looking at an optimizer state_dict in Python, they have an index mapping between the parameter_group and the state group. 

E.g., see here a sample state_dict of an Adam optimizer:

```json
{
  "state": {
      0: {
        "step": tensor(1.),
        "exp_avg": tensor(...),
        "exp_avg_seq": tensor(...)
      },
      1: {
        "step": tensor(1.),
        "exp_avg": tensor(...),
        "exp_avg_seq": tensor(...)
      },
   },
  "param_groups": [
    {
      "lr": 0.001,
      "betas": [
        0.8,
        0.9
      ],
      "eps": 1e-08,
      "weight_decay": 0,
      "amsgrad": false,
      "maximize": false,
      "foreach": null,
      "capturable": false,
      "differentiable": false,
      "fused": null,
      "params": [
        0,
        1
      ]
    }
  ]
}
```

For TorchSharp, there is no need for the params, but PyTorch requires the Params object which maps between the params and the states to be included. 

Therefore, I propose this addition which will include the mapping between parameters the state, without inserting any breaking changes.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -4,6 +4,13 @@ Releases, starting with 9/2/2021, are listed with the most recent release at the
 
 ## NuGet Version 0.101.3
 
+__Breaking Changes__:
+
+The base `OptimizerState` class was modified and includes two changes:
+
+1. Custom optimizer state objects derived from `OptimizerState` must now explicitly pass the related `torch.nn.Parameter` object to the `OptimizerState` base constructor to maintain correct linkage.
+2. Custom state objects must implement an Initialize function. This function is responsible for initializing the properties of the state. Note that this function can be called as a re-intialization, so proper disposal of the previous tensor objects should be handled.
+
 __API Changes__:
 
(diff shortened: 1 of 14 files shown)


NiklasGustafsson:
Since you're digging around (deep) in this -- would you mind attacking #1153 while you're at it?

shaltielshmid:
Sure! I think this relates directly to what I emailed you about on Thursday, figuring out a way to reinitialize a state dynamically.

shaltielshmid:
Yes! Working on that + Release notes.

shaltielshmid:
@NiklasGustafsson done - I added a check in `exportsd.py` and updated the release notes.

NiklasGustafsson (approved):

PR #1152, opened 2023-11-22
Implement LayerNorm module in terms of functional.layer_norm.

GilesBathgate:
This is essentially to allow the bias parameter, which setting to false allows computing layernorms without a bias.

### DIFF
--- a/src/TorchSharp/NN/Normalization/LayerNorm.cs
+++ b/src/TorchSharp/NN/Normalization/LayerNorm.cs
@@ -63,7 +63,15 @@ namespace TorchSharp
                 return F.layer_norm(tensor, _normalized_shape, weight, bias, _eps);
             }
 
-            public Parameter? bias { get; set; }
+            public Parameter? bias {
+                get => _bias;
+                set {
+                    _bias?.Dispose();
+                    _bias = value?.DetachFromDisposeScope() as Parameter;
+                    ConditionallyRegisterParameter(nameof(bias), _bias);
+                }
+            }
+            private Parameter? _bias;
 
             public Parameter weight {
                 get => _weight!;
--- a/src/TorchSharp/NN/Normalization/LayerNorm.cs
+++ b/src/TorchSharp/NN/Normalization/LayerNorm.cs
@@ -42,7 +42,6 @@ namespace TorchSharp
                     register_parameter("bias", null);
                 }
 
-                RegisterComponents();
 
                 reset_parameters(elementwise_affine);
             }
--- a/src/TorchSharp/NN/Normalization/LayerNorm.cs
+++ b/src/TorchSharp/NN/Normalization/LayerNorm.cs
@@ -1,57 +1,72 @@
 // Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
 using System;
 using static TorchSharp.torch;
+using static TorchSharp.torch.nn;
 using static TorchSharp.PInvoke.NativeMethods;
 
 #nullable enable
 namespace TorchSharp
 {
     using Modules;
+    using F = TorchSharp.torch.nn.functional;
 
     namespace Modules
     {
-
         /// <summary>
         /// This class is used to represent a LayerNorm module.
         /// </summary>
         public sealed class LayerNorm : torch.nn.Module<Tensor, Tensor>
         {
-            internal LayerNorm(IntPtr handle, IntPtr boxedHandle) : base(handle, boxedHandle)
+            private long[] _normalized_shape;
+            private double _eps;
+
+            internal LayerNorm(long[] normalized_shape, double eps, bool elementwise_affine, bool bias, Device? device, ScalarType? dtype) : base(IntPtr.Zero, IntPtr.Zero)
             {
+                _normalized_shape = normalized_shape;
+                _eps = eps;
+                if (elementwise_affine)
+                {
+                    weight = Parameter(torch.empty(normalized_shape, dtype, device));
+                    if (bias)
+                    {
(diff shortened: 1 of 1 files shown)


src/TorchSharp/NN/Normalization/LayerNorm.cs
```diff
@@ -1,57 +1,72 @@
...
+                    init.zeros_(bias);
                 }
             }
+
+            public override Tensor forward(Tensor tensor)
+            {
+                return F.layer_norm(tensor, _normalized_shape, weight, bias, _eps);
+            }
+
+            public Parameter? bias { get; set; }
```
NiklasGustafsson, line 67: This is likely to cause trouble -- the backing field is going to appear as a parameter, which means that you shouldn't call `RegisterComponents()` in this one.

I have a (not quite abandoned) branch in my private fork that demonstrates the pattern to use:

https://github.com/NiklasGustafsson/TorchSharp/blob/unit/src/TorchSharp/NN/Linear.cs

src/TorchSharp/NN/Normalization/LayerNorm.cs
```diff
@@ -1,57 +1,72 @@
...
-                    THSNN_LayerNorm_set_bias(handle, (value is null ? IntPtr.Zero : value.Handle));
-                    torch.CheckForErrors();
-                    ConditionallyRegisterParameter("bias", value);
+                else
+                {
+                    register_parameter("weight", null);
+                    register_parameter("bias", null);
                 }
+
+                RegisterComponents();
```
GilesBathgate, line 45:
```suggestion

```

src/TorchSharp/NN/Normalization/LayerNorm.cs
```diff
@@ -1,57 +1,72 @@
...
             }
+
+            public override Tensor forward(Tensor tensor)
+            {
+                return F.layer_norm(tensor, _normalized_shape, weight, bias, _eps);
+            }
+
+            public Parameter? bias { get; set; }
+
+            public Parameter? weight { get; set; }
```
GilesBathgate, line 69:
```suggestion
            public Parameter weight {
                get => _weight!;
                set {
                    if (value is null) throw new ArgumentNullException(nameof(weight));
                    if (value.Handle != _weight?.Handle) {
                        _weight?.Dispose();
                        _weight = (value.DetachFromDisposeScope() as Parameter)!;
                        ConditionallyRegisterParameter(nameof(weight), _weight);
                    }
                }
            }

            private Parameter? _weight;
```

src/TorchSharp/NN/Normalization/LayerNorm.cs
GilesBathgate, line 67:
```suggestion
            public Parameter? bias {
                get => _bias;
                set {
                    _bias?.Dispose();
                    _bias = value?.DetachFromDisposeScope() as Parameter;
                    ConditionallyRegisterParameter(nameof(bias), _bias);
                }
            }
            private Parameter? _bias;
```

GilesBathgate (commented):
@NiklasGustafsson So I think you are suggesting this?

src/TorchSharp/NN/Normalization/LayerNorm.cs
NiklasGustafsson: Yes -- cut-and-paste from the modified Linear.cs, and don't call RegisterComponents. 

If you want to be really efficient, then override the three `_to()` overloads and implement them directly, since you know what the parameters are (no need to call `parameters()`) That's optional, though.

GilesBathgate:
> I don't know this off the top of my head -- are there unit tests for LayerNorm?

I think it makes sense to add tests for the code I've changed, i.e verifying that we are calling `F.layer_norm` with the correct parameters, But really `F.layer_norm` should already have tests. I couldn't find them.

NiklasGustafsson:
> I couldn't find them.

That doesn't surprise me. There are, still, functionality without unit tests in the repo. If you wouldn't mind adding something, that would be much appreciated.

src/TorchSharp/NN/Normalization/LayerNorm.cs
```diff
@@ -1,57 +1,91 @@
...
+                    else
+                    {
+                        register_parameter("bias", null);
+                    }
+                }
+                else
+                {
+                    register_parameter("weight", null);
+                    register_parameter("bias", null);
+                }
```
GilesBathgate, line 43:
```suggestion

```

src/TorchSharp/NN/Normalization/LayerNorm.cs
```diff
@@ -1,57 +1,91 @@
...
+                {
+                    weight = Parameter(torch.empty(normalized_shape, dtype, device));
+                    if (bias)
+                    {
+                        this.bias = Parameter(torch.empty(normalized_shape, dtype, device));
+                    }
+                    else
+                    {
+                        register_parameter("bias", null);
+                    }
```
GilesBathgate, line 37:
```suggestion

```

GilesBathgate (commented):
To remove.

NiklasGustafsson:
Apparently, GH is having trouble with PRs at the moment:

https://www.githubstatus.com/

NiklasGustafsson (approved):

GilesBathgate:
@NiklasGustafsson Regarding the unit test(s) what would we do for expected output vs input, should we just hard code some random values, run it through pytorch, and see what the output should be?

NiklasGustafsson:
Random or specifically picked values... Like 'arange(N).to(torch.float32).reshape(...)' and then see what PT produces.

NiklasGustafsson:
I'm going to merge this. Unit tests can wait for now.

Issue #1154, opened 2023-11-22
mu_product in NAdam optimizer is never initialized


shaltielshmid:
The `mu_product` field in the NAdam state is never initialized and therefore defaults to zero. See here the initialization of the state (I also searched the rest of the code and there was no other initialization):
![image](https://github.com/dotnet/TorchSharp/assets/20192446/1cde85eb-907f-4946-b371-4a2dc14c4289)

The `mu_product` by definition is always going to be the product of other multiplications, and therefore will always remain 0. 

In the `PyTorch` documentation, you can see that in the source code for [NAdam](https://pytorch.org/docs/stable/_modules/torch/optim/nadam.html#NAdam) in the `_init_group` function they initialize the `mu_product` to 1:

![image](https://github.com/dotnet/TorchSharp/assets/20192446/6c1e893a-7765-4fee-aee8-512accf042fd)

shaltielshmid:
Solved with https://github.com/dotnet/TorchSharp/pull/1155

PR #1155, opened 2023-11-22
Updated mu_product initialization

shaltielshmid:
Following the issue [here](https://github.com/dotnet/TorchSharp/issues/1154), the mu_product is now initialized to 1.

### DIFF
--- a/src/TorchSharp/Optimizers/NAdam.cs
+++ b/src/TorchSharp/Optimizers/NAdam.cs
@@ -175,7 +175,7 @@ namespace TorchSharp
                         var mu_next = beta1 * (1.0 - 0.5 * Math.Pow(0.96, (state.step + 1) * momentum_decay));
 
                         var mu_product = state.mu_product * mu;
-                        var mu_product_next = mu_product * mu * mu_next;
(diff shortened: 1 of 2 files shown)


shaltielshmid:
In addition to the bug with mu_product never being initialized, I discovered that `mu_product_next` wasn't being calculated like in PyTorch.
In TorchSharp: `var mu_product_next = mu_product * mu * mu_next;` [See here](https://github.com/dotnet/TorchSharp/blob/main/src/TorchSharp/Optimizers/NAdam.cs#L178)
In PyTorch: `mu_product_next = mu_product * mu_next`  [See here](https://github.com/pytorch/pytorch/blob/HEAD/torch/optim/nadam.py#L322)

Those two changes resulted in the tests failing due to the optimization calculation being changed. I updated the scores, and confirmed that they match the scores from PyTorch. Attaching code to run the NAdam tests in Python [here](https://www.dropbox.com/scl/fi/duonta6d34i3yat2v5ris/testnadam.py?rlkey=rd6hlt45acsavzy7rdd3x3b12).

PR #1156, opened 2023-11-26
Added tensor extension ToArray

shaltielshmid:
When using the library, I encountered a need to get the actual content from the Tensor to regular C# arrays, so I wrote this extension - and I thought it could be a useful addition to the library.
As opposed to the .tolist function which is generic and works for any type, this is a strongly typed function which allows you to get the data easily in a strongly typed array without any type checks.

### DIFF
--- a/src/TorchSharp/Tensor/TensorExtensionMethods.cs
+++ b/src/TorchSharp/Tensor/TensorExtensionMethods.cs
@@ -699,9 +699,12 @@ namespace TorchSharp
             if (torch.ToScalarType(typeof(T)) != tensor.dtype)
                 throw new ArgumentException("Tensor type different from requested array type");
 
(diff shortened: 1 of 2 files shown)


NiklasGustafsson (commented):
Doesn't `t.data<T>().ToArray()` already do this?

shaltielshmid:
Aha! Yes it does. I didn't know about that. 
Is it worth keeping for the 2d arrays or should I discard this PR?

There is an advantage of speed here since the array is copied without indexing, but I don't think it is that important.

NiklasGustafsson:
Unless there's a specific ask for 2D, I would abandon it.

shaltielshmid:
I've been using it this week in my code (e.g., when I want to get the prediction scores for a whole batch), but I'll just write a local helper function.

PR #1163, opened 2023-11-28
Add unit test for LayerNorm

GilesBathgate:
Here is the simple unit test I wrote, for LayerNorm. Not able to run this locally at this time, so no idea if it works. Please let me know if it works for you @NiklasGustafsson

### DIFF
--- a/test/TorchSharpTest/TestNormalize.cs
+++ b/test/TorchSharpTest/TestNormalize.cs
@@ -45,7 +45,7 @@ namespace TorchSharp.Tests
 
             var layerNorm = torch.nn.LayerNorm(normalized_shape, eps, true, true);
 
-            output = layerNorm.forward(weight);
+            output = layerNorm.forward(input_data);
 
             Assert.True(torch.allclose(output, precomputedOutput, atol: 1e-4));
         }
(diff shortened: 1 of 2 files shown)


NiklasGustafsson:
OK, I'll try it.

NiklasGustafsson:
I had to modify the test to this:

```C#
            var layerNorm = torch.nn.LayerNorm(normalized_shape, eps, true, true);
            layerNorm.weight = new Parameter(weight);
            layerNorm.bias = new Parameter(bias);

            output = layerNorm.forward(input_data);

            Assert.True(torch.allclose(output, precomputedOutput, atol: 1e-4));
```

in order for it to pass.

test/TorchSharpTest/TestNormalize.cs
```diff
@@ -0,0 +1,54 @@
...
+            var weight = torch.arange(1, 21, dtype: torch.float32).reshape(4, 5);
+            var bias = torch.arange(1, 21, dtype: torch.float32).reshape(4, 5);
+            var eps = 1e-5;
+
+            var output = torch.nn.functional.layer_norm(input_data, normalized_shape, weight, bias, eps);
+
+            Assert.True(torch.allclose(output, precomputedOutput, atol: 1e-4));
+
+            var layerNorm = torch.nn.LayerNorm(normalized_shape, eps, true, true);
+
```
GilesBathgate, line 47:
```suggestion
            layerNorm.weight = new Parameter(weight);
            layerNorm.bias = new Parameter(bias);
```

PR #1164, opened 2023-11-29
Implement LayerNorm.Dispose()

NiklasGustafsson:


NiklasGustafsson:
Some newer versions of .NET don't like it when tensors held by modules aren't disposed.

Issue #1170, opened 2023-12-03
Calling `pad_packed_sequence` with tensors on cuda and `enforce_sorted=false` causes issue


shaltielshmid:
This bug is actually rooted in `LibTorch`, as reported here: https://github.com/pytorch/pytorch/issues/115027

The issue is that when `enforce_sorted=false` the `unsorted_indices` are used to index on the `lengths` tensor, but the indices are stored on the same device as the input tensor, and `lengths` is stored on the CPU. 

Example code to replicated:
```cs
var input = torch.randn(new[] { 5L, 5L }).@float().cuda();
var lengths = torch.full(5, 5);

var packed = torch.nn.utils.rnn.pack_padded_sequence(input, lengths, enforce_sorted: false);
var error = torch.nn.utils.rnn.pad_packed_sequence(packed);
```

PR #1171, opened 2023-12-03
Fix https://github.com/dotnet/TorchSharp/issues/1170

shaltielshmid:
Fix proposal:

As I noted in the `pytorch` issue, the Python wrapper actually reimplements the `pad_packed_sequence` function in Python, with the fix for moving the `unsorted_indices` to the CPU before indexing on `lengths`. You can see that [here](https://github.com/pytorch/pytorch/blob/main/torch/nn/utils/rnn.py#L338).

An option for a fix would be to do the same - in the LibTorchSharp wrapper, copy in the implementation for `pad_packed_sequence` into a new function, and move the `unsorted_indices` to CPU in the correct place.

This would be a temporary fix until the bug is fixed in `LibTorch`.

### DIFF
--- a/test/TorchSharpTest/TestTorchTensorBugs.cs
+++ b/test/TorchSharpTest/TestTorchTensorBugs.cs
@@ -1411,5 +1411,17 @@ namespace TorchSharp
             var ys = model.forward(xs);
 
         }
+
+        [Fact]
+        public void Validate1170()
+        {
+            if (torch.cuda.is_available()) {
+                var input = torch.randn(new[] { 5L, 5L }).@float().cuda();
+                var lengths = torch.full(5, 5);
+
+                var packed = torch.nn.utils.rnn.pack_padded_sequence(input, lengths, enforce_sorted: false);
+                var error = torch.nn.utils.rnn.pad_packed_sequence(packed);
+            }
+        }
     }
 }


NiklasGustafsson:
@shaltielshmid -- could you add a unit test to the 'TestTorchTensorBugs.cs' file that would fail before the fix, pass after?

NiklasGustafsson (changes requested):
Requested change -- see other comment.

shaltielshmid:
Done.

test/TorchSharpTest/TestTorchTensorBugs.cs
```diff
@@ -1411,5 +1411,17 @@ public void Validate1126()
...
+        [Fact]
+        public void Validate1170()
+        {
+            if (torch.cuda.is_available()) {
+                var input = torch.randn(new[] { 5L, 5L }).@float().cuda();
+                var lengths = torch.full(5, 5);
+
+                var packed = torch.nn.utils.rnn.pack_padded_sequence(input, lengths, enforce_sorted: false);
+                var error = torch.nn.utils.rnn.pad_packed_sequence(packed);
+            }
```
NiklasGustafsson, line 1424: No asserts needed?

shaltielshmid: The error is that an exception was thrown.

Issue #1172, opened 2023-12-06
`optim.LoadStateDict` from existing StateDict doesn't clone tensors


shaltielshmid:
When I call `optim.LoadStateDict` from a state dictionary of an exiting optimizer, the tensors are copied by reference, so if the old optimizer gets disposed the tensors are invalid.

Sample code:
```cs
var lin1 = torch.nn.Linear(10, 10);

var optim1 = torch.optim.Adam(lin1.parameters());
var optim2 = torch.optim.Adam(lin1.parameters());
optim2.load_state_dict(optim1.state_dict());
optim1.Dispose();

torch.nn.functional.mse_loss(lin1.call(torch.rand(10)), torch.rand(10)).backward();
optim2.step();
```

Throws:
```cs
System.InvalidOperationException: 'Tensor invalid -- empty handle.'
```

shaltielshmid:
This also causes issues with devices, if you copy a state dict from an optimizer that was on a different device.

For example:
```cs
var lin1 = torch.nn.Linear(10, 10);
var optim1 = torch.optim.Adam(lin1.parameters());
var sd = optim1.state_dict();

lin1.cuda();
var optim2 = torch.optim.Adam(lin1.parameters());
Console.WriteLine((optim2.state_dict().State[0] as Adam.State).exp_avg.device.type); // CUDA
optim2.load_state_dict(sd);
Console.WriteLine((optim2.state_dict().State[0] as Adam.State).exp_avg.device.type); // CPU

```

PR #1173, opened 2023-12-06
Fix optimizer load state dict copy tensor by reference

shaltielshmid:
Fixes: https://github.com/dotnet/TorchSharp/issues/1172

### DIFF
--- a/src/TorchSharp/Optimizers/Adadelta.cs
+++ b/src/TorchSharp/Optimizers/Adadelta.cs
@@ -237,8 +237,8 @@ namespace TorchSharp
                     acc_delta.Dispose();
 
                     step = st_state.step;
-                    square_avg = st_state.square_avg.clone();
-                    acc_delta = st_state.acc_delta.clone();
+                    square_avg = st_state.square_avg.to(_parameter.device, copy: true);
+                    acc_delta = st_state.acc_delta.to(_parameter.device, copy: true);
                 }
 
                 public override bool ApproximatelyEquals(OptimizerState other)
--- a/src/TorchSharp/Optimizers/Adadelta.cs
+++ b/src/TorchSharp/Optimizers/Adadelta.cs
@@ -237,8 +237,8 @@ namespace TorchSharp
                     acc_delta.Dispose();
 
                     step = st_state.step;
-                    square_avg = st_state.square_avg;
-                    acc_delta = st_state.acc_delta;
+                    square_avg = st_state.square_avg.clone();
+                    acc_delta = st_state.acc_delta.clone();
                 }
 
                 public override bool ApproximatelyEquals(OptimizerState other)
--- a/src/TorchSharp/Optimizers/SGD.cs
+++ b/src/TorchSharp/Optimizers/SGD.cs
@@ -253,7 +253,7 @@ namespace TorchSharp
                 {
                     var st_state = source as State;
                     momentum_buffer?.Dispose();
-                    momentum_buffer = st_state.momentum_buffer.clone();
+                    momentum_buffer = st_state.momentum_buffer.to(_parameter.device, copy: true);
                 }
 
                 /// <summary>
--- a/src/TorchSharp/Optimizers/SGD.cs
+++ b/src/TorchSharp/Optimizers/SGD.cs
@@ -233,6 +233,7 @@ namespace TorchSharp
                 public override void LoadStateDict(BinaryReader reader)
                 {
                     LoadConditionalStateTensor(reader, ref momentum_buffer);
+                    momentum_buffer = momentum_buffer.to(_parameter.device); 
                 }
 
                 /// <summary>
--- a/src/TorchSharp/Optimizers/SGD.cs
+++ b/src/TorchSharp/Optimizers/SGD.cs
@@ -251,10 +251,8 @@ namespace TorchSharp
                 public override void LoadStateDict(OptimizerState source)
                 {
                     var st_state = source as State;
-                    if (momentum_buffer is not null) {
-                        momentum_buffer.Dispose();
-                    }
-                    momentum_buffer = st_state.momentum_buffer;
+                    momentum_buffer?.Dispose();
+                    momentum_buffer = st_state.momentum_buffer.clone();
                 }
 
                 /// <summary>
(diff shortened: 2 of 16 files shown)


shaltielshmid:
Also fixes https://github.com/dotnet/TorchSharp/issues/1176 with the conditional state being copied to the right device.

src/TorchSharp/Optimizers/SGD.cs
```diff
@@ -233,6 +233,7 @@ public override void to(Device device)
                 public override void LoadStateDict(BinaryReader reader)
                 {
                     LoadConditionalStateTensor(reader, ref momentum_buffer);
+                    momentum_buffer = momentum_buffer.to(_parameter.device);
```
NiklasGustafsson, line 236: No need to dispose the old momentum buffer if it's actually moved?

src/TorchSharp/Optimizers/SGD.cs
```diff
@@ -251,10 +252,8 @@ public override void SaveStateDict(BinaryWriter writer)
                 public override void LoadStateDict(OptimizerState source)
                 {
                     var st_state = source as State;
-                    if (momentum_buffer is not null) {
-                        momentum_buffer.Dispose();
-                    }
-                    momentum_buffer = st_state.momentum_buffer;
+                    momentum_buffer?.Dispose();
+                    momentum_buffer = st_state.momentum_buffer.to(_parameter.device, copy: true);
```
NiklasGustafsson, line 256: What happens to the old state.momentum_buffer? Does it leak?

shaltielshmid: Are you referring to the momentum_buffer located inside the `st_state` variable?
If so, I don't think that's an issue. The state dictionary contains references to tensors inside other existing optimizers. It's not the responsibility of the loader to dispose the tensors it received, since those tensors might still be in use.

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -777,13 +777,15 @@ public Tensor to(ScalarType type, torch.Device device, bool copy = false)
...
             /// </summary>
             /// <param name="device">A string denoting the target device.</param>
-            public Tensor to(string device) => to(new torch.Device(device));
+            /// <param name="copy">When copy is set, a new Tensor is created even when the Tensor already matches the desired conversion.</param>
+            public Tensor to(string device, bool copy = false) => to(new torch.Device(device), copy);
 
             /// <summary>
             /// Moves the tensor data.
             /// </summary>
             /// <param name="device">The target device</param>
```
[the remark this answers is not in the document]
shaltielshmid: Good point, adding it in

src/TorchSharp/Optimizers/SGD.cs
shaltielshmid: Good catch. Will fix.

shaltielshmid: Question: Do you want me to write something like:
```cs
if (momentum_buffer.device_type != _parameter.device_type || momentum_buffer.device_index != _parameter.device_index) {
    using var copy = result;
    result = copy.to(_parameter.device);
}
```

shaltielshmid: (That's not handled in the regular `Optimizer.to(Device device)` function. See [here](https://github.com/dotnet/TorchSharp/blob/main/src/TorchSharp/Optimizers/SGD.cs#L226))

NiklasGustafsson: I think it's sufficient to check the `Handle` before and after the call to `to'. I think.

NiklasGustafsson: It's handled in the Module variants of `to()`. Follow that pattern.

NiklasGustafsson: Maybe it's worth writing an internal utility function that has the 'move or not move' predicate, so it can be reused.

shaltielshmid:
@NiklasGustafsson What do you think of my solution?
I added a `disposeAfter` flag to the `to(..)` function which dispose the input tensor after moving to the new tensor. 
Turns out we didn't need to check if the tensor actually moved, because we get a new handle regardless. 

I wrote a unit test attempting to move and cast tensors to the same and different types/devices, and made sure the behavior was as expected. 

What do you think?

If this solution is good, I'll wait for this to be merged and then use it in the other PR.

NiklasGustafsson:
@shaltielshmid -- here's a tip: Once you have created a PR, try to consolidate a number of commits before pushing again. Every push starts a new build on the CI/CD pipeline, which wastes resources.

shaltielshmid:
Of course, I apologize. 
I generally try to make a few commits, but only push when I complete changes.

src/TorchSharp/Optimizers/Adadelta.cs
```diff
@@ -237,8 +237,8 @@ public override void LoadStateDict(OptimizerState source)
                     acc_delta.Dispose();
 
                     step = st_state.step;
-                    square_avg = st_state.square_avg;
-                    acc_delta = st_state.acc_delta;
+                    square_avg = st_state.square_avg.to(_parameter.device, copy: true);
```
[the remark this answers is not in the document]
shaltielshmid: If the State Dictionary is actually from another existing tensor, then the tensors are copied by reference and we don't want two different optimizers sharing a state tensor.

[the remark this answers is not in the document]
NiklasGustafsson: In that case, shouldn't the 'copy' argument be passed down from where you know whether it's necessary or not?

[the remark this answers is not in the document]
shaltielshmid: I can add it in, if you think it's the right way to it.

The way I see it, the StateDictionary is an object which is created only by calling `Optimizer.state_dict()`, and the source Optimizer is the one which manages the tensors that are in the dictionary. If they are to be disposed, then the Optimizer would handle them. And therefore, the new Optimizer should receive a fresh copy of the tensors to manage itself.

NiklasGustafsson:
You have run the unit tests on CUDA, right?

NiklasGustafsson:
Merged

Issue #1174, opened 2023-12-06
`Tensor.Load` when tensor is on CUDA fails


shaltielshmid:
When trying to load the contents of a Tensor from a stream, if the tensor isn't located on the CPU then the loading will fail. The reason seems to be located [here](https://github.com/dotnet/TorchSharp/blob/main/src/TorchSharp/Tensor/TensorExtensionMethods.cs#L466), since the `.to()` method returns a new tensor when the tensor is copied to a new device. 

```cs
var tensor1 = torch.rand(10, device: torch.CUDA);

// Save to memory stream
using var ms = new MemoryStream();
tensor1.Save(new BinaryWriter(ms, Encoding.UTF8, true));

ms.Position = 0;
tensor1.Load(new BinaryReader(ms));
```

PR #1175, opened 2023-12-06
Make sure to handle tensor copy when moving to cpu before loading

shaltielshmid:
Fixes https://github.com/dotnet/TorchSharp/issues/1174

### DIFF
--- a/src/TorchSharp/Tensor/TensorExtensionMethods.cs
+++ b/src/TorchSharp/Tensor/TensorExtensionMethods.cs
@@ -463,9 +463,14 @@ namespace TorchSharp
 
             if (!skip) {
                 var device = tensor.device;
-                if (device.type != DeviceType.CPU) tensor.to(CPU);
-                tensor.bytes = bytes;
-                tensor.to(device);
+                if (device.type != DeviceType.CPU) {
+                    using var copy = tensor.to(CPU);
+                    copy.bytes = bytes;
+                    tensor.set_(copy.to(device));
+                }
+                else {
+                    tensor.bytes = bytes;
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/TorchSharp/Tensor/TensorExtensionMethods.cs
```diff
@@ -463,9 +463,14 @@ public static void Load(this Tensor tensor, System.IO.BinaryReader reader, bool
 
             if (!skip) {
                 var device = tensor.device;
-                if (device.type != DeviceType.CPU) tensor.to(CPU);
-                tensor.bytes = bytes;
-                tensor.to(device);
+                if (device.type != DeviceType.CPU) {
+                    using var copy = tensor.to(CPU);
+                    copy.bytes = bytes;
+                    tensor.set_(copy.to(device));
```
NiklasGustafsson, line 469: What happens to the argument to `set_()`? Does it get disposed somewhere?

src/TorchSharp/Tensor/TensorExtensionMethods.cs
```diff
@@ -532,9 +537,9 @@ public static void Load(ref Tensor? tensor, System.IO.BinaryReader reader, bool
 
             if (!skip) {
                 var device = tensor.device;
-                if (device.type != DeviceType.CPU) tensor.to(CPU);
+                if (device.type != DeviceType.CPU) tensor = tensor.to(CPU); 
                 tensor.bytes = bytes;
-                tensor.to(device);
+                tensor = tensor.to(device);
```
NiklasGustafsson, line 542: Dispose the old tensor?

shaltielshmid: Done.

Issue #1148, opened 2023-11-19
Module.to should not be differentiable


zgxnet:
Current Module.to implementation keeps the moved tensors in the same graph.

```C#
                        case Parameter param: {
                                var t = param.to(dtype, device); //in the same graph
                                t.retain_grad();
                                ...
                            }

                        case Tensor tensor when (device.type != tensor.device_type || device.index != tensor.device_index): {
                                var t = tensor.to(dtype, device); //in the same graph
                                ...
                            }
```
While in the pytorch implementation, the moved tensors are detached.
```python
        for key, param in self._parameters.items():
            if param is None:
                continue
            # Tensors stored in modules are graph leaves, and we don't want to
            # track autograd history of `param_applied`, so we have to use
            # `with torch.no_grad():`
            with torch.no_grad():  #detached from the original tensor
                param_applied = fn(param)
                ...
            if param.grad is not None:
                with torch.no_grad(): #detached from the original tensor
                    grad_applied = fn(param.grad)
                ...
```
As the module tensors should be leaf tensors, it makes no sense to keep the original tensors. If a CUDA module is not detached from the original CPU tensors, there are unnecessary gradient traffics from the GPU to the CPU in very backpropagation.

NiklasGustafsson:
@shaltielshmid -- since you have been spending time in 'to' lately... Can you comment, please?

NiklasGustafsson:
> I'm leaning towards option 2 since that is a more direct way of solving the issue, what do you think?

Whichever uses less memory. We don't want to hit the upper limit of GPU memory with temporary storage use.

Issue #1179, opened 2023-12-07
Module.to doesn't transfer parameters in `ParameterList` or `ParameterDict`


shaltielshmid:
When moving a `ParameterList` or `ParameterDict` Module, the registered values get moved but the field doesn't get the updated values. 

For example, in the case below, the output would be Float32. If you would inspect the _internal_params list you would see that the parameter there is actually Float16. 

This also means that the garbage collector will never dispose of the old parameter since they are both being referenced. 

Example code:

```cs
private class TestModule1 : Module {
    public TestModule1() : base("TestModule1") {
        list.append(Parameter(torch.rand(10)));
    }
    public ParameterList list = new ParameterList();
}
private static void Test() {
    var m = new TestModule1(torch.randn(2, 2), true);
    m.to(ScalarType.Float16);
    
    Console.WriteLine(m.list[0].dtype);
}
```

PR #1181, opened 2023-12-09
Fixed `Module.to()` bugs with `ParameterList` and `ParameterDict`, and with autograd tracking movements between CPU & GPU

shaltielshmid:
Fixes https://github.com/dotnet/TorchSharp/issues/1148 and https://github.com/dotnet/TorchSharp/issues/1179.

I went on a whim and attempted to make a style change on the _toEpilog method. If you see a fatal error or prefer to leave it as it used to be - no problem, I'll revert that change.

Main changes:
1] Merged the three _toEpilog methods into one method.
2] Instead of re-iterating the fields every time and reassigning the parameters, I build a dictionary of the fields and as we go through the registered parameters we check if they have a field, and if so - assign it. Saves us from duplicating the code which handles the moving of the parameters.
3] When moving Parameters, we do it in a "torch.no_grad()" scope to avoid having autograd track the movement and so that the resulting tensor will be a leaf.
4] After moving the parameters and buffers, the old ones are disposed. 
5] Overrode the `_to()` method in the `ParameterList` and `ParameterDict` classes. I didn't do it in the `ModuleDict` and `ModuleList` methods since modules themselves don't get a new reference, only their parameters and buffers, and `_toEpilog()` iterates through every registered module.
6] Added `Tensor` extension method `toWillCopy()` which given a set of arguments to `to()` will return whether the tensor will be copied.

Question about the memo properties `_deviceType` and `_deviceIndex`. I see the value in having them, but that can cause an issue if someone calls `.to()` on a sub module, and then tries to call it on the main module() to make sure all the submodules() are aligned, it wont work. Same thing if any of the parameters are moved separately, then calling `Module.to()` wont behave as people expect.

### DIFF
--- a/src/TorchSharp/NN/Module.cs
+++ b/src/TorchSharp/NN/Module.cs
@@ -229,19 +229,12 @@ namespace TorchSharp
                         else sm._to(device, dtype.Value);
                     }
 
-                    // Define our validation function once
-                    bool willMove(Tensor tensor)
-                    {
-                        return (dtype is not null && dtype.Value != tensor.dtype) ||
-                            (device is not null && (device.type != tensor.device_type || device.index != tensor.device_index));
-                    }
-
                     var fieldsByComponentName = GetType().GetFields(BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Instance)
                                                             .ToDictionary(field => field.ComponentName());
 
                     foreach (var (name, param) in named_parameters(false).ToList()) {
-                        if (!willMove(param)) continue;
-                        
+                        if (!param.toWillCopy(dtype ?? param.dtype, device ?? param.device)) continue;
+
                         // Store the requires_grad flag ahead, since we dispose the parameter after moving
                         bool requiresGrad = param.requires_grad;
                         Parameter p;
@@ -258,8 +251,8 @@ namespace TorchSharp
                     }
 
                     foreach (var (name, buffer) in named_buffers(false).ToList()) {
-                        if (!willMove(buffer)) continue;
-                        
+                        if (!buffer.toWillCopy(dtype ?? buffer.dtype, device ?? buffer.device)) continue;
+
                         // Buffers don't get grads so we don't need to detach them afterwards
                         var t = buffer.to(dtype ?? buffer.dtype, device ?? buffer.device, disposeAfter: true);
                         ConditionallyRegisterBuffer(name, t);
--- a/src/TorchSharp/NN/Module.cs
+++ b/src/TorchSharp/NN/Module.cs
@@ -229,20 +229,11 @@ namespace TorchSharp
                         else sm._to(device, dtype.Value);
                     }
 
-                    // Define our validation function
+                    // Define our validation function once
                     bool willMove(Tensor tensor)
                     {
                         return (dtype is not null && dtype.Value != tensor.dtype) ||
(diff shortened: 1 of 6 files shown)


src/TorchSharp/NN/Module.cs
```diff
@@ -325,55 +208,62 @@ protected internal virtual Module _to(ScalarType dtype)
...
-                        case Parameter param: {
-                                var t = param.to(dtype);
-                                t.retain_grad();
-                                var p = new Parameter(t, param.requires_grad);
-                                field.SetValue(this, p);
-                                ConditionallyRegisterParameter(fieldName, p);
-                                alreadyHandled.Add(p.handle);
-                                break;
-                            }
+                    var fieldsByComponentName = GetType().GetFields(BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Instance)
```
NiklasGustafsson, line 232: Does this have any performance implications?

src/TorchSharp/NN/Module.cs
```diff
@@ -325,55 +208,62 @@ protected internal virtual Module _to(ScalarType dtype)
...
-                        ConditionallyRegisterParameter(name, t);
+                        // If this parameter is a field, set it
+                        if (fieldsByComponentName.TryGetValue(name, out var field))
+                            field.SetValue(this, p);
                     }
 
                     foreach (var (name, buffer) in named_buffers(false).ToList()) {
-                        if (alreadyHandled.Contains(buffer.handle)) continue;
-                        var t = buffer.to(dtype);
+                        if (!buffer.toWillCopy(dtype ?? buffer.dtype, device ?? buffer.device)) continue;
```
NiklasGustafsson, line 254: Are old buffers/parameters disposed anywhere?

src/TorchSharp/NN/Module.cs
shaltielshmid: Under the assumption that the costly operation is the reflection, then there shouldn't be. 
We're creating a hash collection of the same size (alreadyHandled vs the Dictionary), so I guess the cost would be a the lookups in the dictionary, which should be very minimal.

src/TorchSharp/NN/Module.cs
shaltielshmid: Yup. In the ".to()" call I use the disposeAfter parameter.

src/TorchSharp/NN/Module.cs
NiklasGustafsson: The brings up a functional concern -- the alreadyHandled set is there to make sure that we don't accidentally deal with the same tensor twice. Is that no longer a possibility?

shaltielshmid: I don't think so. 
The reason it was needed before was because the function was iterating through two different lists of parameters. One using reflection (the parameters registered through the `RegisterComponents()` function), and the other is using the internal list of registered parameters. 
In the proposed code we only go through the list of registered parameters, so there isn't a concern of dealing with the same tensor twice.
Unless there is a case where someone can register the same tensor twice?

shaltielshmid: If someone registers the same parameter under two different names, then we have an issue - the first encounter will dispose the parameter, and then the second time it will have a null tensor error. 
Should this be a use case we should handle?

NiklasGustafsson (approved):

Issue #1180, opened 2023-12-07
`Module.to(ScalarType)` has restrictions in `PyTorch` which aren't restricted in `TorchSharp`


shaltielshmid:
When calling `Module.to(ScalarType)` on a module, PyTorch has a restriction that the target dtype must be a floating point or a complex number. See [here](https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/module.py#L1136). 

Also, when it gets a floating point type, PyTorch only moves parameters that are already a floating_point type to the new type. See for example [here](https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/module.py#L1136). 

This is relatively trivial to add, would TorchSharp want to add the same restrictions?

NiklasGustafsson:
Under the assumption that modules always participate in backprop, this constraint makes sense.

PR #1182, opened 2023-12-11
Validate the target data type when converting a Module using `to()`

NiklasGustafsson:
See #1180


shaltielshmid:
In PyTorch there is an additional restriction that when it does accept a floating point dtype, it only applies the move to parameters and buffers that are already floating points, see [here](https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/module.py#L1150).

NiklasGustafsson:
> In PyTorch there is an additional restriction that when it does accept a floating point dtype, it only applies the move to parameters and buffers that are already floating points, see [here](https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/module.py#L1150).

Thanks for pointing it out. It's not exactly what the expression says, but it was something I missed.

Issue #1185, opened 2023-12-14
Incomplete transfer of module to device (only with 0.101.3)


dje-dev:
Upon upgrading to 0.101.3 a previously working network now throws exception upon calling forward. The module is constructed on CPU with multiple layers in constructor and as normal, RegisterComponents is called. Module is subsequently moved to cuda. Most layers successfully are transferred, but at least one (LayerNorm) remains on cpu. This causes exception below.

Downgrading back to 0.101.2 immediately fixed the problem.
  
Unhandled exception. System.Runtime.InteropServices.ExternalException (0x80004005): Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking argument for argument weight in method wrapper_CUDA__native_layer_norm)

NiklasGustafsson:
@shaltielshmid -- I think this one will require your attention.

shaltielshmid:
Can you please send sample code to replicate this behavior?

shaltielshmid:
Okay, after some digging, I found the root of the problem to be from here: https://github.com/dotnet/TorchSharp/pull/1152

In TorchSharp 0.101.3 the LayerNorm class was changed to act like torch.nn.functional.
In the constructor, the LayerNorm module calls the base constructor with a "IntPtr.Zero" handle. 
When `module.to()` gets called, then it goes through all the registered modules and calls `.to()` on them as well. Because LayerNorm's handle is zero, so `LayerNorm.handle.IsInvalid` always returns true and therefore it is not registered as a module. 

You can see the problematic line [here](https://github.com/dotnet/TorchSharp/blob/main/src/TorchSharp/NN/Normalization/LayerNorm.cs#L23).

@NiklasGustafsson Is there any reason to register the LayerNorm module with an invalid handle? Ideally we should just pass a string with the name of the module (either default it to "LayerNorm" or require it in the constructor).

GilesBathgate:
@shaltielshmid if `base(IntPtr.Zero, IntPtr.Zero)` is problematic then it can easily be replaced with a named module base constructor. I chose to use IntPtr.Zero as this seems to be the way its done elsewhere, when there is no module handle required for calls to native functions. How does module.to work with user defined modules?

GilesBathgate:
@NiklasGustafsson Perhaps I am meant to call `base(IntPtr.Zero, IntPtr.Zero ownsHandle: false)` ?

shaltielshmid:
[...]

PR #1188, opened 2023-12-14
Address #1185

NiklasGustafsson:
@shaltielshmid, @GilesBathgate -- please review.

### DIFF
--- a/src/TorchSharp/NN/Module.cs
+++ b/src/TorchSharp/NN/Module.cs
@@ -251,7 +251,9 @@ namespace TorchSharp
                         // In addition, we need the new tensor to be a leaf to accumulate gradients, so if we didn't
                         // disable grad we would need to call .detach() on the moved tensor.
                         using (var d = torch.no_grad())
-                            p = new Parameter(param.to(paramType, device ?? param.device, disposeAfter: true), requiresGrad);
+                            p = new Parameter(
+                                param.to(paramType, device ?? param.device, disposeAfter: true).DetachFromDisposeScope(), requiresGrad)
+                                .DetachFromDisposeScope() as Parameter;
                         ConditionallyRegisterParameter(name, p);
 
                         // If this parameter is a field, set it
@@ -266,7 +268,7 @@ namespace TorchSharp
                             dtype != null && (buffer.dtype.IsFloatingPoint() || buffer.dtype.IsComplex()) ? dtype.Value : buffer.dtype;
 
                         // Buffers don't get grads so we don't need to detach them afterwards
-                        var t = buffer.to(bufferType, device ?? buffer.device, disposeAfter: true);
+                        var t = buffer.to(bufferType, device ?? buffer.device, disposeAfter: true).DetachFromDisposeScope();
                         ConditionallyRegisterBuffer(name, t);
 
                         if (fieldsByComponentName.TryGetValue(name, out var field))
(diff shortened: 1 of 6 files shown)


test/TorchSharpTest/NN.cs
```diff
@@ -4714,6 +4714,12 @@ public void TestLayerNorm()
                     Assert.Equal(ones.shape, pooled.shape);
                 }
             }
+            if (torch.cuda.is_available()) {
+                using var pool = LayerNorm(new long[] { 12, 28, 28 }, device: torch.CPU);
+
+                using var pool1 = pool.to(torch.CUDA);
+                Assert.Equal(DeviceType.CUDA, pool1.weight.device_type);
+            }
```
shaltielshmid, line 4722: I don't think this unit test covers the bug reported. 
The issue was the LayerNorm not being recognized as a submodule, so perhaps placing it inside a sequential?

src/TorchSharp/NN/Module.cs
```diff
@@ -251,7 +251,9 @@ private void _toEpilog(ScalarType? dtype, Device device)
                         // In addition, we need the new tensor to be a leaf to accumulate gradients, so if we didn't
                         // disable grad we would need to call .detach() on the moved tensor.
                         using (var d = torch.no_grad())
-                            p = new Parameter(param.to(paramType, device ?? param.device, disposeAfter: true), requiresGrad);
+                            p = new Parameter(
+                                param.to(paramType, device ?? param.device, disposeAfter: true).DetachFromDisposeScope(), requiresGrad)
+                                .DetachFromDisposeScope() as Parameter;
```
shaltielshmid, line 256: I may be wrong about this - If I understood the Parameter class correctly, the handle is moved from the Tensor to the Parameter on creation (and the Tensor's handle is set to 0, so we don't care if it's disposed), and therefore we don't need to detach the internal tensor, just the final parameter.

shaltielshmid:
Also - that extra test you added for the bug, were there any changes done to fix it?

test/TorchSharpTest/NN.cs
NiklasGustafsson: But that wasn't the actual issue. 

The issue was that a) it used IntPtr.Zero in the base constructor call, and then (once that was addressed) b) that the 'weight' and 'bias' properties weren't set, so the parameters fetched by the properties were the old ones. The ComponentName attribute fixed that.

NiklasGustafsson:
> Also - that extra test you added for the bug, were there any changes done to fix it?

No, but I liked testing that indexing with an empty tensor filter works. :-)

src/TorchSharp/NN/Module.cs
NiklasGustafsson: Yeah... better safe than sorry.

shaltielshmid: Fair enough :)

test/TorchSharpTest/NN.cs
shaltielshmid: Got it. So maybe just add a check to clarify that it isn't invalid?

```cs
Assert.False(pool.IsInvalid)
```

NiklasGustafsson: `Module.handle` is not public, so no can do. However, any call to `to()` will blow up if it's invalid, so the fact that the call does not should prove the same point.

shaltielshmid: Good point, I missed that part

shaltielshmid (approved):

Issue #1191, opened 2023-12-19
Having trouble moving a module from one GPU to another.


Biotot:
### Discussed in https://github.com/dotnet/TorchSharp/discussions/1190

<div type='discussions-op-text'>

<sup>Originally posted by **Biotot** December 19, 2023</sup>
I've been banging my head against this for a couple days and I'm still coming up empty. 

I have multiple modules and multiple GPUs, however this sequence continues to fail. I've narrowed it down to being a problem with the model. I can load it fresh from a file each loop and the error no longer exists.

(Pseudocode)
ModuleA.to(cuda:0)
TrainLoop
ModuleA.to(cpu)

ModuleA.to(cuda:1)
TrainLoop
Exception: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cuda:1

The error is consistently in the loss output.backward() call if that helps. 

This error doesn't happen if I load the module from a file each loop. Input data is not the issue, the model isn't correctly switching devices. I've tried many different combinations of code and have tried directly moving from cuda:0 to cuda:1 without luck. 
I'm not sure what is going wrong, I've been porting over my code from pytorch and I've been trying to get over this hurdle. Any help would be appreciated. </div>

Running on TorchSharp-cuda-windows 0.101.4

NiklasGustafsson:
Is there a more detailed traceback. 

Given that it's in `backward()` that it happens, maybe the gradients aren't being moved? What if you first zero the gradients using `zero_grad()`, and then move the model?

Biotot:
Sorry for the massive block. 
```

System.Runtime.InteropServices.ExternalException
  HResult=0x80004005
  Message=Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cuda:1!
Exception raised from compute_types at C:\actions-runner\_work\pytorch\pytorch\builder\windows\pytorch\aten\src\ATen\TensorIterator.cpp:484 (most recent call first):
00007FFB67B6AF4200007FFB67B6AEE0 c10.dll!c10::Error::Error [<unknown file> @ <unknown line number>]
00007FFB67B6AACA00007FFB67B6AA70 c10.dll!c10::detail::torchCheckFail [<unknown file> @ <unknown line number>]
00007FFB4C48704A00007FFB4C486740 torch_cpu.DLL!at::TensorIteratorBase::compute_types [<unknown file> @ <unknown line number>]
[...]

PR #1193, opened 2023-12-20
Fixed gradients moving and zeroing

shaltielshmid:
Fixing https://github.com/dotnet/TorchSharp/issues/1191

Modified to behave like PyTorch / LibTorch, see here: https://github.com/pytorch/pytorch/blob/main/torch/csrc/api/src/nn/module.cpp#L253 and https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/module.py#L836

Important to note: This introduces a breaking change, as calling zero_grad() by default now converts gradients to null. 
This is probably a good breaking change, because up until now it wasn't consistent and modules that were managed by LibTorch (like Linear) had their gradients set to null whereas the modules managed by TorchSharp had their gradients set to zero.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -4,6 +4,13 @@ Releases, starting with 9/2/2021, are listed with the most recent release at the
 
 ## NuGet Version 0.101.4
 
+__Bug Fixes__:
+
+#1191 : Having trouble moving a module from one GPU to another with gradients.<br/>
+
+
+## NuGet Version 0.101.4
+
 A fast-follow release addressing a regression in v0.101.3
 
 __Bug Fixes__:
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -2,7 +2,7 @@
 
 Releases, starting with 9/2/2021, are listed with the most recent release at the top.
 
-## NuGet Version 0.101.4
+## NuGet Version 0.101.5
 
 __Bug Fixes__:
 
--- a/src/TorchSharp/NN/Module.cs
+++ b/src/TorchSharp/NN/Module.cs
@@ -364,7 +364,7 @@ namespace TorchSharp
                         if (grad is not null) {
                             if (set_to_none) {
                                 p.set_grad(null);
-                                grad.Dispose();
+                                grad.DetachFromDisposeScope().Dispose();
                             } else {
                                 grad.zero_();
                             }
--- a/src/TorchSharp/NN/Module.cs
+++ b/src/TorchSharp/NN/Module.cs
@@ -239,10 +239,10 @@ namespace TorchSharp
                                                             .ToDictionary(field => field.ComponentName());
 
                     foreach (var (name, param) in named_parameters(false).ToList()) {
-                        if (!param.toWillCopy(dtype ?? param.dtype, device ?? param.device)) continue;
+                        if (!param.toWillCopy(dtype ?? param.dtype, device ?? param.device) &&
+                            (param.grad() is null || !param.grad().toWillCopy(dtype ?? param.dtype, device ?? param.device)))
+                            continue;
 
-                        // Store the requires_grad flag ahead, since we dispose the parameter after moving
-                        bool requiresGrad = param.requires_grad;
                         Parameter p;
                         ScalarType paramType =
                             dtype != null && (param.dtype.IsFloatingPoint() || param.dtype.IsComplex()) ? dtype.Value : param.dtype;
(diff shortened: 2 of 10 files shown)


RELEASENOTES.md
```diff
@@ -2,6 +2,13 @@
 
 Releases, starting with 9/2/2021, are listed with the most recent release at the top.
 
+## NuGet Version 0.101.4
```
NiklasGustafsson, line 5: Should be 0.101.5

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -1246,6 +1246,17 @@ public Tensor pin_memory()
                 return new Tensor(res);
             }
 
+            /// <summary>
+            /// This function will set the `tensor.grad()` attribute to a custom tensor. Make sure to detach the tensor from
```
NiklasGustafsson, line 1250: Should the detach be taken care of here, then, perhaps? Of should it be transferred to the dispose scope of the tensor the grad belongs to?

NiklasGustafsson (changes requested):

RELEASENOTES.md
shaltielshmid: Will fix.

src/TorchSharp/Tensor/Tensor.cs
shaltielshmid: Sure, will transfer to the dispose scope of the tensor.

shaltielshmid: I don't see an easy way to do that. How would I do that?
For now, just applying the detach.

src/TorchSharp/NN/Module.cs
```diff
@@ -341,15 +354,20 @@ public virtual void eval()
...
+                    THSNN_Module_zero_grad(handle, set_to_none);
                     CheckForErrors();
 
                     foreach (var (_, p) in named_parameters()) {
                         var grad = p.grad();
                         if (grad is not null) {
-                            grad.zero_();
+                            if (set_to_none) {
+                                p.set_grad(null);
+                                grad.Dispose();
```
NiklasGustafsson, line 367: Is it worth first detaching from its dispose scope? I think it's a no-op, but it feels cleaner.

shaltielshmid: Meaning, call `grad.DetachFromDisposeScope().Dispose()`?

shaltielshmid:
Before you merge, I'm still perplexed about something in his code. I'll update you when I figure it out.

src/TorchSharp/NN/Module.cs
NiklasGustafsson: Yes

src/TorchSharp/Tensor/Tensor.cs
NiklasGustafsson: How about adding a method to DisposeScope for a direct transfer rather than transfer-to-parent?

NiklasGustafsson:
There's still time to make a release tomorrow. After that, I'll be out for the rest of the year.

shaltielshmid:
Okay, so this is the core reason of this issue, is that the underlying `Module.to()` in LibTorch moves the data of each tensor, but keeps the gradient as is.

NiklasGustafsson:
> Okay, so this is the core reason of this issue, is that the underlying `Module.to()` in LibTorch moves the data of each tensor, but keeps the gradient as is.

And that should only be a problem if there is ever a scenario when you would want to move a module without clearing the gradients, such as in the middle of an epoch. Why would you ever do that?

src/TorchSharp/Tensor/Tensor.cs
shaltielshmid: Sure, but perhaps we should keep this in a different release?
I'd rather not mess with the dispose scopes right before you're off.

src/TorchSharp/NN/Module.cs
shaltielshmid: Done

shaltielshmid:
I agree that it's a weird scenario, but the way LibTorch handles it seems wrong. 
When you move a tensor separately, it doesn't copy the gradients. But when you move tensors as part of a module, it does copy the gradients.

NiklasGustafsson:
> I agree that it's a weird scenario, but the way LibTorch handles it seems wrong. When you move a tensor separately, it doesn't copy the gradients. But when you move tensors as part of a module, it does copy the gradients.

So is your change consistent with LibTorch, or consistent between Tensor and Module?

shaltielshmid:
Neither. Consistent with PyTorch. 
Tensor.to will create a new tensor without the gradient. 
Module.to will move the tensors and the gradients with them. 

In LibTorch module.to leaves the gradients on the old device.

NiklasGustafsson:
> Neither. Consistent with PyTorch. Tensor.to will create a new tensor without the gradient. (Let me confirm that this is the behavior of PyTorch). Module.to will move the tensors and the gradients with them.
> 
> In LibTorch module.to leaves the gradients on the old device.

Okay, that sounds best.

src/TorchSharp/Tensor/Tensor.cs
shaltielshmid: If you think it's a good idea, I put it in code. It's currently in a PR on my fork into this branch. https://github.com/shaltielshmid/TorchSharp/pull/2

If this draft looks good to you, I'll merge it in and we can put it into this version.

PR #1195, opened 2023-12-23
Implementation of jacobian method, based on PyTorch implementation

### LINKED ISSUE #1149: Compute jacobian
I have a Module that computes a Tensor vector from a Tensor vector. I wish to compute the gradient of each output element with respect to each input element. In the Pytorch documentation, they show this can be done through the use of `torch.autograd.functional.jacobian` (see [torch.autograd.functional.jacobian](https://pytorch.org/docs/stable/generated/torch.autograd.functional.jacobian.html#torch.autograd.functional.jacobian)). I'm not sure the same can be done using TorchSharp.

I tried looping over the elements of the output tensor and doing a `backward` call:
```
for (int i = 0; i < output.Length; i++)
    output[i].backward(create_graph: false, retain_graph: true);
```
But even with `retain_graph=true` that results in the error:
```
System.Runtime.InteropServices.ExternalException: 'Trying to backward through the graph a second time (or directly access saved tensors after they have already been freed). Saved intermediate values of the graph are freed when you call .backward() or autograd.grad(). Specify retain_graph=True if you need to backward through the graph a second time or if you need to access saved tensors after calling backward.
```

Is there currently already a way to compute the jacobian of a function/module in TorchSharp? If not, can you suggest how I could get started on adding support for the above Pytorch functionality?

roelofsmn:
Initial implementation of the jacobian method, based on https://pytorch.org/docs/stable/_modules/torch/autograd/functional.html#jacobian .
Resolves #1149.

Omits several options from the PyTorch implementation, such as `create_graph`, `strict`, `vectorize` and `strategy`.
It behaves as if `create_graph=False`, `strict=False`, `vectorize=False` and `strategy="reverse-mode"`.

I added an example of usage in the documentation of the method itself.

@NiklasGustafsson

### DIFF
--- a/src/TorchSharp/Functional.cs
+++ b/src/TorchSharp/Functional.cs
@@ -106,6 +106,40 @@ namespace TorchSharp
                     }
                 }
 
+                /// <summary>
+                /// Computes the jacobian of a given function with one input and one output.
+                /// </summary>
+                /// <param name="function">The function to compute the jacobian for. It has a single input and single output.</param>
+                /// <param name="inputs">The value for the input at which to compute the jacobian.</param>
+                /// <returns>A single tensor containing the jacobian.</returns>
+                public static Tensor jacobian(Func<Tensor, Tensor> function, Tensor inputs)
+                {
+                    var wrapper = new Func<Tensor[], Tensor[]>((Tensor[] x) => {
+                        return new Tensor[] { function(x.Single()) };
+                    });
+                    return jacobian(wrapper, new Tensor[] { inputs }).Single();
+                }
+
+                /// <summary>
+                /// Computes the jacobian of a given function with one input and multiple outputs.
+                /// </summary>
+                /// <param name="function">The function to compute the jacobian for. It has a single input and multiple outputs.</param>
+                /// <param name="inputs">The value for the input at which to compute the jacobian.</param>
+                /// <returns>A list of tensors containing the jacobian for each output.</returns>
+                public static IEnumerable<Tensor> jacobian(Func<Tensor, Tensor[]> function, Tensor inputs)
+                {
+                    var wrapper = new Func<Tensor[], Tensor[]>((Tensor[] x) => {
+                        return function(x.Single());
+                    });
+                    return jacobian(wrapper, new Tensor[] { inputs });
+                }
+
+                /// <summary>
+                /// Computes the jacobian of a given function with multiple inputs and one output.
+                /// </summary>
+                /// <param name="function">The function to compute the jacobian for. It has multiple inputs and single output.</param>
+                /// <param name="inputs">The values for the inputs at which to compute the jacobian.</param>
+                /// <returns>A list of tensors containing the jacobian for the output with respect to each input.</returns>
                 public static IEnumerable<Tensor> jacobian(Func<Tensor[], Tensor> function, params Tensor[] inputs)
                 {
                     var wrapper = new Func<Tensor[], Tensor[]>((Tensor[] x) => {
--- a/src/TorchSharp/Functional.cs
+++ b/src/TorchSharp/Functional.cs
@@ -1,0 +1,114 @@
+using System;
+using System.Linq;
+using System.Collections.Generic;
+using static TorchSharp.torch.autograd;
+
+namespace TorchSharp
+{
+    public static partial class torch
+    {
+        public static partial class autograd
+        {
+            public static partial class functional
+            {
+                /// <summary>
+                /// Preprocesses the inputs to make sure they require gradient.
+                /// </summary>
+                /// <param name="need_graph">Specifies if we internally want gradients to flow back to the Tensors in the result.</param>
+                /// <param name="inputs">The tensors to preprocess.</param>
+                /// <returns>The same tensors as inputs, but requiring grad.</returns>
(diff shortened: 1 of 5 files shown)


roelofsmn:
@dotnet-policy-service agree

NiklasGustafsson:
@roelofsmn -- thanks for the PR! I've been out on Christmas break, but I'll take a look at this later today.

src/TorchSharp/Functional.cs
```diff
@@ -0,0 +1,114 @@
...
+                /// </summary>
+                /// <param name="function">The function to compute the jacobian for. It may have multiple inputs and multiple outputs.</param>
+                /// <param name="inputs">The values for the inputs at which to compute the jacobian.</param>
+                /// <returns>A list of tensors, one for each output value of each output. For example, a single output with two elements yields two jacobian tensors.</returns>
+                public static IEnumerable<Tensor> jacobian(Func<Tensor[], Tensor[]> function, params Tensor[] inputs)
+                {
+                    List<Tensor> jacobians = new List<Tensor>();
+                    using (enable_grad())
+                    {
+                        inputs = _grad_preprocess(need_graph: true, inputs);
```
NiklasGustafsson, line 79: It may be worth having a dispose scope around the body of this method.

src/TorchSharp/Functional.cs
```diff
@@ -0,0 +1,114 @@
...
+                ///   [1, 5]]]
+                /// And jacobian[1]:
+                /// [[1, 3],
+                ///  [2, 4]]
+                ///  
+                /// </summary>
+                /// <param name="function">The function to compute the jacobian for. It may have multiple inputs and multiple outputs.</param>
+                /// <param name="inputs">The values for the inputs at which to compute the jacobian.</param>
+                /// <returns>A list of tensors, one for each output value of each output. For example, a single output with two elements yields two jacobian tensors.</returns>
+                public static IEnumerable<Tensor> jacobian(Func<Tensor[], Tensor[]> function, params Tensor[] inputs)
```
NiklasGustafsson, line 74: This is very general, but consider whether there are other overloads that would be useful here.

NiklasGustafsson (changes requested):
Also, please add a few unit tests.

roelofsmn:
Thanks for the feedback! Currently, I am away, but I'll get on it as soon
as I get back!

roelofsmn:
I'm getting an error when building the solution:

```
Calling "...\TorchSharp\src\Native\\gen-buildsys-win.bat" "...\TorchSharp\src\Native\" "17 2022" x64
  "C:\Program Files\CMake\bin\cmake.exe" "-DCMAKE_BUILD_TYPE=Debug" "-DCMAKE_INSTALL_PREFIX=.../TorchSharp/src/Native/../../bin/x64.Debug/Native" "-DLIBTORCH_PATH=...\TorchSharp\bin/obj/AnyCPU.Debug\libtorch-cpu\libtorch-win-shared-with-deps-debug-2.1.0cpu\libtorch\share\cmake\Torch" -G "Visual St
  udio 17 2022"  -A x64 -B. -H"...\TorchSharp\src\Native\"
  CMake Warning (dev) in CMakeLists.txt:
  -- Configuring incomplete, errors occurred!
  Failed to generate native component build project!
    No project() command is present.  The top-level CMakeLists.txt file must
    contain a literal, direct call to the project() command.  Add a line of
    code such as

      project(ProjectName)

    near the top of the file, but after cmake_minimum_required().

    CMake is pretending there is a "project(Project)" command on the first
    line.
  This warning is for project developers.  Use -Wno-dev to suppress it.

  CMake Warning (dev) in CMakeLists.txt:
    cmake_minimum_required() should be called prior to this top-level project()
    call.  Please see the cmake-commands(7) manual for usage documentation of
    both commands.
  This warning is for project developers.  Use -Wno-dev to suppress it.

  CMake Error in CMakeLists.txt:
    Generator

      Visual Studio 17 2022

    could not find any instance of Visual Studio.



...\TorchSharp\src\Native\build.proj(53,5): error MSB3073: The command ""...\TorchSharp\src\Native\build.cmd" Debug x64 --libtorchpath ...\TorchSharp\bin/obj/AnyCPU.Debug\libtorch-cpu\libtorch-win-shared-with-deps-debug-2.1.0cpu\libtorch\share\cmake\Torch" exited with code
1.

Build FAILED.

...\TorchSharp\src\Native\build.proj(53,5): error MSB3073: The command ""...\TorchSharp\src\Native\build.cmd" Debug x64 --libtorchpath ...\TorchSharp\bin/obj/AnyCPU.Debug\libtorch-cpu\libtorch-win-shared-with-deps-debug-2.1.0cpu\libtorch\share\cmake\Torch" exited with code
1.
    0 Warning(s)
    1 Error(s)

Time Elapsed 00:00:45.15
```
Do you have an idea of the problem?

I'm on Windows 10, with Visual Studio 2022 17.7.0

NiklasGustafsson:
Hmm, I don't know. I would start by cleaning up the build environment, start afresh with only files from your repo:

```
git clean -xfd .
```

NiklasGustafsson (changes requested):
This looks good to me. One thing that needs to be fixed is that there are two test projects, one that includes CUDA binaries when running locally. Since you added a new file, you need to add it to the other test project, too, with "Add as Link".

PR #1197, opened 2024-01-14
Adding an option to attach a tensor to a different dispose scope

shaltielshmid:
Following the previous PR (a few weeks ago), we talked about the option for attaching a tensor to a different dispose scope, in addition to moving to the outer and detaching completely.

### DIFF
--- a/src/TorchSharp/DisposeScope.cs
+++ b/src/TorchSharp/DisposeScope.cs
@@ -221,7 +221,7 @@ namespace TorchSharp
         {
             if (disposable is torch.Tensor tensor) {
                 if (tensor.OwningDisposeScope == null && !tensor.IsInvalid) {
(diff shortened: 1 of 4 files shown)


NiklasGustafsson:
@shaltielshmid, can you elaborate on the scenario you're supporting with this?

shaltielshmid:
The scenario is for example when we are copying a tensor to a parameter, and we want to move the dispose scope of the tensor to that of the parameter, like here:
image

Also - it was your suggestion here :-) https://github.com/dotnet/TorchSharp/pull/1193#discussion_r1432064659

NiklasGustafsson:
> Also - it was your suggestion here :-) [#1193 (comment)](https://github.com/dotnet/TorchSharp/pull/1193#discussion_r1432064659)

:-)

Issue #1198, opened 2024-01-14
CUDA not available when calling ` backwards` before using CUDA


shaltielshmid:
When I use TorchSharp and I perform a gradient calculation before loading in the CUDA binaries, then TorchSharp no longer recognizes that CUDA is available.

For example:
```cs
var lin = torch.nn.Linear(10, 1, false);
lin.forward(torch.rand(10)).backward();
Console.WriteLine(torch.cuda.is_available()); // False
```

On the other hand, if I call a CUDA function before the `backward()` call, it will work:

```cs
Console.WriteLine(torch.cuda.is_available()); // True
var lin = torch.nn.Linear(10, 1, false);
lin.forward(torch.rand(10)).backward();
Console.WriteLine(torch.cuda.is_available()); // True
```

I've traced this issue back to LibTorch, over [here](https://github.com/pytorch/pytorch/blob/HEAD/aten/src/ATen/detail/CUDAHooksInterface.cpp#L28-L29). The exact quote is:
```cs
// NB: The once_flag here implies that if you try to call any CUDA
// functionality before libATen_cuda.so is loaded, CUDA is permanently
// disabled for that copy of ATen.
```

@NiklasGustafsson How do you think this should be handled?

In PyTorch I'm pretty sure that they load in the CUDA binaries with the first `import torch` call. This would be easy to add to TorchSharp, but is that the desired behavior?

NiklasGustafsson:
Hmm... I think we just need to place a call to 'cuda.is_available()' on some path that is guaranteed to be run before any real work is done. A static initializer block (is that what they're called? I use them so rarely) somewhere in the `torch` static class, perhaps? What do you think, @shaltielshmid?

shaltielshmid:
I was actually thinking we can make torch.CUDA be a field (like torch.CPU, torch.META, etc.). This would require adding an if and replacing it with null, if it doesn't exist. 
Alternatively, a static initializer is also a great solution. It would guarantee that cuda would be initialized before any use of torch operations.

NiklasGustafsson:
We need something that isn't going to throw an exception if the CUDA backend isn't there. Calling cuda.is_available() is safe that way.

shaltielshmid:
I was thinking of this:
```cs
public static Device CUDA = TryInitializeDeviceType(DeviceType.CUDA) ? new Device(DeviceType.CUDA, -1) : null;
```

But I'm not set on it. Should I create a PR with the static initializer?

[...]

PR #1199, opened 2024-01-16
Added to always initialize CUDA 

shaltielshmid:
Fixes https://github.com/dotnet/TorchSharp/issues/1198.

I tested it out locally - I added the DLL and the libtorch-cuda package, and it worked. I then tried it with the libtorch-cpu package and I confirmed that it didn't crash. 

I was going to add a test for it - but in our CUDA tests we always check that cuda is available, and then that defeats the purpose since checking if it is available initializes it. 
Do you have any suggestions?

### DIFF
--- a/src/TorchSharp/Device.cs
+++ b/src/TorchSharp/Device.cs
@@ -4,7 +4,7 @@ using System;
 namespace TorchSharp
 {
     public static partial class torch
-    {
+    {        
         /// <summary>
         /// A torch.Device is an object representing the device on which a torch.Tensor is or will be allocated.
         /// </summary>
@@ -85,12 +85,12 @@ namespace TorchSharp
         /// <summary>
         /// Convenience declaration of a CPU device accessible everywhere.
         /// </summary>
-        public static Device CPU = InitializeDevice(new Device(DeviceType.CPU, -1));
+        public static readonly Device CPU = InitializeDevice(new Device(DeviceType.CPU, -1));
 
         /// <summary>
         /// Convenience declaration of a META device accessible everywhere.
         /// </summary>
-        public static Device META = InitializeDevice(new Device(DeviceType.META, -1));
+        public static readonly Device META = InitializeDevice(new Device(DeviceType.META, -1));
 
         /// <summary>
         /// Factory for a device object, following the Pytorch API.
--- a/src/TorchSharp/Device.cs
+++ b/src/TorchSharp/Device.cs
@@ -80,17 +80,17 @@ namespace TorchSharp
         /// <summary>
         /// Convenience declaration of a CUDA device accessible everywhere.
         /// </summary>
-        public static Device CUDA => InitializeDevice(new Device(DeviceType.CUDA, -1));
+        public static readonly Device CUDA = cuda.is_available() ? new Device(DeviceType.CUDA, -1) : null;
 
         /// <summary>
         /// Convenience declaration of a CPU device accessible everywhere.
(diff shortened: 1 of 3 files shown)


NiklasGustafsson:
> Fixes #1198.
> 
> I tested it out locally - I added the DLL and the libtorch-cuda package, and it worked. I then tried it with the libtorch-cpu package and I confirmed that it didn't crash.
> 
> I was going to add a test for it - but in our CUDA tests we always check that cuda is available, and then that defeats the purpose since checking if it is available initializes it. Do you have any suggestions?

Suggestions on how to testing for a failure of a failure-safe thing? :-)

shaltielshmid:
Was just about to write about that :)

shaltielshmid:
Okay, so there are two questions:

1] How do we test for this bug? Even if I add another cuda-only test file, there is no guarantee that that test will run before a different test which calls a cuda function and loads the binaries in. Do we wan to skip?

2] This silent bug - if someone uses torch.CUDA when initializing a tensor when cuda isn't available, then TorchSharp will just skip over it and load it into CPU. Is this behavior OK? If not, we should probably just implement it as a static initializer or as a different static field (we can have a private static field that isn't used `private static bool initCudaIsAvailable = torch.cuda.is_available();`).

shaltielshmid:
> Suggestions on how to testing for a failure of a failure-safe thing? :-)

Lol. Test locally? ;-)

NiklasGustafsson:
> Okay, so there are two questions:
> 
> 1] How do we test for this bug? Even if I add another cuda-only test file, there is no guarantee that that test will run before a different test which calls a cuda function and loads the binaries in. Do we wan to skip?
> 
> 2] This silent bug - if someone uses torch.CUDA when initializing a tensor when cuda isn't available, then TorchSharp will just skip over it and load it into CPU. Is this behavior OK? If not, we should probably just implement it as a static initializer or as a different static field (we can have a private static field that isn't used `private static bool initCudaIsAvailable = torch.cuda.is_available();`).

Yes, I think the static initializer should work.

shaltielshmid:
So as it turns out, torch already had a static initializer :)
I just added the initialize call to the initializer.

NiklasGustafsson:
Looks good, except the CUDA field is still not readonly.

shaltielshmid:
The CUDA is not a field :-)

NiklasGustafsson:
> torch.CUDA is not a field :-) I had initially made it a field - but then I reverted back

Reason? I thought that's what you wanted it to be?

shaltielshmid:
I did - and that had solved the issue. But the problem was that when we made it a field the following code no longer crashed:
```cs
var t = torch.rand(10, device: torch.CUDA);
```

(This was actually taken from a test)

shaltielshmid:
The issue is that this can cause silent crashes when previous it didn't. See for example the test over here: https://github.com/dotnet/TorchSharp/blob/main/test/TorchSharpTest/TestTorchTensor.cs#L3758

The question is whether silent handling is the way to go. Since when it isn't available we replace it with null which is a valid parameter to pass when creating a tensor (that just means the default device = cpu)

NiklasGustafsson:
Since it is a property, I think the result needs to be memoized (a problem that was present before your PR, too).

shaltielshmid:
Sure, added memoization now.

NiklasGustafsson:
> Sure, added memoization now.

Waiting for the build to finish, and then I'll merge.

Issue #1202, opened 2024-01-20
How to access the attributes of a model?


lintao185:
In TorchSharp, I have an Upsample object, and how can I retrieve the mode attribute of this object? In other words, in PyTorch, it's effortless to access it through .mode, but in TorchSharp, it seems to become quite challenging.

shaltielshmid:
It seems like those attributes aren't exposed in the regular interface.
@NiklasGustafsson I can add those attributes as read only properties in the class, or to have it call into the native object. Which do you think would be better?

lintao185:
@shaltielshmid Many thanks. Concurrently, I'd like to raise a question and propose an idea. Upon utilizing the register_forward_hook to register a hook, the retrieval of its returned value proves challenging. My proposition is whether we can potentially augment the fundamental nn.Module<> class with a property that is of the type ConcurrentDictionary<string, dynamic>. This property would serve as an extension point for attributes during runtime, allowing the return values from such hooks to be conveniently associated with this property.

NiklasGustafsson:
If they can ever be changed in the native object through other means, then it would have to call into that, but if that's not the case, then keeping them in the managed object is fine.

NiklasGustafsson:
The Upsample attributes will be available in the next release. Please let me know if there are other modules missing important attribute exposure.

lintao185:
I wonder, when constructing a model using the nn module, whether all the parameters passed into the constructor can be accessed through `model.`. Currently, most of these parameters are not accessible in this way.

PR #1206, opened 2024-01-23
Address issue #1202.

NiklasGustafsson:
Also: add capability to load a script module from a byte array.

### DIFF
--- a/src/TorchSharp/NN/Upsample.cs
+++ b/src/TorchSharp/NN/Upsample.cs
@@ -20,18 +20,18 @@ namespace TorchSharp
             /// <param name="size">Output spatial sizes</param>
             /// <param name="scale_factor">Multiplier for spatial size. Has to match input size</param>
             /// <param name="mode">The upsampling algorithm: one of 'nearest', 'linear', 'bilinear', 'bicubic' and 'trilinear'. Default: 'nearest'</param>
-            /// <param name="alignCorners">If true, the corner pixels of the input and output tensors are aligned, and thus preserving the values at those pixels.
+            /// <param name="align_corners">If true, the corner pixels of the input and output tensors are aligned, and thus preserving the values at those pixels.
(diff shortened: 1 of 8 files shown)


NiklasGustafsson:
@shaltielshmid -- could you take a quick look at this PR?

src/TorchSharp/NN/Upsample.cs
```diff
@@ -220,6 +226,21 @@ public override Tensor forward(Tensor tensor)
...
+            public ReadOnlySpan<long> size {
+                get { return _size is null ? null : new ReadOnlySpan<long>(_size!); }
+            }
+
+            public ReadOnlySpan<double> scale_factor {
+                get { return _scale_factor is null ? null : new ReadOnlySpan<double>(_scale_factor!); }
+            }
+
+            private long[]? _size;
+            private double[]? _scale_factor;
```
shaltielshmid, line 242: Is there a need for a separate field for the size and scale factor? Why not just store them as regular properties like the other fields?

shaltielshmid (approved):

shaltielshmid:
A general question: Do we want to expose these properties on all the classes?

For example: 
All the dropouts (e.g., `AlphaDropout`, `Dropout1d`) should expose `p` and `inplace` like in PyTorch.
`Flatten` with `start_dim` and `end_dim`
etc.

src/TorchSharp/NN/Upsample.cs
NiklasGustafsson: The C# compiler didn't like the ReadOnlySpan<T> as the property type when the getter/setters were automatically generated.

NiklasGustafsson:
If you put the list together, we can split the job of implementing them...

NiklasGustafsson:
Note that there are some properties that are actually modified in native code, and for those, we have to get them with a native call. The properties on BatchNorm, for example.

shaltielshmid:
Ah, good point! Will definitely keep that in mind.

PR #1211, opened 2024-01-26
Start implementing standard NN modules in managed code.

NiklasGustafsson:


NiklasGustafsson:
This is an attempt at refactoring some of the built-in modules to rely on functional APIs rather than C++ modules. It was started last year, but never merged. It was brought to the forefront by the identified need to make more module attributes available, which requires a lot of work, some of which would be duplicative of this.

It should have no impact on performance or functionality, except that more module attributes will be available.

Note that this is not a PR into main, but rather a creating of a new branch where we will collaborate on this.

GeorgeS2019:
@NiklasGustafsson 

Curious..why it is stopped and closed?

NiklasGustafsson:
Because it is superceded by #1214

PR #1213, opened 2024-01-30
Upgrade Examples from .NET 6 to .NET 8

jonas1ara:
I don't think that this PR needs a description

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -2,6 +2,15 @@
 
 Releases, starting with 9/2/2021, are listed with the most recent release at the top.
 
+# NuGet Version 0.102.5
+
+__Breaking Changes__:
+
+__API Changes__:
+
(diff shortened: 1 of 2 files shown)


jonas1ara:
@dotnet-policy-service agree

NiklasGustafsson:
@jonas1ara, thank you for your first PR! Your contribution is much appreciated, it's really nice to see people eager to help out!

A description is not necessary, but I need to think about whether upgrading to .NET 8.0 is right for the examples, especially the language version. It may tempt us to use features that earlier versions don't allow, and as long as we have to support .NET 6 and .NET FX 4.7, that may be risky... :-)

NiklasGustafsson:
Another thing -- in order to work, the azure_pipelines.yml file would have to be updated to install .NET 6 on the build machines.

PR #1220, opened 2024-02-04
Fix custom autograd function no grad required

shaltielshmid:


shaltielshmid:
The custom autograd function crashed when there were no elements with requires_grad in the input, or when there was no grad enabled at all. There were two issues:
1] The pointer to the node can optionally be null, but in the C++ code I only checked to see if it was an expired pointer, not for null - fixed.
2] When there was no graph to build, the C++ code instantly called the delete node function which disposes the managed node, so we had a situation where the code to dispose the shared pointer was being called twice (once from the main apply function and once from the dispose function), and for some reason even when I added a lock around it that didn't solve it - see my comment. Fixed now.

PR #1234, opened 2024-02-11
Add support for >2GB tensors via byte stream

shaltielshmid:


src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -404,6 +405,76 @@ internal void ValidateType(Type dotnetType)
...
+            /// </summary>
+            /// <param name="stream">Stream to read the bytes from</param>
+            /// <param name="bufferSize">The buffer size to use when reading from the stream</param>
+            public void ReadBytesFromStream(Stream stream, int bufferSize = 65_536)
+            {
+                long totalSize = NumberOfElements * ElementSize;
+
+                if (!is_contiguous()) throw new InvalidOperationException("SetBytes() called on non-contiguous tensor.");
+
+                unsafe {
```
shaltielshmid, line 455: Should we check for CPU here? 
We check in the bytes getter, but not the setter.

src/TorchVision/File.cs
```diff
@@ -58,6 +59,7 @@ public static void write_file(string filename, Tensor data)
             /// <param name="data">One dimensional <c>uint8</c> <cref>Tensor</cref>.</param>
             public static async void write_file_async(string filename, Tensor data)
             {
+                // Currently limited to 2GB - should we duplicate the code for WriteBytesToStream using async ops?
```
shaltielshmid, line 62: Comment should be deleted before merge - wrote it for @NiklasGustafsson

test/TorchSharpTest/TestTorchTensor.cs
```diff
@@ -8269,5 +8269,111 @@ Tensor nbins_ratio(int seed, int size)
                 }
             }
         }
+
+
+        [Fact(Skip = "Very heavy on the compute")]
```
shaltielshmid, line 8274: The tests are very heavy on the compute, so I put them in Skip (but I tested them myself)

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -404,6 +405,76 @@ internal void ValidateType(Type dotnetType)
                 }
             }
 
+
+            /// <summary>
+            /// Writes the bytes of the tensor to a stream. Useful for when tensors are >2GB.
+            /// </summary>
+            /// <param name="stream">Stream to write the bytes to</param>
+            /// <param name="bufferSize">The buffer size to use when writing to the stream</param>
+            public void WriteBytesToStream(Stream stream, int bufferSize = 65_536)
```
shaltielshmid, line 414: Unfortunately the Stream.Write overload which supports Spans was added in a later version of .NET, so we can't use it and therefore resort to using a buffer.

shaltielshmid:
Tagging the discussion https://github.com/dotnet/TorchSharp/discussions/1219

shaltielshmid:
@NiklasGustafsson Please don't merge this in just yet - I want to confirm that it works well with PyBridge.

shaltielshmid:
Benchmarking:

Comparing loading from and saving to a pre-allocated memory (to mitigate the disk speed factor). The model is around 11MB in size when saved to disk. The comparison is between the "old" (current) method and with the new proposed method, with the various buffer sizes (4096, 2048, 1024):

| Method  | Mean      | Error     | StdDev    | Allocated   |
|--------:|----------:|----------:|----------:|------------:|
| SaveOld | 7.760 ms  | 0.0828 ms | 0.0734 ms |    22.04 KB |
| LoadOld | 9.635 ms  | 0.1364 ms | 0.1276 ms | 10929.37 KB |
| Save4K  | 8.016 ms  | 0.0770 ms | 0.0720 ms |    62.27 KB |
| Load4K  | 8.099 ms  | 0.0855 ms | 0.0758 ms |    62.75 KB |
| Save2K  | 7.882 ms  | 0.1496 ms | 0.1470 ms |    42.26 KB |
| Load2K  | 7.935 ms  | 0.1165 ms | 0.1090 ms |    42.75 KB |
| Save1K  | 7.867 ms  | 0.1156 ms | 0.1081 ms |    32.27 KB |
| Load1K  | 7.934 ms  | 0.1237 ms | 0.1157 ms |    32.75 KB |

src/TorchSharp/Tensor/Tensor.cs
NiklasGustafsson: If it's going to blow up if it's not CPU, then it's better to check and throw an exception with good, specific information about the problem rather than some generic I/O exception later.

src/TorchVision/File.cs
NiklasGustafsson: OK

NiklasGustafsson:
> Benchmarking:
> 
> Comparing loading from and saving to a pre-allocated memory (to mitigate the disk speed factor). The model is around 11MB in size when saved to disk. The comparison is between the "old" (current) method and with the new proposed method, with the various buffer sizes (4096, 2048, 1024):

So, roughly the same load speed, slightly better save speed.

src/TorchVision/File.cs
shaltielshmid: Should we write an async function?

src/TorchSharp/Tensor/Tensor.cs
shaltielshmid: Sure, will add

shaltielshmid:
> @shaltielshmid -- you asked me to hold off on merging, so I haven't...

Good point! I'll confirm shortly about PyBridge.

Regarding the write_file_async method in TorchVision. Do we want to duplicate the WriteBytesToStream function with async operations?

shaltielshmid:
@NiklasGustafsson Just confirming you saw the previous message. I'm good to go with the merge as soon as we decide if to add an "async" version of the function.

NiklasGustafsson:
I saw it, but failed to recognize that it was holding up the PR. Yes, async would be doo for reading and writing.

shaltielshmid:
Okay so it's not super trivial - Spans aren't allowed in async methods, so this would require some playing around to get right. I think we should go ahead and add it to the backlog, unless you think it's important in which case I'll work on it on the next few days.

NiklasGustafsson:
Add it to the backlog.

shaltielshmid:
> Add it to the backlog.

Sounds good! The PR is then ready to merge from my perspective. I removed the comment already.

PR #1241, opened 2024-02-22
Added a fix for LinearLR scheduler, and updated behavior of all schedulers to match PyTorch

shaltielshmid:
When playing with the schedulers, I noticed that they didn't match the results coming from PyTorch, so I updated the behavior. 

Justifications:
- Fixing the parentheses in the LinearLR `get_lr` function: https://github.com/pytorch/pytorch/blob/HEAD/torch/optim/lr_scheduler.py#L590-L591

- Removing the loop which sets the InitialLearningRate in the `LRScheduler` constructor: In the PyTorch code [here](https://github.com/pytorch/pytorch/blob/HEAD/torch/optim/lr_scheduler.py#L45-L46) they don't force-set the initial_lr, but rather set it if it wasn't there yet. In TorchSharp the initial_lr value is always set, and therefore that loop shouldn't be called. With one scheduler it isn't an issue, but with multiple you have the issue that each scheduler re-sets the initial_lr, and then sets the `lr`, so we end up having an issue. 

- Introducing `get_closed_form_lr()` as per the PyTorch code: https://github.com/pytorch/pytorch/blob/HEAD/torch/optim/lr_scheduler.py#L158-L159

- Adjusting the scores on the tests - replicated using PyTorch: https://gist.github.com/shaltielshmid/ebf01989367709a9feac88e076dd312b

- Disabling the `check_lr` flag on the Test `TrainingSGDLinearLR`: The start_factor is 0.333, and the end_factor is 0.75, which means the LR should be increasing, not decreasing. 

Breaking changes (kind of):
- The LinearLR default value for `end_factor` was modified to `1.0` , to match PyTorch
- When calling `.step()` with an epoch number, the behavior will used the closed form LR instead of regular LR - to match PyTorch, but it will result in changes for anyone using the library. 

Wasn't sure which of these breaking changes to put in the release notes.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -8,6 +8,11 @@ __API Changes__:
 
 #1219: Added support for loading and saving tensors that are >2GB.<br/> 
 
+__Bug Fixes__:
+
+Fixed LinearLR scheduler calculation with misplaced parentheses<br/>
+Added `get_closed_form_lr` to scheduler to match PyTorch behavior when specifying epoch in `.step()`<br/>
+
 ## NuGet Version 0.101.6
 
 __API Changes__:
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -4,6 +4,10 @@ Releases, starting with 9/2/2021, are listed with the most recent release at the
 
 ## NuGet Version 0.101.7
 
+__Breaking Changes__:
+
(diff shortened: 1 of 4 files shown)


NiklasGustafsson (approved):

NiklasGustafsson:
> Wasn't sure which of these breaking changes to put in the release notes.

All of them. :-)

PR #1247, opened 2024-02-27
Added no grad in loadstatedict

shaltielshmid:


NiklasGustafsson:
> Should I start 0.102.1 in the release notes?

Yes, please!

NiklasGustafsson:
Since you're at it, could you fix the release not for 0.102.0 to note that it's an upgrade to 2.2.1, not 2.2.0?

shaltielshmid:
> Since you're at it, could you fix the release not for 0.102.0 to note that it's an upgrade to 2.2.1, not 2.2.0?

Done! :)

shaltielshmid:
Question: I see this line was added in the recent release:

```
Any code that checks whether a device is 'CUDA' and does something rather than checking that it isn't 'CPU' will now fail to work, since there is now support for the 'MPS' device on MacOS.<br/>
```

Are there cases of that inside the TorchSharp codebase, or is that only for other users?

NiklasGustafsson:
> Are there cases of that inside the TorchSharp codebase, or is that only in downstream uses?

There were cases of that in the TorchVision models, and in the TorchSharpExamples repo. After calling `RegisterComponents,` it would check for 'CUDA' and decide whether to call `to()`. What it should do is check for anything but `CPU` and move. This was specifically raised by the addition of the 'MPS' device on Apple Silicon builds.

PR #1248, opened 2024-02-29
Fixed issue with Module.load() with deflate stream

shaltielshmid:
I encountered a bug with Module.load(), when I gave it a deflate stream. Turns out that when you call `stream.Read(buffer, start, length)` there is no guarantee that it will read length even if there are more bytes in the stream. 

I had assumed that if there were bytes to read, then it would always read to the maximum. Turns out that `DeflateStream` it will sometimes return after having read a partial length, because of some internal logic. So even though I requested the next 1024 bytes, it can return 788 bytes and then in the next request return 1024 again.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -7,6 +7,7 @@ Releases, starting with 9/2/2021, are listed with the most recent release at the
 __Bug Fixes__:
 
 `module.load_state_dict()` throws error for in-place operation on a leaf variable that requires grad. <br/>
+`module.load()` with streams which don't read the requested # of bytes throws error. <br/>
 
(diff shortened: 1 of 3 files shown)


NiklasGustafsson (approved):

shaltielshmid:
@NiklasGustafsson I realized that ZipArchive isn't supported in NetFX so the test is going to fail then.

NiklasGustafsson:
What's the issue that the .NET 6 check is for?

shaltielshmid:
The tests failed on NetFX since ZipArchive doesn't exist in NetFX, I think

NiklasGustafsson (approved):

Issue #1249, opened 2024-03-01
torch.nn.functional.avg_pool1d is not working correctly


yueyinqiu:
Hello! I've found that `torch.nn.functional.avg_pool1d` is not working as expected.

Here is my test code:

```csharp
using TorchSharp;

var x = torch.zeros(5, 7, 128);
Console.WriteLine(x.metastr());
// [5x7x128], type = Float32, device = cpu

var y1 = torch.nn.functional.avg_pool1d(x, 2);
Console.WriteLine(y1.metastr());
// [5x7x127], type = Float32, device = cpu

var y2 = torch.nn.AvgPool1d(2).call(x);
Console.WriteLine(y2.metastr());
// [5x7x64], type = Float32, device = cpu
```

And in PyTorch:

```python
import torch

x = torch.zeros(5, 7, 128)
print(x.size())
# torch.Size([5, 7, 128])

y1 = torch.nn.functional.avg_pool1d(x, 2)
print(y1.size())
# torch.Size([5, 7, 64])

y2 = torch.nn.AvgPool1d(2)(x)
print(y2.size())
# torch.Size([5, 7, 64])
```

There seems to be something going wrong with `torch.nn.functional.avg_pool1d`, while `torch.nn.AvgPool1d` still works well.

shaltielshmid:
Thank you for reporting! The discrepancy between PyTorch and TorchSharp is because of the default value assigned to the `stride` parameter. I submitted a fix for the next version, but until then you can just specify the `stride` parameter to be the same as the `kernelSize` parameter to get the desired results. 

E.g.,
```cs
var y1 = torch.nn.functional.avg_pool1d(x, 2, 2);
```

NiklasGustafsson:
Fixed in v0.102.1.

Issue #1250, opened 2024-03-01
cstr and npstr for 0d tensors


yueyinqiu:
`cstr` (or `ToString(TensorStringStyle.CSharp)`) on a 0d tensor (scalar) will return an empty string, due to [the line 6575 in Tensor.cs](https://github.com/dotnet/TorchSharp/blob/3f771acc3bd0e1f2d7ad8792dc0bc7efcf58d644/src/TorchSharp/Tensor/Tensor.cs#L6575).

Meanwhile I also found [the line 6582](https://github.com/dotnet/TorchSharp/blob/3f771acc3bd0e1f2d7ad8792dc0bc7efcf58d644/src/TorchSharp/Tensor/Tensor.cs#L6582), which also tries to check `t.Dimensions == 0` and print the scalar value, so I think the current behavior is not expected.

(Also for `npstr`.)

![image](https://github.com/dotnet/TorchSharp/assets/18749772/5d77e171-f742-49fd-ae1e-b96d48d91478)

NiklasGustafsson:
Fixed in v0.102.0.

PR #1251, opened 2024-03-02
Fixing 1249 and 1250 (avg pool1d + 0D string conversion)

shaltielshmid:
The modification of the AvgPool1D stride parameter defaulting to kernelSize is as per the documentation here:
https://github.com/pytorch/pytorch/blob/main/torch/nn/functional.py#L354-L355
And the implementation of the module here:
https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/pooling.py#L549-L550

On the way, I also fixed the 0D tensor string printing. 

Fixes: https://github.com/dotnet/TorchSharp/issues/1249 and https://github.com/dotnet/TorchSharp/issues/1250.

@NiklasGustafsson  Question: The PyTorch parameter is named `kernel_size` whereas in TorchSharp it's written as `kernelSize`. Is it worth at this point to change the naming to match PyTorch?

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -6,7 +6,9 @@ Releases, starting with 9/2/2021, are listed with the most recent release at the
 
 __Breaking Changes__:
 
-The `stride` parameter in the `torch.nn.functional.avg_pool1d` call now defaults to `kernelSize` instead of 1, to match the PyTorch behavior. 
+The `kernelSize` parameter in the function and class of `AvgPool1D` was renamed to `kernel_size` to match PyTorch naming.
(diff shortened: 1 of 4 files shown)


NiklasGustafsson:
Yes, it's worth renaming. When we switched from .NET to Python naming conventions, we missed a lot of parameter names.

It needs to be listed as a breaking change, though.

shaltielshmid:
Sounds good. Will do later today

shaltielshmid:
> Yes, it's worth renaming. When we switched from .NET to Python naming conventions, we missed a lot of parameter names.
> 
> It needs to be listed as a breaking change, though.

Done. Listed it as a breaking change as well.

NiklasGustafsson (dismissed):

PR #1259, opened 2024-03-05
Moving more modules to C#.

NiklasGustafsson:
Instead of creating a native-code module object, this PR is starting to move the module logic to C#, when reasonable.

Also, add missing properties for various modules -- the two go hand in hand.


NiklasGustafsson:
@shaltielshmid -- this is a draft PR, but please take a look.

shaltielshmid:
Wow! Looks great!
I'll go into more depth later/tomorrow, but one thing I noticed - all the cases where you have a property with a private backed field needs a `[ComponentName]` attribute added, for the `to` function.

NiklasGustafsson:
Hmmm. Not sure about that -- `[ComponentName]` is cleaner, but the line that goes like:

```C#
ConditionallyRegisterParameter(nameof(weight), _weight);
```

in each setter should take care of the same thing.

shaltielshmid:
Well, not exactly. The magic that happens in the `to()` call looks for fields with that name, and it won't find anything because the backing field is actually named something else.

NiklasGustafsson:
It's looking for things that have been registered. The field magic happens in "RegisterComponents," not in `to()`, I believe.

NiklasGustafsson:
> It's looking for things that have been registered. The field magic happens in "RegisterComponents," not in `to()`, I believe.

Or maybe not... :-)

src/TorchSharp/NN/Bilinear.cs
```diff
@@ -8,49 +8,65 @@
...
+                        _weight?.Dispose();
+                        _weight = (value.DetachFromDisposeScope() as Parameter)!;
+                        ConditionallyRegisterParameter(nameof(weight), _weight);
+                    }
                 }
             }
+
+            [ComponentName(Name = "bias")]
+            private Parameter? _bias;
+            [ComponentName(Name = "weight")]
```
shaltielshmid, line 68: Maybe it should be
```cs
[ComponentName(Name = nameof(weight))]
```

To match the setter

dje-dev:
This work looks ambitious. I'm just wondering what the motivation is. Is there any impact expected on runtime performance or memory?

NiklasGustafsson:
> This work looks ambitious. I'm just wondering what the motivation is. Is there any impact expected on runtime performance or memory?

The intent is to move closer to where Python is and to mimic the logic -- there's \ a lot of PT modules with substantial amounts of Python code, now. It also makes it far easier to expose all the attributes on modules that are currently not accessible in TorchSharp.

Yes, ambitious, but also something that can be done incrementally, so it doesn't have to be a "boil the ocean" effort.

GeorgeS2019:
@NiklasGustafsson 
Could you comment how this PR would help in porting Pytorch-based libraries or framework to .NET? 

e.g. Torch.Text
In 2024, there are more and more .NET tokenization and other codes for working with deep NLP e.g. using onnx.

NiklasGustafsson:
> @NiklasGustafsson Could you comment how this PR would help in porting Pytorch-based libraries or framework to .NET?
> 
> e.g. Torch.Text In 2024, there are more and more .NET tokenization and other codes for working with deep NLP e.g. using onnx.

I have no idea whether it will help or not. The motivation for this PR is making it easier to maintain TorchSharp, nothing else. Call it paying off technical debt, if you will.

GeorgeS2019:
Using the term Technical Debt immediately helps to clarify the goal better. Thx

NiklasGustafsson:
Closing this while we're working on it. No need to trigger automatic builds until we're ready to merge.

NiklasGustafsson:
Re-opening to get some builds running. Still Work in Progress.

yueyinqiu:
I'm trying to deal with `PairwiseDistance`. I thought that the class should extends `ParamLessModule`, as `AvgPool1D` and some other classes do.

However, in PyTorch it is allowed to do like:

```python
import torch.nn.utils

x = torch.nn.AvgPool1d(1)
x.register_module("???", torch.nn.Linear(1, 1))
x.to(dtype=torch.complex128)

print(x.get_submodule("???").weight.dtype)  # torch.complex128
```

Although I think nobody will do this... but if we want it to be as consistent as possible...

I suppose at least we should disable `register_modules`, `register_parameters` and `register_buffer` on `ParamLessModule`.  (throwing an exception instead of working incorrectly)

yueyinqiu:
To be radical, we could make all the `register_xxx` to be internal. (Registering something extra on a built-in module? At leaset I won't do that.)

I think the only problem is that it will cause a great inconsistent with PyTorch. (However throwing an exception is also somewhat inconsistent. I'm not really sure where the boundary is.)

NiklasGustafsson:
You can always throw an exception and then relax the constraint later. I'm adding logic to enforce that submodules must also be parameter-less.

NiklasGustafsson:
> You can always throw an exception and then relax the constraint later. I'm adding logic to enforce that submodules must also be parameter-less.

Pushed these changes to GH.

src/TorchSharp/NN/ParamLessModule.cs
```diff
@@ -45,13 +65,30 @@ public abstract class ParamLessModule<T1, T2, T3> : nn.Module<T1, T2, T3>
             protected internal override nn.Module _to(DeviceType deviceType, int deviceIndex = -1) => this;
 
             protected internal override nn.Module _to(ScalarType dtype) => this;
+
+            public override void register_buffer(string name, Tensor tensor, bool persistent = true)
```
shaltielshmid, line 69: Perhaps it is worth adding sealed to the override?

GeorgeS2019:
@ shaltielshmid

Do you have the link to the todoList : what yet to be done?

I hope this PR continue

yueyinqiu:
@GeorgeS2019 https://github.com/dotnet/TorchSharp/pull/1271#issuecomment-2015230601

GeorgeS2019:
@shaltielshmid
@NiklasGustafsson 
@yueyinqiu 

It is not normal for @NiklasGustafsson  to be silent for so long without letting us know that he is on a long vacation.

@shaltielshmid => I hope you are in communication with him on how best to proceed with this PR.

NiklasGustafsson:
This PR is opened and closed while the ongoing work is tracked in my personal repo. Once ready for release, this PR will be reopened again.

yueyinqiu:
In TorchSharp, buffers are registered in `RegisterComponents`:

```csharp
using TorchSharp;

Console.WriteLine(new MyModel().state_dict().Count);
// 1

class MyModel : torch.nn.Module
{
    private Tensor aTensor;
    public MyModel() : base(nameof(MyModel))
    {
        aTensor = torch.zeros([]);
        this.RegisterComponents();
    }
}
```

However in PyTorch, they are not automatically registered:

```python
import torch.nn


class MyModel(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.a_tensor = torch.zeros([])


print(len(MyModel().state_dict()))
```

Should we also modify that?

yueyinqiu:
Meanwhile I suppose the call of submodule's `RegisterComponents` in `register_module` is strange. But some are true depending on that. https://github.com/dotnet/TorchSharp/issues/1272#issuecomment-2014120758 Should we also change that?

lintao185:
How is this task going?😊😊😊(☆▽☆)

NiklasGustafsson:
> How is this task going?😊😊😊(☆▽☆)

Ah, yes... I was pulled into other tasks (TorchSharp is a background activity for me), and I only found time for smallish things, while this is a big task. One of the smaller tasks was adding the non_blocking argument to a variety of `to()` APIs, which meant that I now have > 80 files with merge conflicts to resolve in the branch where this work is maintained. I'll get to it sometime soon, but I think it requires a day without meetings, to just focus on the monkey-editing task of resolving merge conflicts.

shaltielshmid:
@NiklasGustafsson I was also pulled into quite a few other tasks these past few weeks - let me know when you've resolved the merge conflicts, and I'll finish up the last few items on our list?

NiklasGustafsson:
I can probably set aside some time tomorrow to work in the merge conflicts. With a little bit of luck, it'll be pretty mechanical.

NiklasGustafsson:
I've taken care of the merge conflicts. I'm going to spend a bit of time on the default device issue that isn't completely working, it seems.

PR #1260, opened 2024-03-06
Update README.md

yueyinqiu:
- mention the 'bundled' packages in Download part
- use LibTorch (instead of lowercase libtorch) when it refers to the C++ backend

---

Hello! I'm having my first pull request here. I think editting the readme file could be a great try.

Actually I'm still facing some problems to build the native projects on my laptop. Is there an alternative way recommended to use the existing libtorch packages on nuget instead of rebuilding it?

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Please check the [Release Notes](RELEASENOTES.md) file for news on what's been u
 
 __TorchSharp is now in the .NET Foundation!__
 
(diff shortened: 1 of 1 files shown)


yueyinqiu:
@microsoft-github-policy-service agree

yueyinqiu:
@dotnet-policy-service agree

yueyinqiu:
umm... why could updating a readme file cause a compile error

yueyinqiu:
maybe I should do that in another branch so that I can create more pull requests at the same time

PR #1271, opened 2024-03-16
more properties of modules

yueyinqiu:
Hello! The problem happens when I was trying to deal with `fuse_linear_bn_eval`.

In PyTorch, it requires `copy.deepcopy` to copy the layers. We don't have that in csharp and my idea is just to use `torch.nn.Linear` to create a new one. (I think that could always work since `Linear` is a sealed class so it's impossible for the users to pass me some other customized things. That's great.)

To create it, I should provide `inputSize`, `outputSize`, `hasBias` and also `device` and `dtype`. However, those values cannot be directly get from the given `Linear`. Of course, for `Linear`, I could infer them from its `weight`, but meanwhile I also want the `eps` of a `BatchNorm1d`, which seems completely inaccessible currently.

So I'm creating this pull request to show a possible solution, with adding some properties in the module class. In the codes I prefer to infer the values from a tensor, and if cannot, it have to ask for an extra parameter in the constructor. (Actually I suggest to always use the latter way, keeping more things in csharp level.)

However this may touch some basic designs of TorchSharp, like whether other modules should also add similar properties, which may lead to a great amount of work. Thus the pull request is more for finding a solution, and is not that prepared to be merged.

### DIFF
--- a/src/TorchSharp/NN/Linear.cs
+++ b/src/TorchSharp/NN/Linear.cs
@@ -18,21 +18,8 @@ namespace TorchSharp
             {
             }
 
-            public long in_features {
-                get {
-                    var weight = this.weight;
-                    Debug.Assert(weight is not null);
-                    return weight.size(1);
-                }
-            }
-
-            public long out_features {
-                get {
-                    var weight = this.weight;
-                    Debug.Assert(weight is not null);
-                    return weight.size(0);
-                }
-            }
+            public long in_features => weight!.size(1);
+            public long out_features => weight!.size(0);
 
             public override Tensor forward(Tensor tensor)
             {
--- a/src/TorchSharp/NN/Linear.cs
+++ b/src/TorchSharp/NN/Linear.cs
@@ -1,5 +1,6 @@
 // Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
 using System;
+using System.Diagnostics;
 using static TorchSharp.torch;
 using static TorchSharp.torch.nn;
 using static TorchSharp.PInvoke.NativeMethods;
@@ -15,6 +16,22 @@ namespace TorchSharp
         {
             internal Linear(IntPtr handle, IntPtr boxedHandle) : base(handle, boxedHandle)
             {
+            }
+
+            public long in_features {
+                get {
+                    var weight = this.weight;
+                    Debug.Assert(weight is not null);
+                    return weight.size(1);
+                }
+            }
+
+            public long out_features {
+                get {
+                    var weight = this.weight;
+                    Debug.Assert(weight is not null);
+                    return weight.size(0);
+                }
             }
 
             public override Tensor forward(Tensor tensor)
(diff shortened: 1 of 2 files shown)


yueyinqiu:
why is here a complie error since I have asserted that is not null...

NiklasGustafsson:
@yueyinqiu -- take a look at the ongoing work over on https://github.com/NiklasGustafsson/TorchSharp/tree/unit, and see if that make sense. It's a big rewrite of a lot of the standard modules, including `Linear`

NiklasGustafsson:
> why is here a compile error since I have asserted that is not null...
> 
> ```cs
> var weight = this.weight;
> Debug.Assert(weight is not null);
> return weight.size(0);
> ```
> 
> so strange... well anyway i will just use `!` instead...

I don't believe the compiler pays attention to `Debug.Assert()` calls, which are no-ops in Release mode. As far as I know, it only pays attention to conditionals such as `if` and '?:'

yueyinqiu:
> > why is here a compile error since I have asserted that is not null...
> > ```cs
> > var weight = this.weight;
> > Debug.Assert(weight is not null);
> > return weight.size(0);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > so strange... well anyway i will just use `!` instead...
> 
> I don't believe the compiler pays attention to `Debug.Assert()` calls, which are no-ops in `Release` builds. `Assert` is not intended as a language-level specification of invariants, they're meant to raise a special kind of exception that lets you attach a debugger (which they do in `Debug` builds).
> 
> As far as I know, the C# compiler only pays attention to conditionals such as `if` and '?:'

Well however nullable reference is also a complie time feature which has no effect in runtime. Anyway `Debug.Assert` could work on my laptop and here I will use `!` instead. It does not bother too much.

yueyinqiu:
> BatchNorm1d

Wow that's awesome. I can't wait for that anymore. And perhaps I shall close this pr.

NiklasGustafsson:
> Can't wait for that anymore.

Does that mean you that you are excited about the PR, or that you really cannot wait, that you need it now?

yueyinqiu:
> > Can't wait for that anymore.
> 
> Does that mean you that you are excited about the PR, or that you really cannot wait, that you need it now?

To be honest, both a bit. I have found that the first commit could date back to more than one year ago. So I’m a bit worried about when it will be merged.
﻿
But don't be nxious about it. I don’t need it to implement anything in my project because I could always get the information elsewhere, since the modules are often created by myself.
﻿
Meanwhile a careful consideration and test is really important so I doesn't mean to urge that. If here are something that I could do, I'm glad to help.

NiklasGustafsson:
Yeah, we started this a long time ago, and let it sit because it was only about re-engineering. Then, however, the attribute issue came up, and then it became a functional as well as "technical debt" problem, and we started doing a lot more work. You'll see that most of the commits are very recent.

If you want to help test the preliminary work, or contribute to some of the missing operations, that would be great. We're using the 'unit' (don't ask...) branch in my private repo as the integration point. 

@shaltielshmid set up this spreadsheet to assign and track the work:

https://docs.google.com/spreadsheets/d/1IKEX-O4mgzQWjb3uTqYkRPrtppw8sk81kuXYtBa6SRg/edit#gid=0

PR #1280, opened 2024-04-04
correct torch.finfo

### LINKED ISSUE #1279: torch.finfo(dtype).min is 0 when dtype is torch.float16 while in pytorch the min value is -65504.0
torchsharp version: 0.101.5

yueyinqiu:
fixes #1279 

1. add support for `Float16` and `BFloat16`
2. correct the result (`eps` should not be equal to c# `Epsilon`, and meanwhile `tiny` is not `eps`.)
3. add more fields in `FInfo`
4. `torch.set_default_dtype` will no longer accept non-floating types

Note: `Categorical.entropy`, `_CorrCholesky.check`, `Distribution.ClampProbs`, `FisherSnedecor.rsample`, `Gamma.rsample`, `Geometric.rsample`, `distributions.Gumbel`, `Laplace.rsample`, `SigmoidTransform._call` and `SigmoidTransform._inverse` are influenced. **I'm not sure whether they could work correctly now.**

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -1,6 +1,13 @@
 ## TorchSharp Release Notes
 
 Releases, starting with 9/2/2021, are listed with the most recent release at the top.
+
+# NuGet Version 0.102.4
+
+__Breaking Changes__:
+
+Correct `torch.finfo`. (`torch.set_default_dtype`, `Categorical.entropy`, `_CorrCholesky.check`, `Distribution.ClampProbs`, `FisherSnedecor.rsample`, `Gamma.rsample`, `Geometric.rsample`, `distributions.Gumbel`, `Laplace.rsample`, `SigmoidTransform._call` and `SigmoidTransform._inverse` are influenced.)
+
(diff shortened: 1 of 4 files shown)


yueyinqiu:
The values are generated by `repr` in python:

```python
import torch.nn.utils

types = [
    torch.bfloat16,
    torch.float16,
    torch.float32,
    torch.float64
]

for t in types:
    a = torch.finfo(t)
    print(f"bits = {repr(a.bits)},")
    print(f"eps = {repr(a.eps)},")
    print(f"max = {repr(a.max)},")
    print(f"min = {repr(a.min)},")
    print(f"tiny = {repr(a.tiny)},")
    print(f"smallest_normal = {repr(a.smallest_normal)},")
    print(f"resolution = {repr(a.resolution)}")
    print()
```

I have added a unit test to show that the new `eps` are correct. (They're not c# `Epsilon`.) Currently I haven't find a way to test `tiny` and `resolution` but I believe they will be correct since other values are correct.

PR #1282, opened 2024-04-06
Modify LangVersion of TorchSharpTest

yueyinqiu:
I have found out why [this](https://github.com/dotnet/TorchSharp/pull/1262#issuecomment-1998669584) was happening.


yueyinqiu:
Although `dotnet 6` is corresponding to `c# 9.0`, some features in c# 10 have been used in that project, so I set it to `10.0`.

Or perhaps we could make all the projects to use the latest language version that is supported on CI/CD machine?

PR #1289, opened 2024-04-17
in exportsd.py, call detach() before numpy()

yueyinqiu:
An error occurred when trying to save the downloaded pretained model:

```python
states = weight.get_state_dict()
exportsd.save_state_dict(states, file)
```

(where `weight` is something like `torchvision.models.AlexNet_Weights.IMAGENET1K_V1`)

([full codes here](https://huggingface.co/spaces/yueyinqiu/vision-TorchSharp-generator/tree/main))

### DIFF
--- a/src/Python/exportsd.py
+++ b/src/Python/exportsd.py
@@ -37,7 +37,7 @@ def _write_tensor(t, stream):
     stream.write(leb128.u.encode(len(t.shape)))
     for s in t.shape:
         stream.write(leb128.u.encode(s))
-    stream.write(t.numpy().tobytes())
+    stream.write(t.detach().numpy().tobytes())
 
 def write_tensor(t, file_name):
     f = open(file_name, "wb")
--- a/src/Python/exportsd.py
+++ b/src/Python/exportsd.py
@@ -37,7 +37,7 @@ def _write_tensor(t, stream):
     stream.write(leb128.u.encode(len(t.shape)))
     for s in t.shape:
         stream.write(leb128.u.encode(s))
-    stream.write(t.detach().numpy().tobytes())
+    stream.write(t.detach().cpu().numpy().tobytes())
 
 def write_tensor(t, file_name):
     f = open(file_name, "wb")


yueyinqiu:
I'm not sure why the loaded tensors are in a computation graph, but the error did occur and it could be solved by adding `detach`...

GeorgeS2019:
@yueyinqiu 
/Are we missing unit tests to check for this bug?

NiklasGustafsson:
> @yueyinqiu /Are we missing unit tests to check for this bug?

Right. We have no unit tests for the Python export / import code at all. I've just assumed it works... :-)

yueyinqiu:
Actually I think a simple unit test could not find this out.

Also I'm not really sure whether we should modify this, since we do not use `cpu()` as well. Perhaps this should be done by users before saving.

In most cases the problem does not occur and I don't know why the pretrained model are in computation graphs... Perhaps more check should be done before merging.

NiklasGustafsson:
Calling detach seems like a safe, harmless thing to do.

yueyinqiu:
just to make a note here: `torchvision.models.AlexNet_Weights.IMAGENET1K_V1.get_state_dict()['features.0.weight'].requires_grad` is True

yueyinqiu:
> Calling detach seems like a safe, harmless thing to do.

However it seems that `torch.save`/`torch.load` will record whether it `requires_grad` or not in PyTorch.

yueyinqiu:
Well I think it's fine. The .net format does not save `requires_grad` at all. And the current PyTorch's `state_dict()` implemention is using `detach()` before putting the tensor into the state dictionary, so definitely `requires_grad` won't be accessed when loading the state dictionary. Perhaps it's a behavior exists in some old PyTorch versions.

NiklasGustafsson:
Ready to merge?

yueyinqiu:
> Ready to merge?

I suppose yes

NiklasGustafsson (approved):

PR #1293, opened 2024-04-26
grad property

### LINKED ISSUE #1291: should `Tensor.grad` be a property instead of a method?

yueyinqiu:
close #1291

### DIFF
--- a/src/TorchSharp/Tensor/Tensor.cs
+++ b/src/TorchSharp/Tensor/Tensor.cs
@@ -1351,7 +1351,6 @@ namespace TorchSharp
                     return new Tensor(res);
                 }
                 set {
-                    value?.DetachFromDisposeScope();
                     NativeMethods.THSTensor_set_grad(Handle, value?.Handle ?? IntPtr.Zero);
                     CheckForErrors();
                 }
--- a/src/TorchSharp/Tensor/Tensor.cs
+++ b/src/TorchSharp/Tensor/Tensor.cs
@@ -1340,25 +1340,21 @@ namespace TorchSharp
             /// This attribute is null by default and becomes a Tensor the first time a call to backward() computes gradients for the tensor.
             /// The attribute will then contain the gradients computed and future calls to backward() will accumulate (add) gradients into it.
             /// </summary>
-            public Tensor? grad()
-            {
-                var res = NativeMethods.THSTensor_grad(Handle);
-                CheckForErrors();
+            public Tensor? grad {
+                get {
+                    var res = NativeMethods.THSTensor_grad(Handle);
+                    CheckForErrors();
 
-                if (res == IntPtr.Zero)
-                    return null;
+                    if (res == IntPtr.Zero)
+                        return null;
 
-                return new Tensor(res);
-            }
-
-            /// <summary>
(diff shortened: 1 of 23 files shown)


yueyinqiu:
Should we also deal with the circular reference problem?

```csharp
using TorchSharp;

for (; ; )
{
    var t = torch.zeros(1000);
    t.grad = t;
    // t.set_grad(t);
}
```

PyTorch does not allow that:

image

However it's still ok if you play the trick with two tensors:

```python
import torch

t = torch.zeros(1000)
g = torch.zeros(1000)
t.grad = g
g.grad = t
```

I'm really not sure about that...

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -1340,25 +1340,20 @@ public Tensor pin_memory()
...
-            /// </summary>
-            /// <param name="grad">The new gradient tensor</param>
-            public void set_grad(Tensor grad)
-            {
-                NativeMethods.THSTensor_set_grad(Handle, grad?.DetachFromDisposeScope().Handle ?? IntPtr.Zero);
-                CheckForErrors();
+                    return new Tensor(res);
+                }
+                set {
+                    NativeMethods.THSTensor_set_grad(Handle, value?.Handle ?? IntPtr.Zero);
```
shaltielshmid, line 1354: Why did you remove the detach from dispose scope?
You also removed the MoveToOtherDisposeScope which was listed above in the `to` function.

yueyinqiu: The disposal of the given tensor (`value`) will just delete that pointer, and won't influence this tensors' `grad`.

I have added a unit test to show that, in [TestTorchTensorBugs.cs, line 1686](https://github.com/dotnet/TorchSharp/pull/1293/files#diff-1f04e1c804d5f49ad256443abd868c0e382e907ba5c738f67e8a5247ccbd6304R1686). (My fault, it's hard to find the test since there are so many changes...)

yueyinqiu: Yeah you have noticed me that the one in `_toEpilog` should be directly disposed after the assignment.

yueyinqiu: Ah wait wait... Perhaps I have misunderstand something...

yueyinqiu: It seems that I have misunderstand something...

But now I'm considering something like `alias`. Does it make sense to use `alias` here and let the a tensor's `grad` to be disposed together with this "owner"? Or the original approach is just fine?

yueyinqiu:
Perhaps I shall create a new pull request for this...

yueyinqiu:
#1299

src/TorchSharp/Tensor/Tensor.cs
yueyinqiu: @shaltielshmid

I have reviewed it again and I'm now more confident about the removal. Could you please check [my comment here](https://github.com/dotnet/TorchSharp/pull/1299#discussion_r1581732019)?

PR #1296, opened 2024-04-26
implicit cast from Tensor to TensorIndex

### LINKED ISSUE #1266: Ranges implementation
Hello,
Hope this not an error with my code...
So I really like to use ranges. Lets assume tensor of 3 dimensions.
I can do x[some tensor] ....
I can do x[x<=2] ....
I can do x[..int,..,..] .....
What I cannot do is x[..,some tensor,..] ... where some tensor are indices or booleans.
Is where any way to implement this. Would be really cool because this often appears in pythonic code...

Thanks

yueyinqiu:
closes #1266

---

**It includes some aggressive changes. Careful checks are required.**

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -10,7 +10,7 @@ __API Changes__:
 
 - #1266 Implicit operator from `Tensor` to `TensorIndex` has been added.
     - The implicit operator from `TensorIndex` to `Tensor` has been removed.
-    - The indexer that accepts tensors as the indices has been removed.
+    - The indexers that accepts tensors/longs as the indices have been removed.
 
 __Bug Fixes__:
 
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -8,6 +8,10 @@ __Breaking Changes__:
 
 __API Changes__:
 
+- #1266 Implicit operator from `Tensor` to `TensorIndex` has been added.
+    - The implicit operator from `TensorIndex` to `Tensor` has been removed.
+    - The indexer that accepts tensors as the indices has been removed.
+
 __Bug Fixes__:
 
 
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -10,7 +10,7 @@ __API Changes__:
 
 - #1266 Implicit operator from `Tensor` to `TensorIndex` has been added.
     - The implicit operator from `TensorIndex` to `Tensor` has been removed.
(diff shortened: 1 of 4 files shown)


yueyinqiu:
- I have removed the implicit operator from `TensorIndex` to `Tensor`. I don't understand why it should exist to just throw an exception... **but the behavior is even tested in the unit tests**.
- I have removed the indexer `this[params Tensor[] indices]` since tensors could now be implicitly cast to TensorIndex. **However if there is something else, say a user defined class `A`, that can be implicitly cast to a tensor, it used to be able to be used as the indices, but now it couldn't because it won't be automatically cast twice.**

yueyinqiu:
I will make it ready for review, but please check the above sequences before merging.

yueyinqiu:
**1. add `implicit operator TensorIndex(Tensor tensor)` into `TensorIndex`**

This is what we want.

**2. remove `implicit operator Tensor(TensorIndex value)` of `TensorIndex`**

This operator is just throwing an exception. I don't know why it should exist, but this is even tested in [a unit test](https://github.com/dotnet/TorchSharp/pull/1296/commits/8da9a66267423cf1cf13afa8400fb18f518101b5#diff-fe070c2c72eeb3aa8996dbdcc289118c2dcb37cd10060c408297621f5ebac129L3190).

**3. remove `Tensor this[params Tensor[] indices]` of `Tensor`**

Well... Now I suppose it shouldn't be removed... We need an override for a array.

**4. remove `Tensor this[params Tensor[] indices]` of `Tensor`**

Similar to 3.

NiklasGustafsson:
> This operator is just throwing an exception.

I added it to make sure that some non-C# language (who knows?) doesn't apply different rules for implicit conversions.

yueyinqiu:
> > This operator is just throwing an exception.
> 
> I added it to make sure that some non-C# language (who knows?) doesn't apply different rules for implicit conversions.

Sorry, I haven't catch it. So seems that we could remove this now? Or we shall still keep it?

yueyinqiu:
What??? A unit test failed because of the removal of that operator?

NiklasGustafsson:
Implicit operators are complicated to get right, and since they're invisible in the code that uses them, hard to debug. I actually had to ask Mads Torgersen (lead PM on C#) to help me get this right. The operator you are trying to remove was important to declare but should never be called, as I remember it.

I'd rather we left these operators alone.

yueyinqiu:
Maybe I shall close this PR since it's behind the current version too much.

PR #1299, opened 2024-04-26
property grad

### LINKED ISSUE #1291: should `Tensor.grad` be a property instead of a method?

yueyinqiu:
close #1291

### DIFF
--- a/src/TorchSharp/NN/Module.cs
+++ b/src/TorchSharp/NN/Module.cs
@@ -240,7 +240,7 @@ namespace TorchSharp
 
                     foreach (var (name, param) in named_parameters(false).ToList()) {
                         if (!param.toWillCopy(dtype ?? param.dtype, device ?? param.device) &&
-                            (param.grad() is null || !param.grad().toWillCopy(dtype ?? param.dtype, device ?? param.device)))
+                            (param.grad is null || !param.grad.toWillCopy(dtype ?? param.dtype, device ?? param.device)))
                             continue;
 
                         Parameter p;
@@ -256,11 +256,11 @@ namespace TorchSharp
                                 .DetachFromDisposeScope() as Parameter;
 
                             // Copy the gradient over as well, if it exists
-                            var grad = param.grad();
+                            var grad = param.grad;
                             if (grad is not null) {
-                                p.set_grad(grad.to(paramType, device ?? param.device)
-                                                .with_requires_grad(grad.requires_grad)
-                                                .MoveToOtherDisposeScope(p));
+                                p.grad = grad.to(paramType, device ?? param.device)
+                                    .with_requires_grad(grad.requires_grad)
+                                    .MoveToOtherDisposeScope(p);
                             }
 
                             // Dispose the param and gradient
@@ -360,10 +360,10 @@ namespace TorchSharp
                     CheckForErrors();
 
                     foreach (var (_, p) in named_parameters()) {
-                        var grad = p.grad();
+                        var grad = p.grad;
                         if (grad is not null) {
                             if (set_to_none) {
-                                p.set_grad(null);
+                                p.grad = null;
                                 grad.DetachFromDisposeScope().Dispose();
                             } else {
                                 grad.zero_();
--- a/src/TorchSharp/NN/Module.cs
+++ b/src/TorchSharp/NN/Module.cs
@@ -254,14 +254,15 @@ namespace TorchSharp
                         // disable grad we would need to call .detach() on the moved tensor.
                         using (var d = torch.no_grad()) {
                             p = new Parameter(
-                                param.to(paramType, device ?? param.device).DetachFromDisposeScope(), param.requires_grad)
-                                .DetachFromDisposeScope() as Parameter;
+                                data: param.to(paramType, device ?? param.device),
+                                requires_grad: param.requires_grad);
+                            _ = p.DetachFromDisposeScope();
 
                             // Copy the gradient over as well, if it exists
                             if (grad is not null) {
-                                p.grad = grad.to(paramType, device ?? param.device)
-                                    .with_requires_grad(grad.requires_grad)
-                                    .MoveToOtherDisposeScope(p);
+                                using var newGrad = grad.to(paramType, device ?? param.device)
+                                    .with_requires_grad(grad.requires_grad);
+                                p.grad = newGrad;
                             }
 
                             // Dispose the param
--- a/src/TorchSharp/NN/Module.cs
+++ b/src/TorchSharp/NN/Module.cs
@@ -239,8 +239,10 @@ namespace TorchSharp
                                                             .ToDictionary(field => field.ComponentName());
 
                     foreach (var (name, param) in named_parameters(false).ToList()) {
+                        using var grad = param.grad;
+
                         if (!param.toWillCopy(dtype ?? param.dtype, device ?? param.device) &&
-                            (param.grad is null || !param.grad.toWillCopy(dtype ?? param.dtype, device ?? param.device)))
+                            (grad is null || !grad.toWillCopy(dtype ?? param.dtype, device ?? param.device)))
                             continue;
 
                         Parameter p;
@@ -256,16 +258,14 @@ namespace TorchSharp
                                 .DetachFromDisposeScope() as Parameter;
 
                             // Copy the gradient over as well, if it exists
-                            var grad = param.grad;
                             if (grad is not null) {
                                 p.grad = grad.to(paramType, device ?? param.device)
                                     .with_requires_grad(grad.requires_grad)
                                     .MoveToOtherDisposeScope(p);
                             }
 
-                            // Dispose the param and gradient
+                            // Dispose the param
                             param.Dispose();
-                            grad?.Dispose();
                         }
                         ConditionallyRegisterParameter(name, p);
 
@@ -360,11 +360,10 @@ namespace TorchSharp
                     CheckForErrors();
 
                     foreach (var (_, p) in named_parameters()) {
-                        var grad = p.grad;
+                        using var grad = p.grad;
                         if (grad is not null) {
                             if (set_to_none) {
                                 p.grad = null;
-                                grad.DetachFromDisposeScope().Dispose();
                             } else {
                                 grad.zero_();
                             }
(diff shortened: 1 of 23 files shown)


src/TorchSharp/NN/Module.cs
```diff
@@ -256,16 +258,14 @@ private void _toEpilog(ScalarType? dtype, Device device)
...
-                                                .MoveToOtherDisposeScope(p));
+                                p.grad = grad.to(paramType, device ?? param.device)
+                                    .with_requires_grad(grad.requires_grad)
+                                    .MoveToOtherDisposeScope(p);
                             }
 
-                            // Dispose the param and gradient
+                            // Dispose the param
                             param.Dispose();
-                            grad?.Dispose();
```
NiklasGustafsson, line 268: Does disposing the param automatically dispose the grad?

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -1340,25 +1340,21 @@ public Tensor pin_memory()
...
-            {
-                var res = NativeMethods.THSTensor_grad(Handle);
-                CheckForErrors();
-
-                if (res == IntPtr.Zero)
-                    return null;
+            public Tensor? grad {
+                get {
+                    var res = NativeMethods.THSTensor_grad(Handle);
+                    CheckForErrors();
```
NiklasGustafsson, line 1346: I know it was the same in the old code, but i think we should check for 'res == IntPtr.Zero' before checking for errors.

src/TorchSharp/NN/Module.cs
yueyinqiu: It's [line 242](https://github.com/dotnet/TorchSharp/pull/1299/files/5ac8a73e5f5247791dcdc3e6cae1d9cbf9bcf9eb#diff-c9fe99719d1977951a66f323175f5bcd3ba449f30a3711bbc84defa26e924460R242) disposing that.

src/TorchSharp/Tensor/Tensor.cs
yueyinqiu: It's [line 242](https://github.com/dotnet/TorchSharp/pull/1299/files/5ac8a73e5f5247791dcdc3e6cae1d9cbf9bcf9eb#diff-c9fe99719d1977951a66f323175f5bcd3ba449f30a3711bbc84defa26e924460R242) disposing that.

src/TorchSharp/NN/Module.cs
yueyinqiu: It's [line 242](https://github.com/dotnet/TorchSharp/pull/1299/files/5ac8a73e5f5247791dcdc3e6cae1d9cbf9bcf9eb#diff-c9fe99719d1977951a66f323175f5bcd3ba449f30a3711bbc84defa26e924460R242) disposing that.

src/TorchSharp/Tensor/Tensor.cs
yueyinqiu: I have modified it in [this commit](https://github.com/dotnet/TorchSharp/pull/1299/commits/4e88b62ab2f40e3648e3b6fc6d5e8fec88e136c4).

yueyinqiu:
fixed #1300

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -1340,25 +1340,22 @@ public Tensor pin_memory()
...
-            /// </summary>
-            /// <param name="grad">The new gradient tensor</param>
-            public void set_grad(Tensor grad)
-            {
-                NativeMethods.THSTensor_set_grad(Handle, grad?.DetachFromDisposeScope().Handle ?? IntPtr.Zero);
-                CheckForErrors();
+                    return new Tensor(res);
+                }
+                set {
+                    _ = value?.DetachFromDisposeScope();
```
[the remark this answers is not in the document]
yueyinqiu: It seems it will automatically make the copy. I'm not sure:

```csharp
using System.Reflection;

void DisposeButKeepHandle(Tensor tensor)
{
    var handle = tensor.Handle;
    tensor.Dispose();
    var f = tensor.GetType()
        .GetField("handle", BindingFlags.NonPublic | BindingFlags.Instance);
    f!.SetValue(tensor, handle);
}

var tensor = zeros([]);
var grad = zeros([]);
tensor.set_grad(grad);
DisposeButKeepHandle(grad);

Console.WriteLine(tensor.grad()! + 1);
// OK
Console.WriteLine(grad + 1);
// AccessViolationException: So the C++ tensor is true deleted.
// So tensor.grad (in the C++ level) is not the given grad?
```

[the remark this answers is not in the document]
yueyinqiu: Yes, I have viewed the C++ code. It seems that it will use this operator to assign the value:

```cpp
  Tensor& operator=(const Tensor &x) & {
    return operator=(static_cast<const TensorBase&>(x));
  }
```

which uses:

```cpp
  Tensor& operator=(const TensorBase& x) & {
    impl_ = x.getIntrusivePtr();
    return *this;
  }
```

So `tensor.grad` will always be a new tensor. We can safely delete the original one.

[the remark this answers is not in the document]
yueyinqiu: But how can we implement this behavior in Python:

```python
import torch

x = torch.zeros([])
y = torch.zeros([])

x.grad = y
print(id(y) == id(x.grad))
```

This is pretty strange. Perhaps PyTorch Tensor is using a different technique, instead of a simple pointer to LibTorch.

yueyinqiu:
Are there any suggestion about this behavior in PyTorch?

```python
import torch

x = torch.zeros([], requires_grad=True)
y = torch.zeros([])

x.grad = y
print(y is x.grad)  # True
# Same Python object???
```

This is pretty strange. Perhaps PyTorch tensors are using a different technique instead of a simple wrapper of a pointer to a LibTorch tensor. (Maybe this is very common for Python. Actually I don't know how Python interacts with C++.)

The only way I can find is to keep an extra field in C# level. When setting `grad`, assign the C# tensor to it, and when getting `grad`, always return the tensor.

NiklasGustafsson:
> Are there any suggestion about this behavior in PyTorch?
> 
> ```python
> import torch
> 
> x = torch.zeros([], requires_grad=True)
> y = torch.zeros([])
> 
> x.grad = y
> print(y is x.grad)  # True
> ```

To maintain this behavior seems to be of pretty marginal value, right?

yueyinqiu:
I suppose I will never rely on this, but someone may use it to create a reference to `x.grad`:

```python
import torch


class StrangeOptimizer:
    def __init__(self, target: torch.Tensor, grad: torch.Tensor):
        self._target = target
        self._grad = grad

    def step(self):
        self._target.add_(self._grad)


t = torch.zeros([])
g = torch.zeros([])
optim = StrangeOptimizer(t, g)

for _ in range(3):
    x = torch.zeros([], requires_grad=True)
    x.grad = g
    y = x + 1
    y.backward()
    optim.step()

print(t)  # tensor(6.)  (1 + 2 + 3)
```

Of course this could be replaced by other techniques, so I agree that this is unnecessary. I shall make this pr ready to review.

src/TorchSharp/Tensor/Tensor.cs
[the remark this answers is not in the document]
shaltielshmid: Looks good!

NiklasGustafsson (dismissed):

yueyinqiu:
The unit test seems to fail by chance.

Perhaps we need a seeded random for unit test :)

Would it be a good idea to include pythonnet and check the results with PyTorch?

NiklasGustafsson:
There’s some kind of concurrency issue that is affecting things. We could use seeded random numbers, too, but it would be a lot of work and not all APIs accept generator objects. Setting the global seed won’t work because of reentrancy.

Niklas

From: yueyinqiu ***@***.***>
Date: Wednesday, May 1, 2024 at 09:27
To: dotnet/TorchSharp ***@***.***>
Cc: Niklas Gustafsson ***@***.***>, Comment ***@***.***>
Subject: Re: [dotnet/TorchSharp] property grad (PR #1299)

The unit test seems to fail by chance.

Perhaps we need a seeded random for unit test :)

Would it be a good idea to include pythonnet and check the results with PyTorch?

yueyinqiu:
Yes, and I've found that torch's reproducibility system is not that reliable. I'm just considering something about C# `System.Random`, for our unit test alone.

NiklasGustafsson:
Hmmm. Tests are failing due to MNIST download. That has nothing to do with the PR, but I wonder if the source location has changed.

yueyinqiu:
I suppose it's a temporary problem. [train-images-idx3-ubyte.gz](http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz), [train-labels-idx1-ubyte.gz](http://yann.lecun.com/exdb/mnist/train-labels-idx1-ubyte.gz) and [t10k-images-idx3-ubyte.gz](http://yann.lecun.com/exdb/mnist/t10k-images-idx3-ubyte.gz) can be accessed. (But [t10k-labels-idx1-ubyte.gz](http://yann.lecun.com/exdb/mnist/t10k-labels-idx1-ubyte.gz) cannot.)

NiklasGustafsson:
It just keeps failing. Based on the stack trace, it's failing to download the first file, which is 'train-images-idx3-ubyte.gz' and when I click on the links above, it doesn't work. Maybe it's got something to do with where the requests come from? If the license allows it, we could add those files to the repo, instead. They're not big.

yueyinqiu:
I suppose the best solution is to use mirrors.

In PyTorch there is a `torchvision.datasets.MNIST.mirrors`, whose default value is:

```python
    mirrors = [
        "http://yann.lecun.com/exdb/mnist/",
        "https://ossci-datasets.s3.amazonaws.com/mnist/",
    ]
```

However `torchvision.datasets.MNIST` is not a class in TorchSharp. So

- should we add that in `Modules.MNIST`? But this is inconsistent with PyTorch;
- or make `torchvision.datasets.MNIST` a class? Then it can't be a method. Still inconsistent.
- or maybe we could just simply add an extra parameter? If it's me to make the decision, I would prefer this one.

yueyinqiu:
I have created a PR here #1308

yueyinqiu:
Perhaps it is. I have found a relevant issue in pytorch https://github.com/pytorch/vision/issues/7637

Issue #941, opened 2023-02-27
Multithreading safety


gevorgter:
Great library but please provide some information on multi threading safety.
In production use, models are normally wrapped into REST service which is inherently multithreaded. The Torch itself is safe to use for inference in that matter.
Not sure if your library keeping it this way. Major concern is introduction of scopes for memory management. 
Stack of scopes seems to me existing on global level which can be a problem. Not sure what else.

NiklasGustafsson:
> For the latter, we might just need to switch to `AsyncLocal<T>`. I believe it does work with both threads and async flows.

As always an excellent point, @lostmsu. The current solution accommodates thread parallelism, but not task-based concurrency.

NiklasGustafsson:
>> The Torch itself is safe to use for inference in that matter.

@gevorgter -- I have always wondered where, in all of the torch documentation, that safety is formalized into a guarantee. Do you have a reference?

gevorgter:
Sorry, no reference. But my hunch is that training might not be thread safe but inference is since there is no gradient to keep and it's just calculations.

Pytorch team appears to care about threading issues and fixing them.
https://github.com/pytorch/pytorch/issues/16828

alinpahontu2912:
@gevorgter, thank you for reporting this important issue regarding multithreading safety! We appreciate your insights on the implications of using TorchSharp in multithreaded environments, especially concerning memory management and scopes. 

Since it's been over two years, I wanted to check in to see if this issue is still relevant or if you’ve noticed any improvements in your use of the library. TorchSharp has evolved significantly during this time, and we’d love to hear if your concerns have been addressed or if there are still areas we should focus on. 

Your feedback is invaluable to us, so please let us know!

PR #1304, opened 2024-05-01
replace `assert`s in `RemoveDisposeScope`

### LINKED ISSUE #1303: in `RemoveDisposeScope`, we can't assume that `DisposeScopeStack.Peek() == disposeScope`
```csharp
using TorchSharp;

var d1 = torch.NewDisposeScope();
var d2 = torch.NewDisposeScope();
d1.Dispose();

var tensor = torch.zeros([]);
d2.Dispose();

tensor.print();
// It could print.
// So this tensor leaks.
```

Also:

```csharp
using TorchSharp;

var d = torch.NewDisposeScope();
d.Dispose();
d.Dispose();
// InvalidOperationException
// But exceptions are always not expected when disposing something. 
```

yueyinqiu:
close #1303

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -10,6 +10,7 @@ __API Changes__:
 
 __Bug Fixes__:
 
+- Now `DisposeScopes` could be correctly disposed in a wrong order.
 
 # NuGet Version 0.102.4
 
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -8,11 +8,14 @@ __Breaking Changes__:
 
 - `torchvision.dataset.MNIST` will try more mirrors.
     - The thrown exception might be changed when it fails to download `MNIST`, `FashionMNIST` or `KMNIST`.
+- `ObjectDisposedException` will now be thrown when trying to use the disposed dispose scopes.
+    - The constructor of dispose scopes is no longer `public`. Use `torch.NewDisposeScope` instead.
 
 __API Changes__:
 
 - #1291 `Tensor.grad()` and `Tensor.set_grad()` have been replaced by a new property `Tensor.grad`.
     - A potential memory leak caused by `set_grad` has been resolved.
+- `Include` method of dispose scopes has been removed. Use `Attach` instead.
 
 __Bug Fixes__:
 
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -10,7 +10,7 @@ __API Changes__:
 
 __Bug Fixes__:
 
-- Now `DisposeScopes` could be correctly disposed in a wrong order.
+- #1303 Allow dispose scopes to be disposed out of LIFO order.
 
 # NuGet Version 0.102.4
 
--- a/src/TorchSharp/DisposeScopeManager.cs
+++ b/src/TorchSharp/DisposeScopeManager.cs
@@ -18,7 +18,7 @@ namespace TorchSharp
         internal ThreadDisposeScopeStatistics StatisticsInstance { get; } = new ThreadDisposeScopeStatistics();
 
         internal static DisposeScopeManager ThreadSingleton => (_threadSingleton ??= new DisposeScopeManager());
-        internal Stack<DisposeScope> DisposeScopeStack { get; } = new();
+        internal List<DisposeScope> DisposeScopeStack { get; } = new();
 
         public static ThreadDisposeScopeStatistics Statistics => ThreadSingleton.StatisticsInstance;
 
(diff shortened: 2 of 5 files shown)


yueyinqiu:
Actually new items could still be added to the disposed `DisposeScope`s, which will influence the `StatisticsInstance`. Should we completely disallow it?

Or maybe we could directly add support for "custom" dispose scopes, without the attachment to the thread singleton manager. And the `DisposeScope`s will return to the custom ones when being disposed.

RELEASENOTES.md
```diff
@@ -10,6 +10,7 @@ __API Changes__:
 
 __Bug Fixes__:
 
+- Now `DisposeScopes` could be correctly disposed in a wrong order.
```
NiklasGustafsson, line 13: This comment doesn't make sense. If it's the wrong order, it's not correct.

src/TorchSharp/DisposeScopeManager.cs
```diff
@@ -18,7 +18,7 @@ public class DisposeScopeManager
         internal ThreadDisposeScopeStatistics StatisticsInstance { get; } = new ThreadDisposeScopeStatistics();
 
         internal static DisposeScopeManager ThreadSingleton => (_threadSingleton ??= new DisposeScopeManager());
-        internal Stack<DisposeScope> DisposeScopeStack { get; } = new();
+        internal List<DisposeScope> DisposeScopeStack { get; } = new();
```
NiklasGustafsson, line 21: Why is it better to use a list rather than a stack. Dispose scopes should always be LIFO.

yueyinqiu: @NiklasGustafsson I have first created an issue for this. #1303

RELEASENOTES.md
yueyinqiu: @NiklasGustafsson

Hmm... Sorry for my poor English. Could you please find a more accurate description for this?

NiklasGustafsson: Now that I understand the reasoning behind it (seem my comments in the issue), I think:

'Allow dispose scopes to be disposed out of LIFO order.'

src/TorchSharp/DisposeScopeManager.cs
```diff
@@ -42,15 +42,15 @@ internal static DisposeScope NewDisposeScope()
 
         internal void RemoveDisposeScope(DisposeScope disposeScope)
         {
-            Debug.Assert(DisposeScopeStack.Count > 0);
-            Debug.Assert(DisposeScopeStack.Peek() == disposeScope);
-            DisposeScopeStack.Pop();
+            var index = DisposeScopeStack.LastIndexOf(disposeScope);
+            if (index is not -1)
```
NiklasGustafsson, line 46: I think it should be an error (exception) if you are removing a dispose scope that isn't registered. The previous code treated it as an assertion, but I think the right behavior is to throw.

yueyinqiu: This method is called by dispose scope's `Dipose`. If the dispose scope is not in that stack, it means that it has been disposed previously, so doing nothing is what we expect.

yueyinqiu: And could you please check my comment [here](https://github.com/dotnet/TorchSharp/pull/1304#issuecomment-2088234426). This is also about the "relationship" between `DisposeScope` and `DisposeScopeManager`.

NiklasGustafsson: Yes, I note that it's an `internal` method, so it should be okay.

Anyway, I'm going to be busy with some private matters today. I'll merge the two PRs that were accepted later today.

NiklasGustafsson: BTW, since you have made significant contributions to the repo, it would be great if you would introduce yourself a little bit in this thread:

https://github.com/dotnet/TorchSharp/discussions/333

It's always nice to know a little bit about the people who build the software, and the people you work with on it.

The same goes for you, @shaltielshmid... :-)

yueyinqiu:
Shall we considering about async local here? https://github.com/dotnet/TorchSharp/issues/941#issuecomment-1446800519

yueyinqiu:
I shall finish this todo here. https://github.com/dotnet/TorchSharp/pull/1306/files#diff-059c1143f16c2894e06fd333909c11ae581bd17225c24e58d3e078949863dc56R339

NiklasGustafsson (approved):

PR #1306, opened 2024-05-03
tensor disposal in data loader

yueyinqiu:
Main changes:

- add `disposeBatch` option to indicate whether to dispose the collated tensors automatically
    - otherwise data loaders cannot work with some linq methods like `First`, because the returned tensors would be disposed
    - currently we are still using a list. if we allows ["custom" dispose scopes](https://github.com/dotnet/TorchSharp/pull/1304#issuecomment-2088234426), we could switch to this
- rewrite the enumerator

---

Other changes:

- collate functions will now dispose the intermediate tensors, rather than wait for the next iteration
- add `disposeDataset` option
- keep the aliases of the tensors in `TensorDataset`

---

Perhaps we need a document about custom datasets and dataloaders.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -2,6 +2,15 @@
 
 Releases, starting with 9/2/2021, are listed with the most recent release at the top.
 
+# NuGet Version 0.102.5
+
+__Breaking Changes__:
+
+__API Changes__:
+
+__Bug Fixes__:
+
+
 # NuGet Version 0.102.4
 
 __Breaking Changes__:
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -8,8 +8,15 @@ __Breaking Changes__:
 
 __API Changes__:
 
+- A new option `autoDispose` has been added into `DataLoader`s, which indicates whether to dispose the collated tensors before the next iteration.
+    - The default collate functions will now always dispose the intermediate tensors, rather than wait for the next iteration.
+- A new abstract method `DisposeTensor` has been added to `Dataset<>`s.
+    - This method will be used by `DataLoader`, to dispose the values provided by `GetTensor`.
+    - `Dataset` and `IterableDataset` has implemented this method, so if your dataset is inherited from them, please check whether the disposal should be avoided in your case.
+
 __Bug Fixes__:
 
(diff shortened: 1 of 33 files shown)


yueyinqiu:
The tensors got from `dataset.GetTensor` is not disposed. This could cause memory leaks when using some datasets, but might also cause a unexpected disposal. I would check that and use `alias` when returning those tensors.

Perhaps we need a document about custom datasets and dataloaders.

yueyinqiu:
data loader will also dispose the given dataset. so the name `autoDispose` seems to be not that clear.

yueyinqiu:
Found that the shuffle is disposed after the first iteration...

```csharp
using System.Collections;
using TorchSharp;

var dataset = torch.utils.data.TensorDataset(torch.zeros([3]));
var dataLoader = torch.utils.data.DataLoader(dataset, 1, new MyShuffler());
foreach (var _ in dataLoader) ; // ObjectDisposedException

class MyShuffler : IEnumerable<long>
{
    public IEnumerator<long> GetEnumerator() => new ShufflerEnumerator();
    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
    class ShufflerEnumerator : IEnumerator<long>
    {
        bool disposed = false;
        public long Current
        {
            get
            {
                if (disposed)
                    throw new ObjectDisposedException("");
                return 1;
            }
        }
        object IEnumerator.Current => Current;
        public void Dispose()
        {
            disposed = true;
        }
        public bool MoveNext() => true;
        public void Reset() { }
    }
}
```

yueyinqiu:
Hmmm... `drop_last` is also ignored:

```csharp
using TorchSharp;

var dataset = torch.utils.data.TensorDataset(torch.tensor(new long[] { 1, 2, 3 }));
Console.WriteLine(dataset.Count); // 3
var dataLoader = torch.utils.data.DataLoader(dataset, 2, drop_last: true);
Console.WriteLine(dataLoader.Count); // 1
Console.WriteLine(dataLoader.Count()); // 2
```

NiklasGustafsson:
Tests related to DataLoader are failing.

yueyinqiu:
Ah... I forgot to check whether it is coming to an end...

PR #1308, opened 2024-05-07
nmist mirrors

yueyinqiu:


test/TorchSharpTest/LinearAlgebra.cs
```diff
@@ -404,19 +404,19 @@ public void SolveTriangularTest()
                 var A = randn(3, 3).triu_();
                 var b = randn(3, 4);
                 var x = linalg.solve_triangular(A, b, upper: true);
-                Assert.True(A.matmul(x).allclose(b, rtol: 1e-03, atol: 1e-06));
```
yueyinqiu, line 407: Runned under the new tolerance for more than 100 times and it does not fail.

test/TorchSharpTest/LinearAlgebra.cs
```diff
@@ -404,19 +404,19 @@ public void SolveTriangularTest()
...
+                Assert.True(A.matmul(x).allclose(b, rtol: 1e-03, atol: 1e-05));
```
yueyinqiu, line 407: Runned under the new tolerance for more than 100 times and it does not fail.

NiklasGustafsson (approved):

PR #1309, opened 2024-05-08
more fixes

yueyinqiu:


NiklasGustafsson:
I love the energy you're putting into this, but please provide a description of what the PR is for.

src/TorchSharp/DataLoader.cs
```diff
@@ -412,35 +411,42 @@ public bool MoveNext()
                 {
                     DisposeCurrent();
 
-                    using (var scope = torch.NewDisposeScope()) {
```
yueyinqiu, line 315: I just realized that the dispose scope does not work on other threads. Instead dispose scopes should be separately created in the below `ForAll`.

src/TorchSharp/DisposeScope.cs
```diff
@@ -40,7 +40,7 @@ internal DisposeScope(DisposeScopeManager disposeScopeManager)
         public IReadOnlyList<IDisposable> DisposablesView {
             get {
                 if (this._disposeScopeManager is null)
-                    throw new ObjectDisposedException("The dispose scope has been disposed.");
+                    throw new ObjectDisposedException(this.GetType().FullName);
```
yueyinqiu, line 43: 3. The parameter of the constructor of `ObjectDisposedException` is not the message, but is something called objectName.

docfx/articles/memory.md
```diff
@@ -295,6 +295,107 @@ public override Tensor entropy() => torch.WrappedTensorDisposeScope(() => ((scal
 
 ```
 
+## Data loaders and datasets
```
yueyinqiu, line 298: 4. add some document about data loaders and datasets.

yueyinqiu:
@NiklasGustafsson 

I have added some comments on the codes. (4 in total)

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -7377,6 +7377,12 @@ public static long max_int_value(ScalarType type)
         public static DisposeScope NewDisposeScope() =>
             DisposeScopeManager.ThreadSingleton.NewDisposeScope();
 
+        /// <summary>
+        /// Get the current dispose scope for the current thread.
+        /// </summary>
+        public static DisposeScope? CurrentDisposeScope =>
```
yueyinqiu, line 7383: 5. expose the current dispose scope and add Attach method for ienumerable, which could be used with the not-disposeBatch data loaders to manually attach the tensors to the current scope

NiklasGustafsson:
@yueyinqiu -- ready to merge, or are there more changes coming?

yueyinqiu:
I think it's ready

PR #1318, opened 2024-05-31
Add a input-free version of pre- and post-hooks that can manipulate t…

NiklasGustafsson:
…he module itself.

### DIFF
--- a/src/TorchSharp/NN/Module.cs
+++ b/src/TorchSharp/NN/Module.cs
@@ -1216,7 +1216,7 @@ namespace TorchSharp
             /// <summary>
             /// Represents a module that accepts 'hook' to the module logic.
             /// </summary>
-            public class HookableModule<TPreHook,TPostHook> : Module
+            public class HookableModule<TModule,TPreHook,TPostHook> : Module
             {
                 protected HookableModule(string name) : base(name) { }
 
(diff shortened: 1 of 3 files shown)


src/TorchSharp/NN/Module.cs
```diff
@@ -1237,22 +1237,41 @@ public HookRemover register_forward_pre_hook(TPreHook hook)
                     return new HookRemover(this, key);
                 }
 
+                public HookRemover register_forward_hook(Action<TModule> hook)
+                {
+                    var key = Guid.NewGuid().ToString();
```
yueyinqiu, line 1242: Should we check whether the key exists now? I know that Guids are almost impossible to duplicate. But that might happen when two guid are generated at the same time...

NiklasGustafsson: It would pretty much have to happen at the same nano-second.

NiklasGustafsson: Also, since these keys are local to the module, you'd have to be registering a hook from two different threads at just the exact same moment.

NiklasGustafsson: If we want to allow multi-threaded construction of models, which I'm not convinced is necessary, then we should instead add locks around the register/remove code.

PR #1319, opened 2024-05-31
Adding 'set/get_default_device()'

### LINKED ISSUE #1317: How to set default device type in torchsharp.
In pytorch, you can set default device using the following api

- https://pytorch.org/docs/stable/generated/torch.set_default_device.html

But that API is missing in torchsharp

NiklasGustafsson:
Fixes #1317

### DIFF
--- a/src/TorchSharp/FFT.cs
+++ b/src/TorchSharp/FFT.cs
@@ -350,10 +350,6 @@ namespace TorchSharp
             /// <param name="requires_grad">If autograd should record operations on the returned tensor.</param>
             public static Tensor fftfreq(long n, double d = 1.0, torch.ScalarType? dtype = null, torch.Device device = null, bool requires_grad = false)
             {
-                if (device is null)
-                {
-                    device = get_default_device();
-                }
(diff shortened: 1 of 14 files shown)


src/TorchSharp/FFT.cs
```diff
@@ -350,6 +350,10 @@ public static Tensor ifftshift(Tensor input, long[] dim = null)
             /// <param name="requires_grad">If autograd should record operations on the returned tensor.</param>
             public static Tensor fftfreq(long n, double d = 1.0, torch.ScalarType? dtype = null, torch.Device device = null, bool requires_grad = false)
             {
+                if (device is null)
+                {
+                    device = get_default_device();
+                }
                 device = torch.InitializeDevice(device);
```
yueyinqiu, line 357: Should we also modify the behavior of `torch.InitializeDevice(null)`?

(And then the 
```csharp
if (device is null)
{
    device = get_default_device();
}
```
before is no need.)

NiklasGustafsson: Good idea!

PR #1328, opened 2024-06-11
Module hooks don't depend on the forward() signature.

NiklasGustafsson:
Add CopyTo/From overloads to TensorAccessor for Span<T>.

### DIFF
--- a/src/TorchSharp/Utils/TensorAccessor.cs
+++ b/src/TorchSharp/Utils/TensorAccessor.cs
@@ -241,7 +241,27 @@ namespace TorchSharp.Utils
             }
         }
 
+        public void CopyTo(Span<T> array, int arrayIndex = 0, long tensorIndex = 0)
+        {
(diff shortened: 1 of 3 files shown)


src/TorchSharp/Utils/TensorAccessor.cs
```diff
@@ -241,6 +241,16 @@ public void CopyTo(T[] array, int arrayIndex = 0, long tensorIndex = 0)
             }
         }
 
+        public void CopyTo(Span<T> array, int arrayIndex = 0, long tensorIndex = 0)
```
LittleLittleCloud, line 244: Is this change related

LittleLittleCloud (approved):

src/TorchSharp/Utils/TensorAccessor.cs
LittleLittleCloud: NVM

PR #1331, opened 2024-06-13
Support creating tensors from a Memory<T> instance.

NiklasGustafsson:


src/TorchSharp/Tensor/Factories/tensor_Complex.cs
```diff
@@ -271,5 +271,14 @@ public static Tensor tensor(System.Numerics.Complex[,,,] rawArray, ScalarType? d
         {
             return _tensor_generic(rawArray, stackalloc long[] { rawArray.GetLongLength(0), rawArray.GetLongLength(1), rawArray.GetLongLength(2), rawArray.GetLongLength(3) }, (sbyte)ScalarType.ComplexFloat64, dtype, device, requires_grad, names: names);
         }
+
+        /// <summary>
+        /// Create a tensor from an array of values, shaping it based on the shape passed in.
+        /// </summary>
+        [Pure]
```
LittleLittleCloud, line 278: What's `Pure` and what's the purpose of using `Pure` here.

Trying search the `Pure` attribution on MS learn

```
Indicates that a type or method is pure, that is, it does not make any visible state changes.
```

Which really doesn't help a lot here...

NiklasGustafsson: Well, no, but it's been systematically used on methods throughout the API.

src/TorchSharp/Tensor/Factories/Tensor.Factories.cs
```diff
@@ -134,7 +134,7 @@ public static Tensor normal(double mean, double std, ReadOnlySpan<long> size, Sc
         /// <returns>A constructed tensor with elements of `dtype`</returns>
         /// <exception cref="ArgumentException"></exception>
         private static Tensor _tensor_generic(Array rawArray, ReadOnlySpan<long> dimensions, sbyte origType, ScalarType? dtype, Device? device, bool requires_grad, bool clone = true, string[]? names = null)
-        {
+        {
```
LittleLittleCloud, line 137:
```suggestion
        {
```

LittleLittleCloud (approved):
Mainly review the test cases and it looks good to me. The change in `_tensor-generic` and `_tensor_generic<T>` looks make sense to me, but I'm not familiar with `PInvoke` and `unsafe` feature so these code might need further review

PR #1338, opened 2024-06-25
Revert the building container back to ubuntu 20.04

LittleLittleCloud:
unblock https://github.com/dotnet/machinelearning/issues/7182

~Need to wait the PR #1351 to be merged in first~
## Why not reverting back to ubuntu 22.04

The glibcxx in ubuntu 22.04 is 3.4.30 and is newer than the one in centos-streaming9 (3.4.29)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -20,7 +20,7 @@
     <SourceDir>$(RepoRoot)src/</SourceDir>
     <PkgDir>$(RepoRoot)pkg/</PkgDir>
 
-    <LibTorchPackageVersion>2.2.1.2</LibTorchPackageVersion>
+    <LibTorchPackageVersion>2.2.1.1</LibTorchPackageVersion>
 
     <!-- when building on local machines the massive downloads get placed up one directory -->
     <!-- so we can clean without triggering a re-download-->
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -20,7 +20,7 @@
     <SourceDir>$(RepoRoot)src/</SourceDir>
     <PkgDir>$(RepoRoot)pkg/</PkgDir>
 
-    <LibTorchPackageVersion>2.2.1.1</LibTorchPackageVersion>
+    <LibTorchPackageVersion>2.2.1.2</LibTorchPackageVersion>
 
     <!-- when building on local machines the massive downloads get placed up one directory -->
     <!-- so we can clean without triggering a re-download-->
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -87,6 +87,7 @@
   <!-- we manually update these -->
   <PropertyGroup Condition="'$(MSBuildProjectName.IndexOf(`libtorch-`))' != '-1'">
     <LibTorchPackageVersion>2.2.1.1</LibTorchPackageVersion>
+    <EnablePackageValidation>false</EnablePackageValidation>
     <VersionPrefix>$(LibTorchPackageVersion)</VersionPrefix>
     <VersionSuffix></VersionSuffix>
   </PropertyGroup>
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -490,7 +490,7 @@ jobs:
         ldd --version && (/sbin/ldconfig -p | grep stdc++) && (strings /usr/lib/x86_64-linux-gnu/libstdc++.so.6 | grep LIBCXX) &&
         sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys DE19EB17684BA42D &&
         sudo apt-get -y update &&
-        sudo apt-get -y install cmake clang git libunwind8 curl libomp-dev libomp5 &&
+        sudo apt-get -y install cmake clang-7.0 git libunwind8 curl libomp-dev libomp5 &&
         wget https://packages.microsoft.com/config/ubuntu/22.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb &&
         sudo dpkg --purge packages-microsoft-prod && sudo dpkg -i packages-microsoft-prod.deb &&
         sudo apt-get update; sudo apt-get install -y apt-transport-https && sudo apt-get update &&
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -114,6 +114,18 @@ jobs:
         rm -rf ~/miniconda3/miniconda.sh
     displayName: Install Miniconda
 
+  - task: Bash@3
+    inputs:
+      targetType: 'inline'
+      script: |
+        ~/miniconda3/bin/conda init bash
+        source ~/.bashrc
+        eval "$(~/miniconda3/bin/conda shell.bash hook)"
+        source activate base
+        echo "activate conda environment"
+        conda activate base
+        conda install cmake=3.22.1 clang=14.0.6 clangxx=14.0.6 gxx_linux-64=11.2.0
+    displayName: Install cmake, clang and gxx_linux-64 dependencies
   - script:
         ldd --version && (/sbin/ldconfig -p | grep stdc++) && (strings /usr/lib/x86_64-linux-gnu/libstdc++.so.6 | grep LIBCXX) &&
         sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys DE19EB17684BA42D &&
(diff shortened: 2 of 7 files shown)


azure-pipelines.yml
```diff
@@ -4,13 +4,40 @@
 
 name: $(TeamProject)_$(Build.DefinitionName)_$(SourceBranchName)_$(Date:yyyyMMdd)$(Rev:.r)
 
-variables:
-  MyRunNumber:  $[counter('MyRunNumber', 52201)]
-
+parameters:
```
LittleLittleCloud, line 7: Move this to parameters so we can change it in azure pipeline UI

image

azure-pipelines.yml
```diff
@@ -4,13 +4,40 @@
...
+    ldd --version && (/sbin/ldconfig -p | grep stdc++) && (strings /usr/lib/x86_64-linux-gnu/libstdc++.so.6 | grep LIBCXX)
+    sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys DE19EB17684BA42D
+    sudo apt-get -y update
+    sudo apt-get -y install cmake git libunwind8 curl libomp-dev libomp5
+    wget https://packages.microsoft.com/config/ubuntu/22.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb
+    sudo dpkg --purge packages-microsoft-prod && sudo dpkg -i packages-microsoft-prod.deb
+    sudo apt-get update; sudo apt-get install -y apt-transport-https && sudo apt-get update
+    ldd --version && (/sbin/ldconfig -p | grep stdc++) && (strings /usr/lib/x86_64-linux-gnu/libstdc++.so.6 | grep LIBCXX)
+
+    # Install Miniconda
```
LittleLittleCloud, line 29: Configure cmake/clang/gcc to match the version used by `win`/`macos` building machine using `miniconda` environment

azure-pipelines.yml
```diff
@@ -106,33 +125,39 @@ jobs:
...
+  - script: |
+      eval "$(~/miniconda3/bin/conda shell.bash hook)"
+      source activate base
+      dotnet build -c $(BuildConfig) src/Redist/libtorch-cpu/libtorch-cpu.proj /p:UpdateSHA=true /p:SkipTests=true /p:TargetOS=linux /t:Build /p:IncludeLibTorchCpuPackages=true
     condition: eq(variables['BuildLibTorchPackages'], 'true')
     displayName: Download libtorch native binaries
 
-  - script: dotnet build -c $(BuildConfig) src/Redist/libtorch-cuda-12.1/libtorch-cuda-12.1.proj /p:UpdateSHA=true /p:SkipTests=true /p:TargetOS=linux /t:Build /p:IncludeLibTorchCudaPackages=true
+  - script: |
+      eval "$(~/miniconda3/bin/conda shell.bash hook)"
```
LittleLittleCloud, line 139: activate conda environment when compile native code.

NiklasGustafsson:
Hey, if you can figure out how to re-enable the push to feed tasks, that would be great. It stopped working for me a while ago, I didn't have the right permissions, so I disabled it.

LittleLittleCloud:
> Hey, if you can figure out how to re-enable the push to feed tasks, that would be great. It stopped working for me a while ago, I didn't have the right permissions, so I disabled it.

Sure, I need to ask for permission to edit artifacts first though

NiklasGustafsson:
> > Hey, if you can figure out how to re-enable the push to feed tasks, that would be great. It stopped working for me a while ago, I didn't have the right permissions, so I disabled it.
> 
> Sure, I need to ask for permission to edit artifacts first though

I'll need the same permission, right? I can ask Chris Sfanos about that.

NiklasGustafsson:
Can we drop this one?

LittleLittleCloud:
@NiklasGustafsson We might still need to revert container back to ubuntu20? Let's still keep this PR open for a while until the ml.net repo side has verified that the libtorch 2.2.1 work with ubuntu 20 | centos 9 helix test machine

NiklasGustafsson:
> @NiklasGustafsson We might still need to revert container back to ubuntu20? Let's still keep this PR open for a while until the ml.net repo side has verified that the libtorch 2.2.1 work with ubuntu 20 | centos 9 helix test machine

@LittleLittleCloud -- still the case?

LittleLittleCloud:
@NiklasGustafsson 

Still the case. Just get some time to verify it on centos streaming 9.

Good news is `libtorch.so` can be loaded on centos 9, bad news is `libLibTorchsharp` requires a higher glibc version.

So we might still need to convert the building container back (to ubuntu 20.2 this time)

image

azure-pipelines.yml
NiklasGustafsson: OK

NiklasGustafsson:
Before publishing to NuGet, we need the built packages tested with ML.NET, so that we're not releasing something that doesn't actually solve the problem.

azure-pipelines.yml
LittleLittleCloud: No longer applied

azure-pipelines.yml
LittleLittleCloud: No longer applied

azure-pipelines.yml
NiklasGustafsson: Not sure I understand what you are saying. What is no longer applied?

LittleLittleCloud: I mean the change of installing conda environment in pipeline no longer applies to the latest commit

LittleLittleCloud:
> Before publishing to NuGet, we need the built packages tested with ML.NET, so that we're not releasing something that doesn't actually solve the problem.

I have verified the cpu package with mlnet. Where can I verify the cuda package

NiklasGustafsson:
> > Before publishing to NuGet, we need the built packages tested with ML.NET, so that we're not releasing something that doesn't actually solve the problem.
> 
> I have verified the cpu package with mlnet. Where can I verify the cuda package

You're the one who has a 4090, right? :-)

LittleLittleCloud:
> > > Before publishing to NuGet, we need the built packages tested with ML.NET, so that we're not releasing something that doesn't actually solve the problem.
> > 
> > 
> > I have verified the cpu package with mlnet. Where can I verify the cuda package
> 
> You're the one who has a 4090, right? :-)

Yes, I also have 1060 and 4050 as well, so a lot of scenarios with GPU I can test with

michaelgsharp:
Just curious, how much does the compiler affect the cuda builds?

NiklasGustafsson:
> Just curious, how much does the compiler affect the cuda builds?

It's not the compiler as much as `libc` that's of concern. Since the CUDA libtorch is a separate distribution, it should be tested, and the build pipeline does not include any GPU machines.

michaelgsharp:
@LittleLittleCloud I can show you how to enable the GPU for the big tests we have if you haven't done them before. We have at least 1 big test for each scenario in ML.NET that should fully exercise the gpu. Or at least for the stuff its touching anyways.

LittleLittleCloud:
@michaelgsharp That would be great, I will reach out to you privately

Directory.Build.props
```diff
@@ -20,7 +20,7 @@
     <SourceDir>$(RepoRoot)src/</SourceDir>
     <PkgDir>$(RepoRoot)pkg/</PkgDir>
 
-    <LibTorchPackageVersion>2.2.1.1</LibTorchPackageVersion>
+    <LibTorchPackageVersion>2.2.1.2</LibTorchPackageVersion>
```
NiklasGustafsson, line 23: There should be no reason to do this.

build/BranchInfo.props
```diff
@@ -2,8 +2,8 @@
   <PropertyGroup>
     <MajorVersion>0</MajorVersion>
     <MinorVersion>102</MinorVersion>
-    <PatchVersion>7</PatchVersion>
-    <PreviousPackageVersion>0.102.6</PreviousPackageVersion>
+    <PatchVersion>8</PatchVersion>
```
NiklasGustafsson, line 5: Actually, 0.102.7 hasn't been released, so there's no need to bump. I usually bump the version number right after a release of the previous version, helps me not forget to do it.

src/TorchSharp/Torch.cs
```diff
@@ -17,8 +17,8 @@ namespace TorchSharp
 {
     public static partial class torch
     {
-#if LIBTORCH_2_2_1_1
-        const string libtorchPackageVersion = "2.2.1.1";
+#if LIBTORCH_2_2_1_2
+        const string libtorchPackageVersion = "2.2.1.2";
```
NiklasGustafsson, line 21: See above.

azure-pipelines.yml
```diff
@@ -40,7 +40,7 @@ resources:
 #   https://github.com/dotnet/versions/blob/main/build-info/docker/image-info.dotnet-dotnet-buildtools-prereqs-docker-main.json
 
    - container: UbuntuContainer
-     image: mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-22.04-mlnet-20240216051055-578f026
+     image: mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-20.04-20240708213715-dcf0bb9
```
NiklasGustafsson, line 43: I think this is the only change that is actually needed.

NiklasGustafsson (changes requested):

PR #1339, opened 2024-06-25
Enable package validation

### LINKED ISSUE #681: Enable API Compat for TorchSharp
Can we enable API Compat for TorchSharp?   It’s part of the .NET SDK since 6.0.

Enabling it can be done with these steps:
[.NET Package Validation - .NET | Microsoft Docs](https://docs.microsoft.com/en-us/dotnet/fundamentals/package-validation/overview)

This will help flag when changes are made that would break the API surface.  You can still choose to make those changes if you want to suppress the errors, but this will make it clear when breaking changes are made.

ericstj:
Fixes #681 

I also fixed the native build which was broken for me on Windows with VS 17.11 / cmake 3.23.

### DIFF
--- a/pkg/Directory.Build.targets
+++ b/pkg/Directory.Build.targets
@@ -1,0 +1,8 @@
+<Project>
+  <Import Project="..\Directory.Build.targets" />
+
+  <PropertyGroup Condition="'$(IsPackable)' != 'false'">
+    <EnablePackageValidation>true</EnablePackageValidation>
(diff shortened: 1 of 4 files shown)


pkg/Directory.Build.targets
```diff
@@ -0,0 +1,8 @@
+<Project>
+  <Import Project="..\Directory.Build.targets" />
+
+  <PropertyGroup Condition="'$(IsPackable)' != 'false'">
```
ericstj, line 4: This is done in a targets file rather than props to honor projects setting for IsPackable.

Issue #1353, opened 2024-07-13
IterableDataset


yueyinqiu:
It seems that PyTorch has something called `IterableDataset`, which is different from `IterableDataset` in TorchSharp.

PR #1354, opened 2024-07-14
concat dataset

yueyinqiu:
https://github.com/dotnet/TorchSharp/discussions/1348#discussioncomment-10035168

- add `torch.utils.data.ConcatDataset`
    - The parameter of `DataLoader`s has been relaxed to `IDataset<out T>` (a new covariant interface), thus accepting the concated dataset
        - now `Dataset<T>` implements `IDataset<T>`
        - `Dataset` implements both `IDataset<Dictionary<string, Tensor>>` and `IDataset<IReadOnlyDictionary<string, Tensor>>`
        - `IterableDataset` implements `IDataset<IList<string, Tensor>>` and `IDataset<IEnumerable<string, Tensor>>`
- parameter of collate functions has been relaxed to `IReadOnlyList`

### DIFF
--- a/src/TorchSharp/Dataset.cs
+++ b/src/TorchSharp/Dataset.cs
@@ -14,15 +14,23 @@ namespace TorchSharp
                 /// <summary>
                 /// Map-style data set
                 /// </summary>
-                public abstract class Dataset : Dataset<Dictionary<string, torch.Tensor>>
+                public abstract class Dataset : Dataset<Dictionary<string, Tensor>>,
+                    IDataset<IReadOnlyDictionary<string, Tensor>>
                 {
+                    // Due to covariation, it should naturally be IDataset<IReadOnlyDictionary<string, Tensor>>.
+                    // However FSharp.Examples will break down without this.
+                    IReadOnlyDictionary<string, Tensor> IDataset<IReadOnlyDictionary<string, Tensor>>.this[long index] => this[index];
                 }
 
                 /// <summary>
                 /// Iterable-style data sets
                 /// </summary>
-                public abstract class IterableDataset : Dataset<IList<Tensor>>
+                public abstract class IterableDataset : Dataset<IList<Tensor>>,
+                    IDataset<IEnumerable<Tensor>>
                 {
+                    // Due to covariation, it should naturally be IDataset<IEnumerable<Tensor>>.
+                    // However FSharp.Examples will break down without this.
+                    IEnumerable<Tensor> IDataset<IEnumerable<Tensor>>.this[long index] => this[index];
                 }
 
                 /// <summary>
--- a/src/TorchSharp/Dataset.cs
+++ b/src/TorchSharp/Dataset.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
 using System;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
 
 namespace TorchSharp
 {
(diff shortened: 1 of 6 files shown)


yueyinqiu:
It's still work in progress. The pull request is created to run the unit tests.

src/TorchSharp/Dataset.cs
```diff
@@ -13,21 +14,29 @@ public static partial class data
                 /// <summary>
                 /// Map-style data set
                 /// </summary>
-                public abstract class Dataset : Dataset<Dictionary<string, torch.Tensor>>
+                public abstract class Dataset : Dataset<Dictionary<string, Tensor>>,
+                    IDataset<IReadOnlyDictionary<string, Tensor>>
```
yueyinqiu, line 18: As it commented here, due to covariation, `Dataset` should naturally be `IDataset<IReadOnlyDictionary<string, Tensor>>` (because it is `IDataset<Dictionary<string, Tensor>>`). However `FSharp.Examples` cannot be complied without this. I don't know why this happens.

I would suggest to remove this line, because it influences the detection of torch.utils.data.ConcatDataset (It doesn't know which type to use. `IReadOnlyDictionary` or `Dictionary` are both ok), but I failed to make the fsharp program work.

yueyinqiu: A more radical idea is to remove `Dataset` and `IterableDataset` at all. Actually we don't need them. `DataLoader`s could work on any `IDataset<IReadOnlyDictionary<string, torch.Tensor>>` and `IDataset<IEnumerable<Tensor>>`.

By the way, `IterableDataset` occupied the position of PyTorch `IterableDataset`. #1353 That's also part of the reason why I wants to remove them at all.

src/TorchSharp/Dataset.cs
```diff
@@ -49,8 +64,31 @@ public void Dispose()
...
+                    /// </summary>
+                    long Count { get; }
+
+                    /// <summary>
+                    /// Get tensor according to index
+                    /// </summary>
+                    /// <param name="index">Index for tensor</param>
+                    /// <returns>Tensors of index. DataLoader will catenate these tensors into batches.</returns>
+                    [IndexerName("DatasetItems")]
+                    T this[long index] { get; }
```
yueyinqiu, line 88: It's using indexers instead of `GetTensor` in the new interface. I'm not really sure about its `IndexerName`.

And... shall we remove `GetTensor` at the same time? Currently `Dataset<T>` is keeping a abstract `GetTensor` to ensure compatibility.

src/TorchSharp/Dataset.cs
NiklasGustafsson: F# samples have to work.

yueyinqiu: Hmm... I suppose so... However it doesn't https://dev.azure.com/dotnet/TorchSharp/_build/results?buildId=107828&view=results

NiklasGustafsson: So. just to check -- with the current commits, the F# examples build, correct? I don't see any errors in the latest build.

yueyinqiu: Yes I could be built by adding `: IDataset<IReadOnlyDictionary<string, Tensor>>`.

NiklasGustafsson:
Ready to merge?

yueyinqiu:
yes

NiklasGustafsson:
@LittleLittleCloud, @yueyinqiu -- After merge, builds are failing:

https://dev.azure.com/dotnet/TorchSharp/_build/results?buildId=107925&view=results

NiklasGustafsson:
This was reverted due to breaking changes preventing a build. We're eager to make a new release with support for Ubuntu 20.04.

PR #1363, opened 2024-08-01
 Increase libtorch version to 2.4.0 and switch to libtorch distribution for osx-arm64

ejhg:
- Compile osx-arm64 against libtorch-macos-arm64 instead of conda py3.10 libraries to improve performance (~6X gain on M2 Macbook Air)
- Increasing libtorch version because libtorch on osx-arm64 requires libomp.dylib, which is only bundled with 2.4.0
- Update SHA signature files and distribution filenames for Windows, Linux, and macOS
- osx-x64 can no longer be built because the last distribution available on https://download.pytorch.org/libtorch/cpu is libtorch-macos-x86_64-2.2.2.zip


ejhg:
> @dotnet-policy-service agree

@dotnet-policy-service agree

ejhg:
The main motivation for this PR is to significantly increase performance on Apple silicon.

I saw a ~6X improvement when running https://github.com/ejhg/llama-torchsharp using these 2.4.0 arm64 binaries on an M2 MacBook Air, 24GB.  Without this change, Llama on TorchSharp would run about 2 times slower than the reference implementation in python.  With this change, it runs 3 times faster than the python version (https://github.com/aggiee/llama-v2-mps)

One concern I have is that while the following build command from the DEVGUIDE.md succeeds,

```
dotnet build src/Redist/libtorch-cpu/libtorch-cpu.proj /p:UpdateSHA=true /p:TargetOS=mac /p:Configuration=Release /t:Build /p:IncludeLibTorchCpuPackages=true
```

specifying `/p:TargetArchitecture=x64` will cause it to fail because libtorch 2.4.0 is not available for x86_64 from https://download.pytorch.org/libtorch/cpu . (I don't know why.)

I have only tested on a Macbook Air.  No testing done on Windows or Linux.

ejhg:
@NiklasGustafsson I think you would be a great reviewer for this.  Thank you!

NiklasGustafsson:
Thank you for the PR! The performance gains are definitely compelling. Obsoleting OSX/x64 will take some thinking about -- it's our only chance to run unit tests on MacOS (we don't have any OSX/arm64 VMs to run tests on). 

Did you build the libtorch-* packages somewhere? The trick in creating a new libtorch release is the "bin packing" of the binaries, which always change in size into the NuGet packages. The compressed size of each package has to be < 250 MB for the NuGet gallery, which mostly affects the Windows and Linux CUDA backends.

ejhg:
@NiklasGustafsson 

(I'll debug the CI errors.)

For what it's worth, 2.4.0 for osx-64 is not available on anaconda either: https://conda.anaconda.org/pytorch/osx-64/

Regarding the packaging, I ran this command (modified slightly from the DEVGUIDE.md step 10) for each TargetOS.

```
dotnet pack -c Release -v:n /p:SkipTests=true /p:TargetOS=linux /p:IncludeTorchSharpPackage=true /p:IncludeLibTorchCpuPackages=true /p:IncludeLibTorchCudaPackages=true /p:ApiCompatGenerateSuppressionFile=true /p:UpdateSHA=true
```

I am getting some errors about netstandard2.0, but it's still producing the following files.  They all seem under 250MB.

```
-rw-r--r--  1 ejhg  staff      11987 Aug  1 13:32 Examples.Utils.0.102.8.nupkg
-rw-r--r--  1 ejhg  staff      25660 Aug  1 13:32 FSharp.Examples.0.102.8.nupkg
-rw-r--r--  1 ejhg  staff       4240 Aug  1 13:38 TorchAudio.0.102.8.nupkg
-rw-r--r--  1 ejhg  staff      10368 Aug  1 13:37 libtorch-cpu-linux-x64.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff      10366 Aug  1 13:37 libtorch-cpu-osx-arm64.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff      10350 Aug  1 13:37 libtorch-cpu-osx-x64.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff      10353 Aug  1 13:37 libtorch-cpu-win-x64.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff      10354 Aug  1 13:37 libtorch-cpu.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff  113927339 Aug  1 13:37 libtorch-cuda-12.1-linux-x64-part1.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff       9988 Aug  1 13:37 libtorch-cuda-12.1-linux-x64-part2-fragment1.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff      12249 Aug  1 13:37 libtorch-cuda-12.1-linux-x64-part2-primary.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff       9988 Aug  1 13:37 libtorch-cuda-12.1-linux-x64-part3-fragment1.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff       9988 Aug  1 13:37 libtorch-cuda-12.1-linux-x64-part3-fragment2.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff       9990 Aug  1 13:37 libtorch-cuda-12.1-linux-x64-part3-fragment3.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff      12250 Aug  1 13:37 libtorch-cuda-12.1-linux-x64-part3-primary.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff       9988 Aug  1 13:37 libtorch-cuda-12.1-linux-x64-part4-fragment1.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff      12254 Aug  1 13:37 libtorch-cuda-12.1-linux-x64-part4-primary.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff       9989 Aug  1 13:37 libtorch-cuda-12.1-linux-x64-part5-fragment1.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff      12250 Aug  1 13:37 libtorch-cuda-12.1-linux-x64-part5-primary.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff  168415736 Aug  1 13:38 libtorch-cuda-12.1-linux-x64-part6.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff  150928982 Aug  1 13:38 libtorch-cuda-12.1-linux-x64-part7.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff  230313812 Aug  1 13:38 libtorch-cuda-12.1-linux-x64-part8.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff      10530 Aug  1 13:38 libtorch-cuda-12.1-linux-x64.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff       9894 Aug  1 13:42 libtorch-cuda-12.1-win-x64-part1.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff       9902 Aug  1 13:42 libtorch-cuda-12.1-win-x64-part10.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff       9892 Aug  1 13:42 libtorch-cuda-12.1-win-x64-part2.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff       9971 Aug  1 13:42 libtorch-cuda-12.1-win-x64-part3-fragment1.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff      12235 Aug  1 13:42 libtorch-cuda-12.1-win-x64-part3-primary.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff       9895 Aug  1 13:42 libtorch-cuda-12.1-win-x64-part4.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff       9973 Aug  1 13:42 libtorch-cuda-12.1-win-x64-part5-fragment1.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff      12234 Aug  1 13:42 libtorch-cuda-12.1-win-x64-part5-primary.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff       9894 Aug  1 13:42 libtorch-cuda-12.1-win-x64-part6.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff       9895 Aug  1 13:42 libtorch-cuda-12.1-win-x64-part7.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff       9893 Aug  1 13:42 libtorch-cuda-12.1-win-x64-part8.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff       9973 Aug  1 13:42 libtorch-cuda-12.1-win-x64-part9-fragment1.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff       9974 Aug  1 13:42 libtorch-cuda-12.1-win-x64-part9-fragment2.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff      12235 Aug  1 13:42 libtorch-cuda-12.1-win-x64-part9-primary.2.4.0.nupkg
-rw-r--r--  1 ejhg  staff      10518 Aug  1 13:42 libtorch-cuda-12.1-win-x64.2.4.0.nupkg
```

NiklasGustafsson:
I'm trying to gauge internal buy-in for jettisoning OSX/x64. If PyTorch doesn't support it, it's reasonable to say TorchSharp shouldn't, either, but I need to ping some folks.

ejhg:
From the PyTorch 2.2 [Release notes](https://github.com/pytorch/pytorch/releases) 

> Please note that we are deprecating macOS x86 support, and PyTorch 2.2.x will be the last version that supports macOS x64.

see also https://github.com/pytorch/pytorch/issues/114602

ejhg:
We obviously cannot stall the train for all other platforms.

I think it's worth keeping osx-x64 support by pinning it to the latest available 2.2.X, perhaps by introducing new build properties?

Ideally, (in my opinion) the native layer should be decoupled from the managed layer on top.  In essence, the LibTouchSharp binary should be shipped with the native nuggets, separate from the managed code.  Currently, it is bundled with the "cpu" nuggets, and that creates a coupling between the managed and unmanaged code.

NiklasGustafsson:
> We obviously cannot stall the train for all other platforms.
> 
> I think it's worth keeping osx-x64 support by pinning it to the latest available 2.2.X, perhaps by introducing new build properties?
> 
> Ideally, (in my opinion) the native layer should be decoupled from the managed layer on top. In essence, the LibTouchSharp binary should be shipped with the native nuggets, separate from the managed code. Currently, it is bundled with the "cpu" nuggets, and that creates a coupling between the managed and unmanaged code.

Each libtorch version usually adds APIs that you want to be able to take advantage of, so there's a coupling no matter what. There are also constant changes to the header files, etc., which affects the C++ ABI and makes it hard to decouple.

That said, it would be useful to find a way to build the MacOS / x64 version of TorchSharp on an older version of libtorch, and move forward with newer versions on others. It would require some conditionals in the build scripts / XML files (and Torch.cs), I think./

NiklasGustafsson:
> We obviously cannot stall the train for all other platforms.

While I agree that is a reasonable opinion, and probably the winning one, it's not clear to me that it is "obvious" :-) Without usage data, which I'm trying to get, it's a tough decision to make. My understanding is that Apple will continue supporting Intel-based HW until MacOS 17.

ejhg:
@NiklasGustafsson I think we're on the same page.

I just realized that the libtorch-* package versions correspond 1-to-1 with the torch backend version, and that this is indeed separate from the TorchSharp version.

(I still think that LibTorchSharp belongs in the libtorch-* packages because it is more strongly coupled with the binaries it is compiled against than with the code that dll imports from it, but this is not important right now.)

The outcome of this PR will be version 2.2.2.0 for libtorch-cpu-osx-arm64 and version 2.4.0.0 for the rest.  (Still waiting for the CI.)

I think that makes sense.  What do you think?

NiklasGustafsson:
> The outcome of this PR will be version 2.2.2.0 for libtorch-cpu-osx-arm64 and version 2.4.0.0 for the rest. (Still waiting for the CI.)
> 
> I think that makes sense. What do you think?

The package dependencies will get complicated -- specifically TorchSharp-cpu, which is empty of binaries and only has dependencies. We'll have to refer users to the 0.102.7 version and should cut a branch for that, probably do another release 0.102.8 with recent bug fixes, too, before merging this PR.

Then, we disable the OSX_x64 builds and drop support.

NiklasGustafsson:
This will take a little bit of finessing...

NiklasGustafsson:
The version number in this PR should be set to 0.103.0 -- we always increment the minor version when upgrading to a new version of libtorch.

ejhg:
@NiklasGustafsson By the way, I have absolutely no rush to get this merged, so don't feel pressured in any way.  I'm just trying my best to wrap this up on my end, and happy to hand it off or wait for others to chime in, or wait for TorchSharp to gradually increase versions through other PRs...

I was just really excited last night when I finally sorted out the binaries for arm64, and I was excited to share it with others :)

Now, I'll be happy when I see that CI turn green :)

NiklasGustafsson:
Yes, I'll probably pick up your work and go through all the nasty details of getting the packages built right. It can only happen on Azure, not locally, and I can start a pipeline from my private repo, which is how I do the validation.

ejhg:
Thank you @NiklasGustafsson !  I won't push anything else.  Let me know if I can help.

Again, no rush at all on my side.  I don't mean to interrupt your day!

NiklasGustafsson:
> Thank you @NiklasGustafsson ! I won't push anything else. Let me know if I can help.
> 
> Again, no rush at all on my side. I don't mean to interrupt your day!

Just pulled your code into: NiklasGustafsson/TorchSharp/2.4.0

NiklasGustafsson:
But every single test fails trying to load the libtorch binaries.

ejhg:
@NiklasGustafsson on what platform?

ejhg:
@NiklasGustafsson You probably already know this, but something I found very useful to diagnose is replacing `NativeLibrary.TryLoad` with `NativeLibrary.Load` in `Torch.cs`.  The exception that `NativeLibrary.Load` throws contains very useful information to trace the root cause.

ejhg:
ok, reproduced test failures...

ejhg:
`dotnet test` should now work fine with the latest commit [05bed90](https://github.com/dotnet/TorchSharp/pull/1363/commits/05bed90462760224050d245d53c74be1e29314d5)

The CI seems to be green now too.

The only thing pending is that the CI is not running the Build_*_Packages steps even though I set `BuildLibTorchPackages` to true.

NiklasGustafsson:
> The only thing pending is that the CI is not running the Build_*_Packages steps even though I set `BuildLibTorchPackages` to true.

That's because the condition is that the flag is set and you're building on the 'main' branch -- the CUDA packages take a significant amount of space, so we don't want to build them unless there's a release coming. That's one reason why I build in my private repo when I'm trying out the package sizes.

Anyway, thanks for all the help!! The tests are working and going to do the bin packing that needs to be done for the CUDA packages.

ejhg:
> the condition is that the flag is set and you're building on the 'main' branch -- the CUDA packages take a significant amount of space, so we don't want to build them unless there's a release coming.

A simple workaround worth considering could be to enable the generation and verification of artifacts without actually archiving them.  (`rm -rf` at the end of the build, if needed.)  It would save a lot of time and effort.

On the contrary, Thank You for taking over and for actively maintaining a viable alternative to Python across all platforms :)  Have a great weekend!

NiklasGustafsson:
This PR was incorporated into #1365, which was just released as v0.103.0 on NuGet Gallery.

ejhg:
Thank you @NiklasGustafsson 

I did a quick test run using the latest Nuget packages, and my results were mixed (in terms of performance improvement).  I'll investigate and circle back.  I'm guessing there may be an issue in the build pipeline for MPS, but I'll have to rebuild locally to double check.

✅ CPU model on Apple silicon: performance improved drastically (~1800 ms to ~200 ms in my benchmark)
❓ MPS model on Apple silicon performance does not seem to be improved (~1200 ms in my benchmark).  I expected it to be down to ~200ms as well.

ejhg:
It seems I can't reproduce the "fast" speeds on MPS even with my local build.  :(  It's possible that the only improvement I saw was on CPU, and I mistakenly believed it was MPS--in the rush to get this out.

The improvement on CPU is still very welcome, and I know that CPU may sometimes be faster, but such a big disparity is unexpected in my opinion.  This may be an issue with my local machine or with my benchmark.  I'll continue digging into it.

PR #1389, opened 2024-10-21
Resolves #1387 and backfill tests for ThreadDisposeScopeStatistics

### LINKED ISSUE #1387: Attaching unowned tensor to a DisposeScope makes Statistics.DetachedFromScopeCount go negative.
Repro:
```
[Fact]
    public void DetachedFromScopeCountNegative()
    {
        var stats = DisposeScopeManager.Statistics;
        stats.Reset();
        var t = torch.tensor(1);
        Assert.Equal(0, stats.DetachedFromScopeCount);
        var scope = torch.NewDisposeScope();
        scope.Attach(t);
        Assert.Equal(0, stats.DetachedFromScopeCount);
    }
```

Removing the branch in `DisposeScope.Attach(IEnumerable<IDisposable> disposables)` that subtracts fixes the issue and all tests pass, minus any tests checking counts (waiting on PR #1386 prior to fixing.

mvphelps:
Added test to verify that attaching an unscoped Tensor or PackedSequence no longer decrements DetachedFromScopeCount, making it possibly become negative. Note also removed the scope counting tests from TestDisposeScopesPackedSequence, which now only tests disposal behavior. Metrics are now all tested in a dedicated set of test fixtures.

Added additional tests to document behavior and improve test coverage for both Tensor and PackedSequence objects, in scenarios for both scoped and unscoped instances of these objects (total of 4 new fixtures). This should help future changes as one can see what the system does to count the statistics.

Tests reflect the expected behavior based on the doc comments for ThreadDisposeScopeStatistics. Note that the ThreadTotalLiveCount property is still "approximate" and can go negative, not changes to this behavior were made. 

Let me know if I've added too much here. I really wasn't sure what was supposed to be counted when, so I just went though it incrementally and added the tests to document it.

### DIFF
--- a/src/TorchSharp/DisposeScope.cs
+++ b/src/TorchSharp/DisposeScope.cs
@@ -237,17 +237,6 @@ namespace TorchSharp
 
             var result = new List<IDisposable>();
             foreach (var disposable in disposables) {
-                if (disposable is torch.Tensor tensor) {
-                    if (tensor.OwningDisposeScope == null && !tensor.IsInvalid) {
-                        _disposeScopeManager.StatisticsInstance.DetachedFromScopeCount--;
-                    }
-                }
-                else if (disposable is torch.nn.utils.rnn.PackedSequence sequence) {
-                    if (sequence.OwningDisposeScope == null && !sequence.IsInvalid) {
-                        _disposeScopeManager.StatisticsInstance.DetachedFromScopeCount--;
-                    }
-                }
-
                 AddToOther(this, disposable);
                 result.Add(disposable);
             }
--- a/test/TorchSharpTest/TestDisposeScopesPackedSequence.cs
+++ b/test/TorchSharpTest/TestDisposeScopesPackedSequence.cs
@@ -1,121 +1,82 @@
 using System.Reflection;
-using TorchSharp;
 using Xunit;
 
-namespace TorchSharpTest;
-
-public class TestDisposeScopesPackedSequence
+namespace TorchSharp
 {
-    [Fact]
-    public void MoveDisposeScope()
+    [Collection("Sequential")]
+    public class TestDisposeScopesPackedSequence
     {
-        var sequences = CreateTestSequences();
-        torch.nn.utils.rnn.PackedSequence packed_sequence;
-        var otherScope = torch.NewDisposeScope();
-        using (torch.NewDisposeScope())
+        [Fact]
+        public void MoveDisposeScope()
         {
-            using (torch.NewDisposeScope())
-            {
-                packed_sequence = torch.nn.utils.rnn.pack_sequence(sequences, enforce_sorted: false);
+            var sequences = CreateTestSequences();
+            torch.nn.utils.rnn.PackedSequence packed_sequence;
+            var otherScope = torch.NewDisposeScope();
+            using (torch.NewDisposeScope()) {
+                using (torch.NewDisposeScope()) {
+                    packed_sequence = torch.nn.utils.rnn.pack_sequence(sequences, enforce_sorted: false);
+                    AssertPackedSequenceValid(packed_sequence);
+
+                    packed_sequence.MoveToOuterDisposeScope();
+                }
+
                 AssertPackedSequenceValid(packed_sequence);
 
-                packed_sequence.MoveToOuterDisposeScope();
+                packed_sequence.MoveToOtherDisposeScope(otherScope);
             }
-            AssertPackedSequenceValid(packed_sequence);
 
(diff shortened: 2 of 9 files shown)


src/TorchSharp/DisposeScope.cs
```diff
@@ -237,17 +237,6 @@ public IReadOnlyList<IDisposable> Attach(IEnumerable<IDisposable> disposables)
 
             var result = new List<IDisposable>();
             foreach (var disposable in disposables) {
-                if (disposable is torch.Tensor tensor) {
-                    if (tensor.OwningDisposeScope == null && !tensor.IsInvalid) {
```
NiklasGustafsson, line 241: Here, it may be worth adding a separate interface for just the 'IsInvalid' rather than expanding on the IDisposable interface, for separation of concerns. Allow Tensor and PackedSequence to implement IValidatable.

test/TorchSharpTest/TestDisposeScopesPackedSequence.cs
```diff
@@ -1,121 +1,82 @@
...
-        torch.nn.utils.rnn.PackedSequence packed_sequence;
-        var otherScope = torch.NewDisposeScope();
-        using (torch.NewDisposeScope())
+        [Fact]
+        public void MoveDisposeScope()
         {
-            using (torch.NewDisposeScope())
-            {
-                packed_sequence = torch.nn.utils.rnn.pack_sequence(sequences, enforce_sorted: false);
+            var sequences = CreateTestSequences();
```
NiklasGustafsson, line 12: Indentation issue? Are you using TAB?

NiklasGustafsson (changes requested):

src/TorchSharp/DisposeScope.cs
mvphelps: Not sure I follow, IsInvalid is an existing public method on Tensor and this is casting specifically to use it. And we left it internal on PackedSequence to leave it hidden. Are you saying we want to make both of them public now, under that interface? I can do that just want to make sure I'm certain what you are looking for here as last time we talked about an interface, we agreed against it to keep the API smaller.

test/TorchSharpTest/TestDisposeScopesPackedSequence.cs
mvphelps: This one is weird, I am using spaces, and it shows fine in my IDE directly on that commit. I even used the arrow keys to count my spaces, there are 20 just like the one line above it. I want to fix this but have never seen that before.  I've even switched to main and back to this branch with my IDE closed, then opened with a text editor and still get 20 spaces and proper alignment. Any ideas what I need to check?

mvphelps: It also looks fine in raw view: https://raw.githubusercontent.com/dotnet/TorchSharp/27bd06b368d6f1c1e6f60bd95b87f0cc753fbe82/test/TorchSharpTest/TestDisposeScopesPackedSequence.cs

mvphelps: @NiklasGustafsson I think I see the issue. There isn't one. The lines you are seeing above that are _replaced_ by the less indented line. It is properly spaced. It just looks odd because of the diff display. I'm going to close this one.

NiklasGustafsson:
Since this resolves an issue, that fact must go in the RELEASENOTES.md file.

mvphelps:
Release notes updated!

NiklasGustafsson:
@mvphelps, I think there's a problem handling the statistics for this situation (in Parameter.cs):

```C#
        public class Parameter : Tensor
        {
            /// <summary>
            /// Constructor
            /// </summary>
            /// <param name="data">A tensor, which will become empty.</param>
            /// <param name="requires_grad"></param>
            public Parameter(Tensor data, bool requires_grad = true) :
                base(data.with_requires_grad(requires_grad).MoveHandle())
            {
                var scope = data.OwningDisposeScope;
                if (scope is not null) {
                    this.OwningDisposeScope = scope;
                    scope.Attach(this);
                    scope.Detach(data);
                }
            }

            /// <summary>
            /// Constructor
            /// </summary>
            /// <param name="handle">A tensor handle.</param>
            internal Parameter(System.IntPtr handle) : base(handle)
            {
            }
        };
```

I don't have a small repo, but I don't think the stats are collected properly when a parameter takes over the native handle from a non-Parameter tensor and registers itself with its dispose scope.

There are two tests that fail when I implement Linear in .NET rather than native code.

PR #1391, opened 2024-10-22
Better metrics

### LINKED ISSUE #1390: Provide better metrics to enable memory leak troubleshooting
DisposeScopeManager.Statistics does not capture count of objects disposed that are not on a DisposeScope, or number of items attached to a DisposeScope. Also the ThreadTotalLiveCount is "approximate" and often goes negative. Ideally it would be exact.

mvphelps:
Improve visibility of disposables (Tensor and PackedSequence) to make memory leaks easier to troubleshoot. No breaking interface changes. This resolves #1390.

This PR is what I'd actually like to see for object lifetime statitics. It gives complete statistics for object creation and disposables whether in a DisposeScope or not, and accurate Attach/Detach counts. Lastly the ThreadTotalLiveCount is no longer estimated, it is exact. Also added documentation on how to use the statistics for troubleshooting memory leaks.

New Documentation:
memory leak troubleshooting.md - provides practical guidance on how to use the changes of this PR for useful troubleshooting of memory issues.memory.md was also updated to link to this document.

New Interface Members
* DisposedOutsideScopeCount - # of objects .Dispose when not attached to a scope
* AttachedToScopeCount - # of objects attached to a scope via .Attach
* ThreadTotalLiveCount (No longer approximate, is exact)
* ToString - creates a useful debug string for the developer.

Semantic changes: 
None of the statistics are decremented anymore. All statistics are additive only. If an event happens, it is incremented. The original DisposeScopeManager.Statistics object now represents the total combined counts for both PackedSequence and Tensor. It also has individual attributes for .TensorStatistics and .PackedSequenceStatistics where these counts are tracked individually. A common interface ILifetimeStatistics was introduced for these. Additionally the Dispose for both Tensor and PackedSequence were both modified to prevent reentry and double incrementing disposal metrics.

Defects discovered during implementing this:
1. TestDisposeScopes.MinimalDisposeWorks had commented assertions, and actually reveals the prior approximate ThreadTotalLiveCount was not capturing live tensors. It now captures them accurately. This also led to discovery of the next item (the reason for the commented assertions).
2. The .ToTensor method has always leaked a tensor that is never disposed. There are two new tests to capture this behavior as it is currently, both in and out of a DisposeScope, with comments on how they need to change when this defect is fixed. These tests are in different fixtures but both named "ToTensorCreatesOrphanedTensor". I can create a new issue for this if you decide this PR is worth moving forward, or I can try to fix it here in this PR.

I understand this is a larger PR and some siginificant API changes. Given that it provided accurate counts and helped identify defects withing TorchSharp itself, the useful utility provided by these changes is demonstrated. Let me know what you like, hate, or want changed. 

Note dependent on PR #1389 (although maybe not, the commit appears in this PR too so maybe we can close that other one).

### DIFF
--- a/src/TorchSharp/DisposeScope.cs
+++ b/src/TorchSharp/DisposeScope.cs
@@ -154,9 +154,13 @@ namespace TorchSharp
         {
             if (this._disposeScopeManager is null)
                 throw new ObjectDisposedException(this.GetType().FullName);
-            foreach (var disposable in disposables) {
-                if (Disposables.Remove(disposable)) {
-                    AddToOther(scope, disposable);
+            if (scope == null) {
+                Detach(disposables);
+            } else {
+                foreach (var disposable in disposables) {
+                    if (Disposables.Remove(disposable)) {
+                        AddToOther(scope, disposable);
+                    }
                 }
             }
         }
@@ -209,11 +213,11 @@ namespace TorchSharp
                 throw new ObjectDisposedException(this.GetType().FullName);
             foreach (var disposable in disposables) {
                 if (Disposables.Remove(disposable)) {
-                    _disposeScopeManager.StatisticsInstance.DetachedFromScopeCount++;
                     if (disposable is torch.Tensor tensor) {
+                        _disposeScopeManager.StatisticsInstance._TensorStatistics.DetachedFromScopeCount++;
                         tensor.OwningDisposeScope = null;
-                    }
-                    else if (disposable is torch.nn.utils.rnn.PackedSequence sequence) {
+                    } else if (disposable is torch.nn.utils.rnn.PackedSequence sequence) {
+                        _disposeScopeManager.StatisticsInstance._PackedSequenceStatistics.DetachedFromScopeCount++;
                         sequence.OwningDisposeScope = null;
                     }
                 }
@@ -237,7 +241,13 @@ namespace TorchSharp
 
             var result = new List<IDisposable>();
             foreach (var disposable in disposables) {
-                AddToOther(this, disposable);
+                if (AddToOther(this, disposable)) {
+                    if (disposable is torch.Tensor tensor) {
+                        _disposeScopeManager.StatisticsInstance._TensorStatistics.AttachedToScopeCount++;
+                    } else if (disposable is torch.nn.utils.rnn.PackedSequence sequence) {
+                        _disposeScopeManager.StatisticsInstance._PackedSequenceStatistics.AttachedToScopeCount++;
+                    }
+                }
                 result.Add(disposable);
             }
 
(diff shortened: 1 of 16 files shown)


src/TorchSharp/DisposeScope.cs
```diff
@@ -154,9 +154,13 @@ public void MoveToOther(DisposeScope? scope, IEnumerable<IDisposable> disposable
         {
             if (this._disposeScopeManager is null)
                 throw new ObjectDisposedException(this.GetType().FullName);
-            foreach (var disposable in disposables) {
-                if (Disposables.Remove(disposable)) {
-                    AddToOther(scope, disposable);
+            if (scope == null) {
+                Detach(disposables);
+            } else {
```
NiklasGustafsson, line 159: Style: in this code base, I believe we usually put the 'else' on the next line.

mvphelps: I searched the code base for "} else {" and found 100+ matches. Searching for "else {" found 2 additional matches. The style for this code base is } else {, i.e. keeping the else on the same line as the preceeding closing brace. The two non-conformants  are Tensor.cs:132 and Examples/ResNet.cs:140.

Would you like me to update those two non-conforming files to match style? Then there would be no confusion for future devs. I know it is specifically requested to not supply stylistic only commits, but I can see the value here if you wish to waive that requirement this time.

mvphelps: Also the editorconfig says no newline:
https://github.com/dotnet/TorchSharp/blob/b30a60d361470918c4db6dbbd054606ae269182a/.editorconfig#L72

NiklasGustafsson: I stand corrected.

src/TorchSharp/ThreadDisposeScopeStatistics.cs
```diff
@@ -1,48 +1,141 @@
...
+        /// <summary>
+        /// Keeps track of the Tensor statistics for the current thread. Can be queried to figure out performance/memory issues.
+        /// </summary>
+        public ILifetimeStatistics TensorStatistics => _TensorStatistics;
+        internal LifetimeStatistics _TensorStatistics { get; set; } = new LifetimeStatistics();
+        /// <summary>
+        /// Keeps track of the PackedSequence statistics for the current thread. Can be queried to figure out performance/memory issues.
+        /// </summary>
+        public ILifetimeStatistics PackedSequenceStatistics => _PackedSequenceStatistics;
+        internal LifetimeStatistics _PackedSequenceStatistics { get; set; } = new LifetimeStatistics();
```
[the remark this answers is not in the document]
NiklasGustafsson: Get rid of the interface, is my preference.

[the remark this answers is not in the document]
mvphelps: Done. Unforeseen duplication was created. Maybe we do a middle ground approach where the interface is internal, and the concrete types are still is what is exposed? That would reduce the duplication this created (the .ToString, and the assertion method in TestDisposeScopesStatisticsBase). I also don't see these as excessive duplication either.

NiklasGustafsson:
Ready to merge?

PR #1393, opened 2024-10-23
Fix memory leaks in .ToTensor,  resolves #1392

### LINKED ISSUE #1392: ToTensor extension method creates an orphaned tensor (Memory Leak)
When invoking .ToTensor(), the method creates two tensors. The first one is sometimes overwritten and the reference is lost. Since it won't get Dispose(), this is a memory leak.

Reproduced in TestDisposeScopesStatisticsTensorUnscoped
```
[Fact]
public void ToTensorCreatesOrphanedTensor()
{
    //Defect: This needs fixing but is unrelated to the commit that discovered
    //it - adding better lifetime statistics. ToTensor() leaks 1 tensor
    // //every time it is called.
    var stats = DisposeScopeManager.Statistics;
    stats.Reset();
    var a1 = 1.ToTensor();
    //Should be 1 ideally. If it remains two...
    Assert.Equal(2, stats.CreatedOutsideScopeCount);
    // ... this should be 1
    Assert.Equal(0, stats.DisposedOutsideScopeCount);
    a1.Dispose();
    //... and this should also be 2
    Assert.Equal(1, stats.DisposedOutsideScopeCount);
}
```

See also `TestDisposeScopesStatisticsTensorScoped.ToTensorCreatesOrphanedTensorButDisposeScopeCleansItUp`

mvphelps:
This resolved the orphan tensor issue. 

Please note for tensor creation, I removed branches that were checking if device was supplied - this branch would always take precedence over the else branch that was checking dtype, because of the `InitializeDevice` call at the beginning. Not sure why that branch was there, since the tensor is being created on the correct device initially.

Since there was much duplication related to managing disposing the initial tensor when a dtype conversion was necessary, I've created new internal factory method `Tensor.InstantiateTensorWithLeakSafeTypeChange` to manage this. I can't say I love the name, but it's accurate, suggestions welcome. Also if there is a better place to put that method am happy to move it.

In the case of complex numbers, there were overloads that excepted a tuple (real, imaginary) and another that accepted them separately. I updated on to delegate to the other to reduce duplication since the implementations were identical. Release notes updated too.

Lastly I removed the two tests I added on the prior commit that capture the leak behavior, since there are now new and better tests that cover a broader range of scenarios.

### DIFF
--- a/src/TorchSharp/Tensor/Factories/tensor_double.cs
+++ b/src/TorchSharp/Tensor/Factories/tensor_double.cs
@@ -26,10 +26,7 @@ namespace TorchSharp
         /// </summary>
         public static Tensor tensor((double Real, double Imaginary) scalar, ScalarType? dtype = null, Device? device = null, bool requires_grad = false)
         {
-            device = InitializeDevice(device);
-            var handle = THSTensor_newComplexFloat64Scalar(scalar.Real, scalar.Imaginary, (int)device.type, device.index, requires_grad);
-            if (handle == IntPtr.Zero) { CheckForErrors(); }
-            return InstantiateTensorWithLeakSafeTypeChange(handle, dtype);
+            return tensor(scalar.Real, scalar.Imaginary, dtype, device, requires_grad);
         }
 
         /// <summary>
@@ -37,7 +34,11 @@ namespace TorchSharp
         /// </summary>
         public static Tensor tensor(double real, double imaginary, ScalarType? dtype = null, Device? device = null, bool requires_grad = false)
         {
-            return tensor((real, imaginary), dtype, device, requires_grad);
+
+            device = InitializeDevice(device);
+            var handle = THSTensor_newComplexFloat64Scalar(real, imaginary, (int)device.type, device.index, requires_grad);
+            if (handle == IntPtr.Zero) { CheckForErrors(); }
+            return InstantiateTensorWithLeakSafeTypeChange(handle, dtype);
         }
 
         /// <summary>
--- a/src/TorchSharp/Tensor/Factories/tensor_double.cs
+++ b/src/TorchSharp/Tensor/Factories/tensor_double.cs
@@ -18,43 +18,26 @@ namespace TorchSharp
             device = InitializeDevice(device);
             var handle = THSTensor_newFloat64Scalar(scalar, (int)device.type, device.index, requires_grad);
             if (handle == IntPtr.Zero) { CheckForErrors(); }
-            var tensor = new Tensor(handle);
-            tensor = dtype.HasValue ? tensor.to(dtype.Value, device) : tensor.to(device);
-            return tensor;
+            return InstantiateTensorWithLeakSafeTypeChange(handle, dtype);
         }
 
         /// <summary>
-        /// Create a scalar tensor from a single value
+        /// Create a scalar complex number tensor from a tuple of (real, imaginary)
         /// </summary>
         public static Tensor tensor((double Real, double Imaginary) scalar, ScalarType? dtype = null, Device? device = null, bool requires_grad = false)
         {
             device = InitializeDevice(device);
             var handle = THSTensor_newComplexFloat64Scalar(scalar.Real, scalar.Imaginary, (int)device.type, device.index, requires_grad);
             if (handle == IntPtr.Zero) { CheckForErrors(); }
-            var tensor = new Tensor(handle);
-            if (device is { }) {
-                tensor = dtype.HasValue ? tensor.to(dtype.Value, device) : tensor.to(device);
-            } else if (dtype.HasValue) {
-                tensor = tensor.to_type(dtype.Value);
-            }
-            return tensor;
+            return InstantiateTensorWithLeakSafeTypeChange(handle, dtype);
         }
 
         /// <summary>
-        /// Create a scalar tensor from a single value
+        /// Create a scalar complex number tensor from independent real and imaginary components
         /// </summary>
         public static Tensor tensor(double real, double imaginary, ScalarType? dtype = null, Device? device = null, bool requires_grad = false)
         {
-            device = InitializeDevice(device);
-            var handle = THSTensor_newComplexFloat64Scalar(real, imaginary, (int)device.type, device.index, requires_grad);
-            if (handle == IntPtr.Zero) { CheckForErrors(); }
-            var tensor = new Tensor(handle);
-            if (device is { }) {
-                tensor = dtype.HasValue ? tensor.to(dtype.Value, device) : tensor.to(device);
-            } else if (dtype.HasValue) {
-                tensor = tensor.to_type(dtype.Value);
-            }
-            return tensor;
+            return tensor((real, imaginary), dtype, device, requires_grad);
         }
 
         /// <summary>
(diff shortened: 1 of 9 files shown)


src/TorchSharp/Tensor/Factories/tensor_double.cs
```diff
@@ -18,43 +18,26 @@ public static Tensor tensor(double scalar, ScalarType? dtype = null, Device? dev
...
-            var handle = THSTensor_newComplexFloat64Scalar(real, imaginary, (int)device.type, device.index, requires_grad);
-            if (handle == IntPtr.Zero) { CheckForErrors(); }
-            var tensor = new Tensor(handle);
-            if (device is { }) {
-                tensor = dtype.HasValue ? tensor.to(dtype.Value, device) : tensor.to(device);
-            } else if (dtype.HasValue) {
-                tensor = tensor.to_type(dtype.Value);
-            }
-            return tensor;
+            return tensor((real, imaginary), dtype, device, requires_grad);
```
NiklasGustafsson, line 40: I'm wondering if this creates a heap-allocated tuple instance, which is not ideal. If we want only one method, it should probably to in the other direction, i.e. that this is the one that has the implementation, and the other one calls this method.

src/TorchSharp/Tensor/Factories/tensor_float.cs
```diff
@@ -22,37 +22,22 @@ public static Tensor tensor(float scalar, Device? device = null, bool requires_g
...
             var handle = THSTensor_newComplexFloat32Scalar(scalar.Real, scalar.Imaginary, (int)device.type, device.index, requires_grad);
             if (handle == IntPtr.Zero) { CheckForErrors(); }
-            var tensor = new Tensor(handle);
-            if (device is { }) {
-                tensor = dtype.HasValue ? tensor.to(dtype.Value, device) : tensor.to(device);
-            } else if (dtype.HasValue) {
-                tensor = tensor.to_type(dtype.Value);
-            }
-            return tensor;
+            return InstantiateTensorWithLeakSafeTypeChange(handle, dtype);
```
NiklasGustafsson, line 40: See comments for double-precision complex numbers.

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -7402,5 +7402,15 @@ public static Tensor WrappedTensorDisposeScope(Func<Tensor> expr)
             var result = expr();
             return result.MoveToOuterDisposeScope();
         }
+        internal static Tensor InstantiateTensorWithLeakSafeTypeChange(IntPtr handle, ScalarType? dtype)
+        {
+            var tensor = new Tensor(handle);
+            if (dtype.HasValue) {
```
NiklasGustafsson, line 7408: Now that I see this code isolated, I'm wondering if it wouldn't be faster to check here whether the type is already the right one and skip the call to `to_type()`

src/TorchSharp/Tensor/Tensor.cs
```diff
@@ -7402,5 +7402,15 @@ public static Tensor WrappedTensorDisposeScope(Func<Tensor> expr)
...
+                var typed = tensor.to_type(dtype.Value);
+                tensor.Dispose();
```
NiklasGustafsson, line 7410: Can you explain why the tensor is orphaned? Why does it not get added to the DisposeScope?

test/TorchSharpTest/TestDisposeScopesStatisticsTensor.cs
```diff
@@ -73,26 +73,5 @@ public void DisposingScopeAfterDetachingDoesNothing()
             AssertTensorCounts(0, 0, 1, 0, 0, 1, 1);
             AssertTotalsCounts(0, 0, 1, 0, 0, 1, 1);
         }
-
```
NiklasGustafsson, line 76: Why remove the tests?

NiklasGustafsson (changes requested):

src/TorchSharp/Tensor/Tensor.cs
mvphelps: What was happening prior is that the `var tensor` is added to the scope if one was present. Then when either the device or dtype branch was hit, a new tensor would be created. The old tensor was never disposed and it's reference never returned, although it is no longer needed. It would remain on the DisposeScope until the scope got disposed. 

If you had no dispose scope, the tensor could last until the application terminates.

test/TorchSharpTest/TestDisposeScopesStatisticsTensor.cs
mvphelps: Because the other tests that were added in this PR are more comprehensive, and test in the same way. I created those two as a placeholder to remind me what was wrong and how to recreate it.

src/TorchSharp/Tensor/Factories/tensor_double.cs
mvphelps: Great point, fixed. Same for float, I'll resolve this one and leave this one for you to review.

src/TorchSharp/Tensor/Factories/tensor_float.cs
mvphelps: Fixed.

src/TorchSharp/Tensor/Tensor.cs
mvphelps: Great point I should have seen that myself, one less Tensor allocation is better. Fixed.

src/TorchSharp/Tensor/Tensor.cs
NiklasGustafsson: I see. So, it wasn't really orphaned, then, its lifespan wasn't optimal. In that case, doesn't it have to be detached from any DisposeScope that might have picked it up?

mvphelps: Your description is better than mine. Inside the Dispose of Tensor the DisposeScopeManager is notified to update statistics, and the call to OwningDisposeScope.MarkAsDisposed method takes care of removing the object from the DisposeScope that owns it. In the event there was never a DisposeScope, it is still properly disposed.

NiklasGustafsson: Yes, you're right. I forgot that Tensor.Dispose() had that logic.

NiklasGustafsson:
Okay, I have no further issues with this PR. 

It makes it easy to review and accept when each PR is narrowly focused.

mvphelps:
Great I'm ready for merge if you are. Also any chance maybe this one could be a real release to Nuget? There is quite a bit stacked up and these are some pretty useful changes.

NiklasGustafsson:
Yes, it's probably enough to warrant a release.

NiklasGustafsson:
Oooh! You added test files, didn't you? In that case, you should **link** to them from the WithCUDABinaries test project.

NiklasGustafsson:
And... since you actually added an article to the docs folder -- I would love it if you could duplicate that content in the Wiki on this repo.

mvphelps:
I did, and was not aware I needed to do that. Do I open a new PR or push to here again?

Sure I'll add that document link to the wiki the correct placement is obvious.

NiklasGustafsson:
Since I already merged, you will have to create a new PR.

NiklasGustafsson:
And no 'New Page' button, either, presumably?

NiklasGustafsson:
I'll add the article.

mvphelps:
Yes correct no new page either. Thanks for adding. I assume that's for maintainers, I don't believe I have that much street cred yet. Thanks!

PR #1398, opened 2024-10-28
Replace native module implementation with managed code.

NiklasGustafsson:
This PR, which has been in the works for a while, replaces the implementation of many (but not all) built-in modules with an equivalent managed code implementation.

The rationale is primarily to follow what is done in Pytorch, where most modules are now implemented in Python. It also makes it easier to support access to a number of module 'attributes' in Pytorch, i.e. properties, which was harder to do with the native implementations, which did not support access to all such attributes.

This is a massive PR, but the patterns for replacements are just a few:

1. Change the constructors to not call the native code module factories. Importantly, the constructors must not call `RegisterComponents()` like most custom modules should.

2. Change the `forward` methods to invoke functional APIs instead of the native forward method.

3. Implement each 'attribute' using a .NET property with a setter and a getter. The backing fields should be nullable, even if the property itself is not, because the constructor is expected to use the property setter for initialization.


NiklasGustafsson:
And thank you, @shaltielshmid for all the help on getting this one ready.

PR #1411, opened 2024-11-15
concat dataset

yueyinqiu:
not ready

### DIFF
--- a/src/TorchSharp/Dataset.cs
+++ b/src/TorchSharp/Dataset.cs
@@ -25,9 +25,27 @@ namespace TorchSharp
                 }
 
                 /// <summary>
+                /// The base interface for all Datasets.
+                /// </summary>
+                public interface IDataset<out T> : IDisposable
+                {
+                    /// <summary>
+                    /// Size of dataset
+                    /// </summary>
+                    long Count { get; }
+
+                    /// <summary>
+                    /// Get tensor according to index
+                    /// </summary>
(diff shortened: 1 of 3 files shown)


yueyinqiu:
related to #1354 #1358 https://github.com/dotnet/TorchSharp/discussions/1348#discussioncomment-10035168

yueyinqiu:
As mentioned here https://github.com/dotnet/TorchSharp/pull/1357#issuecomment-2478038515, the current version has no API changes (except the type of dataset and collate_fn in `DataLoader<T, S>`). However we don't actually need `Dataset<T>`, `Dataset` and `IterableDataset`. Do we have any plan to remove them later? They may also occupy the position of some classes in PyTorch, like #1353.

src/TorchSharp/Dataset.cs
```diff
@@ -47,11 +45,35 @@ public void Dispose()
...
+            /// Size of dataset
+            /// </summary>
+            long Count { get; }
+
+            /// <summary>
+            /// Get tensor according to index
+            /// </summary>
+            /// <param name="index">Index for tensor</param>
+            /// <returns>Tensors of index. DataLoader will catenate these tensors into batches.</returns>
+            T this[long index] { get; }
```
yueyinqiu, line 76: Here we are using `Int64`. However actually most .net containers does not support `Int64` indices. And it actually is not practical to have so much data. Should we switch to `Int32` instead?

PR #1415, opened 2024-11-25
Fixing default print newline

### LINKED ISSUE #1414: tensor.print() - Missing default "newLine" Parameter
**Introduction:**

After the PR https://github.com/dotnet/TorchSharp/pull/635 ; default `newLine` character for `tensor.print()` became `""` from `"\n"`.

In the Jupyter notebook runs, output of the following code looks incorrect due to missing `newLine`:

Code:
```c#
var tensor = torch.ones(2,2);
tensor.print();
```
Output:
Image

===

When we especially add `newLine`, output is shown in correct format:

Code:
```c#
tensor.print(newLine:"\n");
```

Output:
Image

ghost:
Fixes #1414 

We're attempting to add back the default `newLine` as `\n` to have correct results in Jupyter notebooks.

### DIFF
--- a/test/TorchSharpTest/TestTorchTensor.cs
+++ b/test/TorchSharpTest/TestTorchTensor.cs
@@ -192,6 +192,29 @@ namespace TorchSharp
         }
 
         [Fact]
+        [TestOf(nameof(TensorExtensionMethods.print))]
+        public void TestTensorDefaultPrint()
+        {
+            // Arrange
+            Tensor t = torch.ones(2,2);
+
+            // Default parameters for the print
+            string expectedOutput = tensor.ToString(TensorStringStyle.Default, "g5", 100, null, "\n");
+
+            using (var sw = new StringWriter())
+            {
+                Console.SetOut(sw);
+
+                // Act
+                tensor.print();
+
+                // Assert
+                var result = sw.ToString().Trim();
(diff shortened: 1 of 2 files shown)


yueyinqiu:
But we have this line in `ToString`:

```csharp
if (String.IsNullOrEmpty(newLine))
    newLine = Environment.NewLine;
```

I'm wondering why this happens.

yueyinqiu:
Well I suppose what we should do is to switch the default value from `""` to `null`, and then it will use `torch.newLine`.

And... These lines are useless...

```csharp
if (String.IsNullOrEmpty(newLine))
    newLine = Environment.NewLine;
```

since we are later passing `nl` instead of `newLine`...

So maybe remove it?

test/TorchSharpTest/TestTorchTensor.cs
```diff
@@ -191,6 +191,29 @@ public void Test3DToJuliaString()
...
+            {
+                Console.SetOut(sw);
+
+                // Act
+                tensor.print();
+
+                // Assert
+                var result = sw.ToString().Trim();
+                Assert.Equal(expectedOutput, result);
+            }
```
yueyinqiu, line 214: By the way maybe we shall set the original console.out back? I don't know ((

yueyinqiu:
Oh I didn't notice that.

PR #1420, opened 2024-12-03
switch default newlines to `null` (`torch.newLine`) from `\n`

yueyinqiu:


yueyinqiu:
By the way, `NU1903` occurs on my machine (and warnings are regarded as errors), so I can't run the code.

NiklasGustafsson:
Please try this in a VS Code notebook, with both `.print()` and `.str()`

yueyinqiu:
I believe it should work:

image

environment: 

image

on Windows

NiklasGustafsson (approved):

PR #1428, opened 2024-12-26
Refactor Normalize method like pytorch.

AhmedZero:
Refactor Normalize method like pytorch to support different Tensor Image (C, H, W) and Batch of Tensor Images (B, C, H, W).

### DIFF
--- a/src/TorchVision/Functional.cs
+++ b/src/TorchVision/Functional.cs
@@ -24,6 +24,19 @@ namespace TorchSharp
         {
             public static partial class functional
             {
+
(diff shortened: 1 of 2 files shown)


whuanle:
This is exactly what I need, currently TorchSharp in image loading and transformer tensor need to manually adjust the shape to adapt to dimensional changes.

AhmedZero:
@dotnet-policy-service agree

AhmedZero:
@alinpahontu2912 Could you please take a look at this PR when you have time? Your feedback would be appreciated. Thanks!

PR #1439, opened 2025-02-03
Fix LinearLR end_factor default.

hiyuh:
`torch.optim.lr_scheduler.LinearLR` `end_factor` default was `5`, but should be `1.0`.
Most probably, it would be a typo influenced from `total_iters` default (`5`).
See also https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.LinearLR.html

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -6,6 +6,8 @@ Releases, starting with 9/2/2021, are listed with the most recent release at the
 __Bug Fixes__:
 
 #1426 Sequential.eval() does not put model into eval mode<br/>
+`torch.optim.lr_scheduler.LinearLR` `end_factor` default has been corrected, is now 1.0.<br/>
+
 # NuGet Version 0.105.0
 
 Move to libtorch 2.5.1. As with the 2.4.0 release, MacOS / Intel is no longer supported by libtorch, so TorchSharp doesn, either.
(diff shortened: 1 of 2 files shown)


ds5678 (approved):

alinpahontu2912 (changes requested):
LGTM, thanks for the catch, can you also update the RELEASENOTES ?

hiyuh:
@alinpahontu2912 
* added `RELEASENOTES.md` update after rebase.
  + according to `CONTRIBUTING.md`, rebase is preferred, instead of merge commit.
  + however, it looks you're doing a merge commit in your branch.
    - https://github.com/alinpahontu2912/TorchSharp/commit/e7e07ea3fe486701d489d84f28b08085f0f09be2
  + i'm not sure which is true preference.

RELEASENOTES.md
```diff
@@ -6,6 +6,8 @@ Releases, starting with 9/2/2021, are listed with the most recent release at the
 __Bug Fixes__:
 
 #1426 Sequential.eval() does not put model into eval mode<br/>
+#1439 `torch.optim.lr_scheduler.LinearLR` `end_factor` default has been corrected, is now 1.0.<br/>
```
alinpahontu2912, line 9: The number should point to an issue or discussion thread, since this is not one, I say we follow the pattern from here: https://github.com/dotnet/TorchSharp/blob/5ba14022749a20b98798105a32988da59f2356ca/RELEASENOTES.md?plain=1#L287

alinpahontu2912:
Hey @hiyuh, you are right about the git rebase - it is preferred. I'll also follow that going forward.

RELEASENOTES.md
hiyuh: thanks, removed the number.

alinpahontu2912 (approved):
LGTM, thanks for the fix :)

Blackhex (approved):
LGTM as well

PR #1447, opened 2025-02-14
update README to point to Furnace project

houstonhaynes:
update README to point to Furnace project


houstonhaynes:
> Hey @houstonhaynes , thanks for your time.
> 
> May we kindly ask for more context regarding this change request?

AFAIK DiffSharp is not maintained. Furnace is a hard fork under active development in fsprojects. It's a simple swap. I just happened to spot the link as I was catching up on your docs and was doing other housekeeping while setting up the new fork.

houstonhaynes:
No it's just a hard fork with no language support changes. But feel free to leave the original reference there. When users find it doesn't compile without giant vulnerabilities showing up because it hasn't been updated, they can check out Furnace. 😉

PR #1452, opened 2025-02-26
WIP - .NET8 upgrade

alinpahontu2912:
Upgrade build from target framework .Net6 to .Net8


alinpahontu2912:
Packages relying on TorchSharp: 
DiffSharp: **deprecated**: https://www.nuget.org/packages?q=diffsharp
Microsoft.ML.TorchSharp: https://www.nuget.org/packages/Microsoft.ML.TorchSharp/, uses .NETStandard 2, so should be compatible with this upgrade

alinpahontu2912:
Will revert last commit ( upgrde ImageSharp package version https://www.nuget.org/packages/SixLabors.ImageSharp#versions-body-tab ) and put in a diferent PR, with extra releasenote update. Build started failing and it was because Azure pipelines refused to use a vulnerable package to build. Similar thing happened before: https://github.com/dotnet/TorchSharp/blob/5ba14022749a20b98798105a32988da59f2356ca/RELEASENOTES.md?plain=1#L288

PR #1461, opened 2025-03-19
add string normalize method

alinpahontu2912:
Updated hardcoded string matching


src/TorchSharp/Torch.cs
```diff
@@ -168,16 +188,17 @@ private static void LoadNativeBackend(bool useCudaBackend, out StringBuilder? tr
 
                     if (torchsharpLoc!.Contains("torchsharp") && torchsharpLoc.Contains("lib") && Directory.Exists(packagesDir) && Directory.Exists(torchsharpHome)) {
 
-                        var torchSharpVersion = Path.GetFileName(torchsharpHome); // really GetDirectoryName
-
+                        var assembly = typeof(torch).Assembly;
+                        var version = assembly.GetName().Version;
+                        var torchSharpVersion = (version != null) ? version.ToString() : Path.GetFileName(torchsharpHome);
```
alinpahontu2912, line 193: When taking the version string directly from the library it results in a 4 number version string: eg: 0.105.0.0, which needs to be normalized to remove last 0. I think it's better to have it taken straight from the dll that was loaded to the project than the nugget install folder.

PR #1462, opened 2025-03-19
Increase Version Number

alinpahontu2912:
Increase version number from 0.105.0 to 0.105.1


alinpahontu2912:
Only thing needed for update version, similar to here: https://github.com/dotnet/TorchSharp/commit/a046a6620deb48c0a815f98fc6f0abc0f3d6c053

PR #1499, opened 2025-10-22
Add support for torch.export exported models

### LINKED ISSUE #1498: Add support for torch.export models
TorchSharp currently lacks support for loading and executing PyTorch models exported via `torch.export` (ExportedProgram format, `.pt2` files). While TorchSharp supports TorchScript (JIT) models via `torch.jit.load()`, it cannot load the newer `torch.export` format introduced in PyTorch 2.x. This is becoming critical as the PyTorch ecosystem, including projects like GR, is transitioning from ONNX to `torch.export` for model deployment.

`torch.export` is PyTorch 2.x's modern, recommended approach for model export, using TorchDynamo for symbolic tracing and producing optimized single-graph representations with only ATen-level operations.

- **Link for `PyTorch` implementation:**

  - [torch.export Documentation](https://docs.pytorch.org/docs/stable/export.html)
  - [torch.export Tutorial](https://docs.pytorch.org/tutorials/intermediate/torch_export_tutorial.html)
  - [torch.export API Reference](https://docs.pytorch.org/docs/stable/export/api_reference.html)
  - [PT2 Archive Spec](https://docs.pytorch.org/docs/stable/export/pt2_archive.html)

- **Example code for `PyTorch` implementation:**

```python
import torch
import torch.nn as nn

# Define a simple model
class MyModule(nn.Module):
    def __init__(self):
        super().__init__()
        self.linear = nn.Linear(10, 5)

    def forward(self, x):
        return self.linear(x)

# Export the model
model = MyModule()
example_inputs = (torch.randn(1, 10),)
exported_program = torch.export.export(model, example_inputs)

# Save to .pt2 file
torch.export.save(exported_program, "model.pt2")

# Load and run the exported model
loaded_program = torch.export.load("model.pt2")
result = loaded_program.module()(torch.randn(1, 10))
```

**Describe the solution you'd like**

Add support for loading and executing `.pt2` ExportedProgram models in TorchSharp, similar to how `torch.jit.load()` works for TorchScript models. The solution should include:

1. **Loading API** - Ability to load `.pt2` files:
```csharp
// Option 1: New namespace
using var exportedProgram = torch.export.load("model.pt2");

// Option 2: Integrate with existing jit.load()
using var module = torch.jit.load("model.pt2");
```

2. **Execution API** - Ability to run forward pass:
```csharp
var inputTensor = torch.randn(1, 10);
var result = exportedProgram.forward(inputTensor);
```

3. **Implementation approach**:
   - Leverage existing TorchSharp/ScriptModule infrastructure (both use similar three-layer architecture: C# → PInvoke → C++ wrapper → LibTorch)
   - Wrap PyTorch C++ API's `torch::export::load()` function
   - Reuse existing TensorOrScalar marshalling system for argument/return value handling
   - Support device management (CPU/CUDA)

**Technical considerations:**
- LibTorch C++ API already supports `torch::export::load()` and ExportedProgram execution
- `.pt2` files are ZIP archives containing serialized graph data
- Can follow the same implementation pattern as ScriptModule (see `src/TorchSharp/JIT/ScriptModule.cs` and `src/Native/LibTorchSharp/THSJIT.cpp`)

**Environment Information:**
 - OS: All platforms (Windows, Linux, MacOS)
 - Package Type: All packages (torchsharp-cpu, torchsharp-cuda-windows, torchsharp-cuda-linux)
 - Version: Latest (currently 0.105.x) and future versions

**Additional context**

**Why this matters:**
- `torch.export` is the recommended path forward for PyTorch model deployment (PyTorch 2.x+)
- GR and other projects are moving away from ONNX to torch.export
- Without this support, .NET developers cannot consume the latest PyTorch model formats
- TorchScript (JIT) is becoming legacy; torch.export is the future

**Related Issues:**
- #782 - Support for non Tensor types for ScriptModule
- #771 - Support more types for ScriptModule (closed)
- #644 - Load TorchScript modules (merged) - Similar pattern can be followed
- #1281 - Will TorchSharp support exporting ONNX models in the future? (closed)

**Comparison: TorchScript vs torch.export:**
| Feature | TorchScript (JIT) | torch.export |
|---------|-------------------|--------------|
| File Format | `.pt` | `.pt2` |
| PyTorch Era | 1.x | 2.x+ (current) |
| Status in TorchSharp | ✅ Supported | ❌ Not supported |
| Use Case | Research | Production deployment |

tolleybot:
# Add support for torch.export exported models (#1498)

Implements functionality to load and execute PyTorch models exported via torch.export (.pt2 files), enabling .NET applications to run ExportedProgram models as the PyTorch ecosystem transitions from ONNX to torch.export.

## Summary

This PR adds support for loading and running AOTInductor-compiled `.pt2` models in TorchSharp using `torch::inductor::AOTIModelPackageLoader` from LibTorch 2.9+.

**Key Points:**
- ✅ Inference-only API (no training support)
- ✅ Models must be compiled with `torch._inductor.aoti_compile_and_package()` in Python
- ✅ 30-40% better latency than TorchScript (according to PyTorch docs)
- ✅ Compatible with LibTorch 2.9+ which includes AOTIModelPackageLoader symbols

### Native Layer (C++)

**Files:**
- `src/Native/LibTorchSharp/Utils.h` - Added AOTIModelPackageLoader header include
- `src/Native/LibTorchSharp/THSExport.h` - C++ API declarations
- `src/Native/LibTorchSharp/THSExport.cpp` - Implementation using `torch::inductor::AOTIModelPackageLoader`

**Key Changes:**
```cpp
// Utils.h - Added header include for all files
#include "torch/csrc/inductor/aoti_package/model_package_loader.h"

// THSExport.cpp - Simple wrapper around AOTIModelPackageLoader
ExportedProgramModule THSExport_load(const char* filename)
{
    auto* loader = new torch::inductor::AOTIModelPackageLoader(filename);
    return loader;
}

void THSExport_Module_run(
    const ExportedProgramModule module,
    const Tensor* input_tensors,
    const int input_length,
    Tensor** result_tensors,
    int* result_length)
{
    std::vector<torch::Tensor> inputs;
    // ... convert inputs
    std::vector<torch::Tensor> outputs = module->run(inputs);
    // ... convert outputs
}
```

### Managed Layer (C#)

**Files:**
- `src/TorchSharp/PInvoke/LibTorchSharp.THSExport.cs` - PInvoke declarations
- `src/TorchSharp/Export/ExportedProgram.cs` - High-level C# API

**API Design:**
```csharp
// Basic usage
using var exported = torch.export.load("model.pt2");
var results = exported.run(input);

// Generic typing for single tensor output
using var exported = torch.export.load<Tensor>("model.pt2");
Tensor result = exported.run(input);

// Generic typing for tuple output
using var exported = torch.export.load<(Tensor, Tensor)>("model.pt2");
var (sum, diff) = exported.run(x, y);
```

**Features:**
- Implements `IDisposable` for proper resource cleanup
- Generic `ExportedProgram<TResult>` for type-safe returns
- Support for single tensors, arrays, and tuples (up to 3 elements)
- `run()`, `forward()`, and `call()` methods (all equivalent)

### Testing

**Files:**
- `test/TorchSharpTest/TestExport.cs` - 7 comprehensive unit tests
- `test/TorchSharpTest/generate_export_models.py` - Python script to generate test models
- `test/TorchSharpTest/*.pt2` - 6 test models

**Test Coverage:**
```csharp
[Fact] public void TestLoadExport_SimpleLinear()       // Basic model
[Fact] public void TestLoadExport_LinearReLU()         // Multi-layer
[Fact] public void TestLoadExport_TwoInputs()          // Multiple inputs
[Fact] public void TestLoadExport_TupleOutput()        // Tuple return
[Fact] public void TestLoadExport_ListOutput()         // Array return
[Fact] public void TestLoadExport_Sequential()         // Complex model
[Fact] public void TestExport_LoadNonExistentFile()    // Error handling
```

All 7 tests pass successfully.

### Dependencies

**Updated:**
- `build/Dependencies.props` - Updated LibTorch from 2.7.1 to 2.9.0

LibTorch 2.9.0 includes the `torch::inductor::AOTIModelPackageLoader` implementation that was previously only available in PyTorch source code.

### Two .pt2 Formats

PyTorch has two different .pt2 export formats:

1. **Python-only** (from `torch.export.save()`):
   - Cannot be loaded in C++
   - Uses pickle-based serialization
   - NOT supported by this implementation

2. **AOTInductor-compiled** (from `torch._inductor.aoti_compile_and_package()`):
   - Can be loaded in C++ via AOTIModelPackageLoader
   - Ahead-of-time compiled for specific device
   - ✅ Supported by this implementation

### Python Model Generation

To create compatible .pt2 files:

```python
import torch
import torch._inductor

model = MyModule()
example_inputs = (torch.randn(1, 10),)

# Export the model
exported = torch.export.export(model, example_inputs)

# Compile with AOTInductor for C++ compatibility
torch._inductor.aoti_compile_and_package(
    exported,
    package_path="model.pt2"
)
```

### Limitations

- **Inference only**: No training, no parameter updates, no gradient computation
- **Device-specific**: Models compiled for CPU cannot run on CUDA and vice versa
- **No device movement**: Cannot move model between devices at runtime
- **LibTorch 2.9+ required**: Older versions don't include AOTIModelPackageLoader

### Performance

According to PyTorch documentation, AOTInductor provides:
- 30-40% better latency compared to TorchScript
- Optimized for production inference workloads
- Single-graph representation with only ATen-level operations

## Testing

```bash
# Build
dotnet build src/TorchSharp/TorchSharp.csproj

# Run tests
dotnet test test/TorchSharpTest/TorchSharpTest.csproj --filter "FullyQualifiedName~TestExport"
```

## Migration Guide

For users currently using TorchScript:

**Before (TorchScript):**
```python
# Python
torch.jit.save(traced_model, "model.pt")
```
```csharp
// C#
var module = torch.jit.load("model.pt");
var result = module.forward(input);
```

**After (torch.export):**
```python
# Python
import torch._inductor
exported = torch.export.export(model, example_inputs)
torch._inductor.aoti_compile_and_package(exported, package_path="model.pt2")
```
```csharp
// C#
using var exported = torch.export.load("model.pt2");
var result = exported.run(input);
```

## References

- Issue: #1498
- PyTorch torch.export docs: https://docs.pytorch.org/docs/stable/export.html
- AOTInductor docs: https://pytorch.org/docs/stable/torch.compiler_aot_inductor.html
- PyTorch source: `torch/csrc/inductor/aoti_package/model_package_loader.h`

Fixes #1498

### DIFF
--- a/src/Native/LibTorchSharp/THSExport.cpp
+++ b/src/Native/LibTorchSharp/THSExport.cpp
@@ -1,0 +1,56 @@
+// Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
+#include "THSExport.h"
+
+// torch.export support via AOTInductor
+// This uses torch::inductor::AOTIModelPackageLoader which is INFERENCE-ONLY
+// Models must be compiled with torch._inductor.aoti_compile_and_package() in Python
+
+ExportedProgramModule THSExport_load(const char* filename)
+{
+    CATCH(
+        // Load .pt2 file using AOTIModelPackageLoader
+        // This requires models to be compiled with aoti_compile_and_package()
+        auto* loader = new torch::inductor::AOTIModelPackageLoader(filename);
+        return loader;
+    );
+
+    return nullptr;
+}
+
+void THSExport_Module_dispose(const ExportedProgramModule module)
+{
+    delete module;
+}
+
+void THSExport_Module_run(
+    const ExportedProgramModule module,
+    const Tensor* input_tensors,
+    const int input_length,
+    Tensor** result_tensors,
+    int64_t* result_length)
+{
+    CATCH(
+        // Convert input tensor pointers to std::vector<torch::Tensor>
+        std::vector<torch::Tensor> inputs;
+        inputs.reserve(input_length);
+        for (int i = 0; i < input_length; i++) {
+            inputs.push_back(*input_tensors[i]);
+        }
+
+        // Run inference
+        std::vector<torch::Tensor> outputs = module->run(inputs);
+
+        // Allocate output array and copy results
+        *result_length = static_cast<int64_t>(outputs.size());
+        *result_tensors = new Tensor[outputs.size()];
+
+        for (size_t i = 0; i < outputs.size(); i++) {
+            (*result_tensors)[i] = new torch::Tensor(outputs[i]);
+        }
+    );
+}
+
+void THSExport_Module_run_free_results(Tensor* result_tensors)
+{
+    delete[] result_tensors;
+}
--- a/src/Native/LibTorchSharp/THSExport.cpp
+++ b/src/Native/LibTorchSharp/THSExport.cpp
@@ -32,6 +32,9 @@ void THSExport_Module_run(
     *result_tensors = nullptr;
     *result_length = 0;
 
+    if (module == nullptr)
+        return;
+
     CATCH(
         // Convert input tensor pointers to std::vector<torch::Tensor>
         std::vector<torch::Tensor> inputs;
--- a/src/Native/LibTorchSharp/THSExport.cpp
+++ b/src/Native/LibTorchSharp/THSExport.cpp
@@ -29,6 +29,9 @@ void THSExport_Module_run(
     Tensor** result_tensors,
     int64_t* result_length)
 {
+    *result_tensors = nullptr;
+    *result_length = 0;
+
     CATCH(
         // Convert input tensor pointers to std::vector<torch::Tensor>
         std::vector<torch::Tensor> inputs;
@@ -41,12 +44,16 @@ void THSExport_Module_run(
         std::vector<torch::Tensor> outputs = module->run(inputs);
 
         // Allocate output array and copy results
-        *result_length = static_cast<int64_t>(outputs.size());
-        *result_tensors = new Tensor[outputs.size()];
+        auto count = outputs.size();
+        auto* tensors = new Tensor[count];
 
-        for (size_t i = 0; i < outputs.size(); i++) {
-            (*result_tensors)[i] = new torch::Tensor(outputs[i]);
+        for (size_t i = 0; i < count; i++) {
+            tensors[i] = new torch::Tensor(outputs[i]);
         }
+
+        // Only expose to caller after full success
+        *result_tensors = tensors;
+        *result_length = static_cast<int64_t>(count);
     );
 }
 
--- a/src/TorchSharp/Export/ExportedProgram.cs
+++ b/src/TorchSharp/Export/ExportedProgram.cs
@@ -1,0 +1,216 @@
+// Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
+
+using System;
+using System.Runtime.InteropServices;
+using TorchSharp.PInvoke;
+using static TorchSharp.PInvoke.NativeMethods;
+
+namespace TorchSharp
+{
+    public static partial class torch
+    {
+        public static partial class export
+        {
+            /// <summary>
+            /// Load a PyTorch ExportedProgram from a .pt2 file compiled with AOTInductor.
+            /// </summary>
+            /// <param name="filename">Path to the .pt2 file</param>
+            /// <returns>ExportedProgram model for inference</returns>
+            /// <remarks>
+            /// IMPORTANT: The .pt2 file must be compiled with torch._inductor.aoti_compile_and_package() in Python.
+            /// Models saved with torch.export.save() alone will NOT work - they require AOTInductor compilation.
+            ///
+            /// This implementation is INFERENCE-ONLY. Training, parameter updates, and device movement
+            /// are not supported. The model is compiled for a specific device (CPU/CUDA) at compile time.
+            ///
+            /// Example Python code to create compatible .pt2 files:
+            /// <code>
+            /// import torch
+            /// import torch._inductor
+            ///
+            /// # Export the model
+            /// exported = torch.export.export(model, example_inputs)
+            ///
+            /// # Compile with AOTInductor (required for C++ loading)
+            /// torch._inductor.aoti_compile_and_package(
+            ///     exported,
+            ///     package_path="model.pt2"
+            /// )
+            /// </code>
+            /// </remarks>
+            public static ExportedProgram load(string filename)
+            {
+                return new ExportedProgram(filename);
+            }
+
+            /// <summary>
+            /// Load a PyTorch ExportedProgram with typed output.
+            /// </summary>
+            public static ExportedProgram<TResult> load<TResult>(string filename)
+            {
+                return new ExportedProgram<TResult>(filename);
+            }
+        }
+    }
+
+    /// <summary>
+    /// Represents a PyTorch ExportedProgram loaded from an AOTInductor-compiled .pt2 file.
+    /// This is an INFERENCE-ONLY implementation - training and parameter updates are not supported.
+    /// </summary>
+    /// <remarks>
+    /// Unlike TorchScript models, ExportedProgram models are ahead-of-time (AOT) compiled for
+    /// a specific device and are optimized for inference performance. They provide 30-40% better
+    /// latency compared to TorchScript in many cases.
+    ///
+    /// Key limitations:
+    /// - Inference only (no training, no gradients)
+    /// - No parameter access or updates
+    /// - No device movement (compiled for specific device)
+    /// - No dynamic model structure changes
+    ///
+    /// Use torch.jit for models that require training or dynamic behavior.
+    /// </remarks>
+    public class ExportedProgram : IDisposable
+    {
+        private IntPtr handle;
+        private bool _disposed = false;
+
+        internal ExportedProgram(string filename)
+        {
+            handle = THSExport_load(filename);
+            if (handle == IntPtr.Zero)
+                torch.CheckForErrors();
+        }
+
+        /// <summary>
+        /// Run inference on the model with the given input tensors.
+        /// </summary>
+        /// <param name="inputs">Input tensors for the model</param>
+        /// <returns>Array of output tensors</returns>
+        /// <remarks>
+        /// The number and shapes of inputs must match what the model was exported with.
+        /// All inputs must be on the same device that the model was compiled for.
+        /// </remarks>
+        public torch.Tensor[] run(params torch.Tensor[] inputs)
+        {
+            if (_disposed)
+                throw new ObjectDisposedException(nameof(ExportedProgram));
+
+            // Convert managed tensors to IntPtr array
+            IntPtr[] input_handles = new IntPtr[inputs.Length];
+            for (int i = 0; i < inputs.Length; i++)
+            {
+                input_handles[i] = inputs[i].Handle;
+            }
+
+            // Call native run method
+            THSExport_Module_run(handle, input_handles, inputs.Length, out IntPtr result_ptr, out long result_length);
+            torch.CheckForErrors();
+
+            // Marshal result array
+            int count = (int)result_length;
+            torch.Tensor[] results = new torch.Tensor[count];
+            IntPtr[] result_handles = new IntPtr[count];
+            Marshal.Copy(result_ptr, result_handles, 0, count);
+
+            for (int i = 0; i < count; i++)
+            {
+                results[i] = new torch.Tensor(result_handles[i]);
+            }
+
+            // Free the native array (tensors are now owned by managed Tensor objects)
+            THSExport_Module_run_free_results(result_ptr);
+
+            return results;
+        }
+
+        /// <summary>
+        /// Synonym for run() - executes forward pass.
+        /// </summary>
+        public torch.Tensor[] forward(params torch.Tensor[] inputs) => run(inputs);
+
+        /// <summary>
+        /// Synonym for run() - executes the model.
+        /// </summary>
+        public torch.Tensor[] call(params torch.Tensor[] inputs) => run(inputs);
+
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!_disposed)
+            {
+                if (handle != IntPtr.Zero)
+                {
+                    THSExport_Module_dispose(handle);
+                    handle = IntPtr.Zero;
+                }
+                _disposed = true;
+            }
+        }
+
+        ~ExportedProgram()
+        {
+            Dispose(false);
+        }
+    }
+
+    /// <summary>
+    /// Generic version of ExportedProgram with typed output.
+    /// </summary>
+    /// <typeparam name="TResult">The return type (Tensor, Tensor[], or tuple of Tensors)</typeparam>
+    public class ExportedProgram<TResult> : ExportedProgram
+    {
+        internal ExportedProgram(string filename) : base(filename)
+        {
+        }
+
+        /// <summary>
+        /// Run inference with typed return value.
+        /// </summary>
+        public new TResult run(params torch.Tensor[] inputs)
+        {
+            var results = base.run(inputs);
+
+            // Handle different return types
+            if (typeof(TResult) == typeof(torch.Tensor))
+            {
+                if (results.Length != 1)
+                    throw new InvalidOperationException($"Expected 1 output tensor, got {results.Length}");
+                return (TResult)(object)results[0];
+            }
+
+            if (typeof(TResult) == typeof(torch.Tensor[]))
+            {
+                return (TResult)(object)results;
+            }
+
+            // Handle tuple types
+            if (typeof(TResult).IsGenericType)
+            {
+                var genericType = typeof(TResult).GetGenericTypeDefinition();
+                if (genericType == typeof(ValueTuple<,>))
+                {
+                    if (results.Length != 2)
+                        throw new InvalidOperationException($"Expected 2 output tensors, got {results.Length}");
+                    return (TResult)Activator.CreateInstance(typeof(TResult), results[0], results[1]);
+                }
+                if (genericType == typeof(ValueTuple<,,>))
+                {
+                    if (results.Length != 3)
+                        throw new InvalidOperationException($"Expected 3 output tensors, got {results.Length}");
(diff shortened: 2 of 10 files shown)


masaru-kimura-hacarus:
@tolleybot 
* i'll attached a report created by Deep Research enabled Google Gemini 2.5 Pro, to answer "why libtorch-cxx11-abi-shared-with-deps-2.9.0+cpu.zip doesn't exists".
  [Technical Analysis of the LibTorch ZIP File Naming Convention Change.pdf](https://github.com/user-attachments/files/23250624/Technical.Analysis.of.the.LibTorch.ZIP.File.Naming.Convention.Change.pdf)
  + as the executive summary said;
     > since PyTorch version 2.8.0, filenames in the format libtorch-cxx11-abi-shared-with-deps-VERSION.zip are no longer present, having been replaced by a unified format: libtorch-shared-with-deps-VERSION.zip.

alinpahontu2912:
Hey @tolleybot can you also address the copilot comments ? Also, there are some test failures from TestExport

src/TorchSharp/Export/ExportedProgram.cs
```diff
@@ -0,0 +1,215 @@
...
+            Marshal.Copy(result_ptr, result_handles, 0, result_length);
+
+            for (int i = 0; i < result_length; i++)
+            {
+                results[i] = new torch.Tensor(result_handles[i]);
+            }
+
+            // Free the native array (tensors are now owned by managed Tensor objects)
+            Marshal.FreeHGlobal(result_ptr);
+
```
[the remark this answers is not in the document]
tolleybot: Fixed. Added a dedicated `THSExport_Module_run_free_results()` native function that uses `delete[]` to free the array. The C# side now calls this instead of `Marshal.FreeHGlobal`.

src/Native/LibTorchSharp/THSExport.cpp
```diff
@@ -0,0 +1,51 @@
...
+            inputs.push_back(*input_tensors[i]);
+        }
+
+        // Run inference
+        std::vector<torch::Tensor> outputs = module->run(inputs);
+
+        // Allocate output array and copy results
+        *result_length = outputs.size();
+        *result_tensors = new Tensor[outputs.size()];
+
```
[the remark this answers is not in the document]
tolleybot: Fixed both issues. Added `THSExport_Module_run_free_results()` for proper `delete[]` cleanup, and changed `result_length` from `int` to `int64_t` to avoid truncation.

alinpahontu2912:
Hey @tolleybot, thanks for the work. Unfortunately we can not run macos tests in our pipelines, so we would need to have a strategy that allows running on both ubuntu/windows machines. I am also not a fan of adding multiple .pt2 files to only use for testing that export works. I think it won't scale nicely. Can you think of a solution for this ?

masaru-kimura-hacarus:
@tolleybot 

> I've just pushed a commit that removes all 7 `.pt2` binary files and ...

just my 2 cents,
* to commit unwelcomed binary files removal naively is known as antipattern that bloats git repository.
* a better way to remove binary files cleanly can be made by interactive rebase and etc to make surgical removal, then force push.

src/TorchSharp/Export/ExportedProgram.cs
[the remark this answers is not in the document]
tolleybot: Fixed — added `THSExport_Module_run_free_results()` that uses `delete[]`.

src/Native/LibTorchSharp/THSExport.cpp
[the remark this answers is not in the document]
tolleybot: Fixed — added `THSExport_Module_run_free_results()` with `delete[]`, and changed `result_length` to `int64_t`.

src/TorchSharp/Export/ExportedProgram.cs
```diff
@@ -0,0 +1,216 @@
+// Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
+
+using System;
+using System.Runtime.InteropServices;
+using TorchSharp.PInvoke;
```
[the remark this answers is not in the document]
tolleybot: Fixed — removed.

src/TorchSharp/Export/ExportedProgram.cs
```diff
@@ -0,0 +1,216 @@
...
+            Marshal.Copy(result_ptr, result_handles, 0, count);
+
+            for (int i = 0; i < count; i++)
+            {
+                results[i] = new torch.Tensor(result_handles[i]);
+            }
+
+            // Free the native array (tensors are now owned by managed Tensor objects)
+            THSExport_Module_run_free_results(result_ptr);
+
```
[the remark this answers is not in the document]
tolleybot: Fixed — wrapped in try/finally.

src/TorchSharp/Export/ExportedProgram.cs
```diff
@@ -0,0 +1,216 @@
...
+            for (int i = 0; i < inputs.Length; i++)
+            {
+                input_handles[i] = inputs[i].Handle;
+            }
+
+            // Call native run method
+            THSExport_Module_run(handle, input_handles, inputs.Length, out IntPtr result_ptr, out long result_length);
+            torch.CheckForErrors();
+
+            // Marshal result array
```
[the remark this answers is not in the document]
tolleybot: Fixed — added range check before casting.

src/TorchSharp/Export/ExportedProgram.cs
```diff
@@ -0,0 +1,216 @@
...
+                {
+                    if (results.Length != 2)
+                        throw new InvalidOperationException($"Expected 2 output tensors, got {results.Length}");
+                    return (TResult)Activator.CreateInstance(typeof(TResult), results[0], results[1]);
+                }
+                if (genericType == typeof(ValueTuple<,,>))
+                {
+                    if (results.Length != 3)
+                        throw new InvalidOperationException($"Expected 3 output tensors, got {results.Length}");
+                    return (TResult)Activator.CreateInstance(typeof(TResult), results[0], results[1], results[2]);
```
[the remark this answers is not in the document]
tolleybot: Fixed — now validates generic type arguments are `torch.Tensor` and throws `NotSupportedException` otherwise.

src/Native/LibTorchSharp/THSExport.cpp
```diff
@@ -0,0 +1,56 @@
...
+        // Run inference
+        std::vector<torch::Tensor> outputs = module->run(inputs);
+
+        // Allocate output array and copy results
+        *result_length = static_cast<int64_t>(outputs.size());
+        *result_tensors = new Tensor[outputs.size()];
+
+        for (size_t i = 0; i < outputs.size(); i++) {
+            (*result_tensors)[i] = new torch::Tensor(outputs[i]);
```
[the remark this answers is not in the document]
tolleybot: Fixed — out-params initialized to `nullptr`/`0` before the try block, and only assigned after full success.

test/TorchSharpTest/TestExport.cs
```diff
@@ -0,0 +1,94 @@
...
+            }
+
+            // If somehow a program was created (shouldn't happen), dispose it twice
+            if (program != null)
+            {
+                program.Dispose();
+                program.Dispose(); // second dispose should not throw
+            }
+
+            // The fact that we reach here without crashing validates idempotent cleanup
```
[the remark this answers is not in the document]
tolleybot: Removed — the test never exercised the dispose path since the constructor throws before an object is assigned. Idempotent dispose will get proper coverage once end-to-end tests with real models are added.

test/TorchSharpTest/TestExport.cs
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
+using System;
```
alinpahontu2912, line 4: can we make a test where we create and then load a model?

src/TorchSharp/Export/ExportedProgram.cs
```diff
@@ -0,0 +1,238 @@
...
+    /// - Inference only (no training, no gradients)
+    /// - No parameter access or updates
+    /// - No device movement (compiled for specific device)
+    /// - No dynamic model structure changes
+    ///
+    /// Use torch.jit for models that require training or dynamic behavior.
+    /// </remarks>
+    public class ExportedProgram : IDisposable
+    {
+        private IntPtr handle;
```
cincuranet: I would make sense to me to wrap this handle into a dedicated class that derives from `SafeHandle`.

test/TorchSharpTest/TestExport.cs
tolleybot: Unfortunately, creating a `.pt2` file can't be done purely from C#. It requires Python's `torch._inductor.aoti_compile_and_package()`, which has no C++ or C# equivalent in LibTorch. There's no way to export and AOT compile a model without the Python runtime.

One option would be a CI step that runs a small Python script to generate platform native `.pt2` models before the test phase. That would give real end-to-end load and run test coverage on each platform.

src/TorchSharp/Export/ExportedProgram.cs
tolleybot: Good suggestion. Wrapped the native handle in a dedicated `HType : SafeHandle` inner class, following the same pattern used by Module, Optimizer, and the other TorchSharp types.

src/TorchSharp/Export/ExportedProgram.cs
```diff
@@ -0,0 +1,262 @@
...
+        /// </remarks>
+        public torch.Tensor[] run(params torch.Tensor[] inputs)
+        {
+            if (handle.IsInvalid)
+                throw new ObjectDisposedException(nameof(ExportedProgram));
+
+            // Convert managed tensors to IntPtr array
+            IntPtr[] input_handles = new IntPtr[inputs.Length];
+            for (int i = 0; i < inputs.Length; i++)
+            {
```
alinpahontu2912, line 131: I think we should add some null checks here. Maybe something like: `if (inputs[i] is null)
         throw new ArgumentNullException($"inputs[{i}]");
     if (inputs[i].Handle == IntPtr.Zero)
         throw new ObjectDisposedException($"Input tensor at index {i}");
     input_handles[i] = inputs[i].Handle;`

src/Native/LibTorchSharp/THSExport.cpp
```diff
@@ -0,0 +1,63 @@
...
+void THSExport_Module_run(
+    const ExportedProgramModule module,
+    const Tensor* input_tensors,
+    const int input_length,
+    Tensor** result_tensors,
+    int64_t* result_length)
+{
+    *result_tensors = nullptr;
+    *result_length = 0;
+
```
alinpahontu2912, line 34: Should we add a null check for module here?

src/TorchSharp/Export/ExportedProgram.cs
```diff
@@ -0,0 +1,262 @@
+// Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
+
```
alinpahontu2912, line 3: would a nullable enable work here, we're using it in several other places too

src/TorchSharp/Export/ExportedProgram.cs
```diff
@@ -0,0 +1,262 @@
...
+            // Convert managed tensors to IntPtr array
+            IntPtr[] input_handles = new IntPtr[inputs.Length];
+            for (int i = 0; i < inputs.Length; i++)
+            {
+                input_handles[i] = inputs[i].Handle;
+            }
+
+            // Call native run method
+            THSExport_Module_run(handle.DangerousGetHandle(), input_handles, inputs.Length, out IntPtr result_ptr, out long result_length);
```
alinpahontu2912: Pass HType SafeHandle directly to P/Invoke instead of using DangerousGetHandle() The existing codebase (e.g. THSJIT_Module_dispose) passes the SafeHandle type directly, which lets the CLR ref-count the handle and prevent a race with concurrent Dispose(). Please update the P/Invoke signatures to accept ExportedProgram.HType instead of IntPtr, and remove the DangerousGetHandle() call.

test/TorchSharpTest/TestExport.cs
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) .NET Foundation and Contributors.  All Rights Reserved.  See LICENSE in the project root for license information.
```
alinpahontu2912, line 1: Since we can't run Python in CI and don't want to commit platform-specific .pt2 binaries, I think it might be a good idea to commit your generate_export_models.py script and add export_models to .gitignore. We would need some positive pass tests that only run locally. I can add notes in contributing/devguide or somehwere so developers know they should run these tests too if they want full coverage.

alinpahontu2912 (commented):
Thanks a lot for the contribution, looks good to me, I just have some concerns about possibly risky code.

src/TorchSharp/Export/ExportedProgram.cs
tolleybot: Added null and disposed-handle checks for each input tensor before marshalling.

src/Native/LibTorchSharp/THSExport.cpp
tolleybot: Added a null guard for the module pointer. The out-params are pre-initialized to nullptr/0 above it, so a null module safely no-ops.

src/TorchSharp/Export/ExportedProgram.cs
tolleybot: Added `#nullable enable` to the file, consistent with the rest of the codebase.

src/TorchSharp/Export/ExportedProgram.cs
tolleybot: Updated the P/Invoke signatures to accept `ExportedProgram.HType` directly and removed the `DangerousGetHandle()` call, matching the pattern used by Module and Optimizer.

test/TorchSharpTest/TestExport.cs
tolleybot: Restored `generate_export_models.py` and added a `.gitignore` for the generated `*.export.pt2` files. Added 7 local-only model tests using an `ExportModelFactAttribute` that skips with a clear message when the models aren't present. CI sees 5 passed + 7 skipped. Developers who run the script get all 12.

alinpahontu2912 (approved):
LGTM, thanks 👍

PR #1508, opened 2025-11-20
Remove invalid P/Invoke methods

ds5678:
These 3 methods have incorrect signatures for their entrypoint. Each of them has an overload with the correct signature.

Edit: I also went through and found a bunch of P/Invoke methods that have no corresponding native export.

### DIFF
--- a/src/TorchSharp/Tensor/TensorTyped.handwritten.cs
+++ b/src/TorchSharp/Tensor/TensorTyped.handwritten.cs
@@ -70,7 +70,7 @@ namespace TorchSharp
                 sz.Add(2);
                 var size2 = sz.ToArray();
 
-                var genHandle = (generator is null) ? IntPtr.Zero : generator.Handle;
+                var genHandle = generator?.Handle ?? IntPtr.Zero;
 
                 unsafe {
                     fixed (long* psizes = size2) {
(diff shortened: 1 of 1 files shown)


src/TorchSharp/Tensor/TensorTyped.handwritten.cs
```diff
@@ -61,7 +61,7 @@ public static Tensor from(float real, float imaginary = 0.0f, torch.Device devic
             ///  Create a new tensor filled with random integer values taken from a uniform distribution in [0, max).
             ///  The real and imaginary parts will be filled independently of each other.
             /// </summary>
-            public static Tensor randint(long max, long[] size, torch.Device device = null, bool requires_grad = false)
+            public static Tensor randint(long low, long high, long[] size, torch.Device device = null, bool requires_grad = false, Generator generator = null)
```
masaru-kimura-hacarus: * this change needs update docstring from "[0, max)" to "[low, high)"?

src/TorchSharp/Tensor/TensorTyped.handwritten.cs
```diff
@@ -160,7 +162,7 @@ public static Tensor from(double real, double imaginary = 0.0f, torch.Device dev
...
```
masaru-kimura-hacarus: * ditto?

PR #1532, opened 2026-02-16
Improve performance on .NetFx

### LINKED ISSUE #546: TorchSharp is slower on NetFX comparing to Net Core

alinpahontu2912:
Fixes #546

Replace [MarshalAs(UnmanagedType.U1)] bool parameters and return values with plain byte in hot-path P/Invoke declarations to eliminate marshalling overhead on .NET Framework.

Non-blittable bool types require the runtime to generate marshalling stubs for each P/Invoke call. On .NET Framework this adds ns per call vs .NET Core. By using byte (a blittable type), the runtime can make direct native calls without marshalling, closing the performance gap.

Converted P/Invoke areas:
- Autograd: grad/backward (retain_graph, create_graph, allow_unused)
- Tensor: pooling, device transfer, gradient, reductions, sort/search
- NN: dropout, batch_norm, instance_norm, attention, loss functions
- Linear algebra: all bool params (keepdim, hermitian, check_errors)
- CUDA backend: all bool getters/setters
- JIT: module train/is_training
- Data: loader/iterator

Call sites updated to convert bool↔byte at the managed boundary:
- Parameters: (byte)(boolValue ? 1 : 0)
- Returns: nativeResult != 0

### DIFF
--- a/test/PInvokeBenchmark/PInvokeBenchmark.csproj
+++ b/test/PInvokeBenchmark/PInvokeBenchmark.csproj
@@ -1,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFrameworks>net6.0;net472</TargetFrameworks>
+    <LangVersion>9.0</LangVersion>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <StartupObject>PInvokeBenchmark.Program</StartupObject>
+  </PropertyGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\TorchSharp\TorchSharp.csproj" />
+  </ItemGroup>
+</Project>
--- a/test/PInvokeBenchmark/Program.cs
+++ b/test/PInvokeBenchmark/Program.cs
@@ -1,0 +1,110 @@
+using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+namespace PInvokeBenchmark
+{
+    /// <summary>
+    /// Measures P/Invoke marshalling overhead difference between
+    /// non-blittable [MarshalAs(UnmanagedType.U1)] bool and blittable byte.
+    /// Uses a minimal native DLL stub to isolate the marshalling cost.
+    /// Also runs TorchSharp API calls if the native library is available.
+    /// </summary>
+    class Program
+    {
+        const int WarmupIterations = 1000;
+        const int Iterations = 10_000_000;
+
+        static void Main(string[] args)
+        {
+            Console.WriteLine($"Runtime: {RuntimeInformation.FrameworkDescription}");
+            Console.WriteLine($"OS: {RuntimeInformation.OSDescription}");
+            Console.WriteLine($"Architecture: {RuntimeInformation.ProcessArchitecture}");
+            Console.WriteLine();
+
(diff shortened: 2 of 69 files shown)


test/PInvokeBenchmark/Program.cs
```diff
@@ -0,0 +1,110 @@
+using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+namespace PInvokeBenchmark
```
cincuranet, line 5: Any reason why _BenchmarkDotNet_ is not used here?

Maybe a more general question. We know this issue exists in .NET Framework and I highly doubt that anything is going to change there. Do we need the benchmark at all? I would post the numbers into the PR and possibly the code for benchmark. And that's it. If somebody wants to reevaluate the number (s)he can rerun it from PR and compare.

test/PInvokeBenchmark/PInvokeBenchmark.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFrameworks>net6.0;net472</TargetFrameworks>
+    <LangVersion>9.0</LangVersion>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <StartupObject>PInvokeBenchmark.Program</StartupObject>
```
cincuranet, line 7: I believe these two lines are not needed.

cincuranet (commented):
All these `x ? 1 : 0` made me wonder... Why not use `Convert.ToByte` instead?

alinpahontu2912:
Will take a look again. I'd say it's importatnt to figure out if it's worth spending time to fix this. (Not sure how to do that, as I can t see metrics on how much the package is used for each variant) The issue was opened a long ago when support to netstandard2.0 to TorchSharp was added. Users running on NETFX are expected to see some differences and I am not sure if we want to help much there. Tarek mentioned that we should encourage people to switch to .net core as much as possible. There is also this comment explaining the same:  https://github.com/dotnet/TorchSharp/issues/546#issuecomment-1435991796. What do you think @cincuranet ?

cincuranet:
The numbers that Tarek posted show about ~3.5ns overhead, which is ~80% slower. However, this is the raw interop cost only. In practice, the actual operations behind each P/Invoke call take orders of magnitude longer, so the marshalling overhead is noise for most workloads. It would only matter for extremely tight loops making millions of P/Invoke calls with trivial native work.

Bottom line. The real-world benefit is questionable. .NET Framework's P/Invoke gap is known. We don't have strong evidence of significant .NET Framework usage of TorchSharp. The #546 has basically non-existing community traction since 2022.

I would suggest closing this PR. And closing the issue, explaining that if you want best perf, you should use latest .NET.

alinpahontu2912:
Thanks, I will close the pr and the issue then

PR #1544, opened 2026-02-25
Add support for bfloat16 type

### LINKED ISSUE #1469: bfloat16 tensor cannot be printed
var a = torch.randn(3, 3).to(torch.bfloat16);
 a.print();

alinpahontu2912:
Fixes #1469


src/TorchSharp/BFloat16.cs
```diff
@@ -0,0 +1,128 @@
...
+#nullable enable
+namespace TorchSharp
+{
+    /// <summary>
+    /// Represents a 16-bit brain floating-point number (BFloat16).
+    /// Binary layout: 1 sign bit, 8 exponent bits, 7 mantissa bits — the upper 16 bits of IEEE 754 float32.
+    /// Binary-compatible with c10::BFloat16 in LibTorch.
+    /// </summary>
+    [StructLayout(LayoutKind.Sequential)]
+    public readonly struct BFloat16 : IComparable<BFloat16>, IEquatable<BFloat16>, IComparable, IFormattable
```
cincuranet, line 15: Is this `BFloat16` different from `System.Numerics.BFloat16` that's now available in [.NET 11](https://learn.microsoft.com/en-us/dotnet/api/system.numerics.bfloat16?view=net-11.0)?

cincuranet (approved):

src/TorchSharp/BFloat16.cs
alinpahontu2912: I have to check, but the problem is we are now building with .net 8 only and I don't think directly jumping to 11 would be okay

cincuranet: Right. But we could, only for .NET 11, exclude this and use the built-in type. And eventually removing this custom code completely.

alinpahontu2912: I will leave a note in releasenotes for that

PR #1554, opened 2026-03-20
Expose Storage.data_ptr() as public API

### LINKED ISSUE #1454: Expose `data_ptr` as a public API
This would be useful for using `Torchsharp.Tensor` with other cuda libraries, like `bitsandbytes`.

Code could be
```
﻿using System;
using System.Runtime.InteropServices;

namespace TorchSharp.BitsAndBytes; 
public static class LibTorchNativeMethod 
{ 
    [DllImport("LibTorchSharp")]
    public static extern IntPtr THSStorage_data_ptr(IntPtr tensor);
}
```

Usage
https://github.com/LittleLittleCloud/TorchSharp.BitsAndBytes/blob/4167aff6ea4d9f177beb8d5a6a3e3e2c4afc06a9/TorchSharp.BitsAndBytes/BitsAndByteUtils.cs#L81

alinpahontu2912:
Change the access modifier of Storage.data_ptr() from protected to public, allowing external consumers to obtain the raw data pointer for interop with other native libraries (e.g. bitsandbytes).

Fixes #1454

### DIFF
--- a/src/TorchSharp/Tensor/Storage.cs
+++ b/src/TorchSharp/Tensor/Storage.cs
@@ -139,7 +139,7 @@ namespace TorchSharp
             /// A pointer to the raw data in memory.
             /// </summary>
             /// <returns></returns>
-            protected IntPtr data_ptr()
+            public IntPtr data_ptr()
             {
                 if (_tensor_data_ptr != IntPtr.Zero)
                     return _tensor_data_ptr;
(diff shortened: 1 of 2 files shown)


alinpahontu2912:
Hey @LittleLittleCloud does this help you?

cincuranet (approved):

PR #1555, opened 2026-03-20
Fix native library loader silently failing when CUDA DLLs are missing

### LINKED ISSUE #1545: Silent Native Library Failure in torch::loadNativeBackEnd
**Describe the bug**
When the native back-end method fails to load a CUDA DLL such as cudnn_adv64_9, it will silently fail if LibTorchSharp appears to load.  The code will also set nativeBackendCudaLoaded to true thus skipping the loader when a second load would want to be attempted after the static constructor has done its bit.

Based on the comment it looks like the behavior may be intentional, but it makes debugging more difficult given that the stringbuilder trace's lost.

**To Reproduce**
Remove cudnn_adv64_9 and check whether cuda's available. When debugging this you'll see that the method has a chain of ok statements of the form
                        ok = TryLoadNativeLibraryByName("cudnn_adv64_9", typeof(torch).Assembly, trace);
where each step will overwrite the prior one. On my end by changing it to bool = ok is true and setting each library to use
                        ok &= TryLoadNativeLibraryByName("lib", typeof(torch).Assembly, trace);
for each of the files to load. This then causes torch to attempt to load again, thus allowing us to get obtain error messages from the library loader.

**Please complete the following information:**
 - OS: Windows 11
 - Package Type [e.g. torchsharp-cpu, torchsharp-cuda-windows, torchsharp-cuda-linux]
 - Version 0.105.2 (same in 0.106.0)

alinpahontu2912:
Change all 'ok = TryLoadNativeLibraryByName(...)' calls to 'ok &= ...' so that failures accumulate instead of being overwritten by subsequent successful loads. Initialize 'ok = true' before the loading chain.

Previously, each load call overwrote the result of the previous one, so if an early CUDA dependency (e.g. cudnn_adv64_9) failed to load but LibTorchSharp succeeded, 'ok' would be true. This caused:
- nativeBackendCudaLoaded set to true despite missing dependencies
- The fallback loading path was skipped
- The diagnostic trace (StringBuilder) was discarded
- Subsequent load attempts were skipped entirely
- CUDA operations failed later with cryptic errors

Now any single load failure keeps 'ok' as false, ensuring the fallback path is attempted and the full diagnostic trace is preserved in error messages.

Fixes dotnet/TorchSharp#1545

### DIFF
--- a/src/TorchSharp/Tensor/Factories/Tensor.Factories.cs
+++ b/src/TorchSharp/Tensor/Factories/Tensor.Factories.cs
@@ -649,7 +649,10 @@ namespace TorchSharp
         static torch()
         {
             deleters = new ConcurrentDictionary<TorchSharp.PInvoke.GCHandleDeleter, TorchSharp.PInvoke.GCHandleDeleter>();
-            TryInitializeDeviceType(DeviceType.CUDA);
+            if (!TryInitializeDeviceType(DeviceType.CUDA)) {
+                // CUDA not available, ensure CPU backend is loaded
+                LoadNativeBackend(false, out _);
+            }
         }
         #endregion
     }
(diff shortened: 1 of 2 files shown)


cincuranet (commented):
The description says the `ok &=` is used, but I don't see that in changes.

PR #1566, opened 2026-05-08
Remove duplicate log1p function declarations

ds5678:
They were already defined on lines 774 and 776.


alinpahontu2912 (approved):
Thanks  👍

PR #1568, opened 2026-05-26
Add documentation comment warning to only load trusted models

svick:
We should warn users not to load untrusted models.

This corresponds to [a similar warning in the PyTorch documentation](https://docs.pytorch.org/docs/2.12/generated/torch.load.html):

> `torch.load()` uses an unpickler under the hood. **Never load data from an untrusted source.**

Equivalent PR in ML.NET: https://github.com/dotnet/machinelearning/pull/7611


alinpahontu2912 (approved):
Thank you 👍

PR #1576, opened 2026-07-20
Add agentic maintenance as local Copilot skills

kotlarmilos:
## Description

This adds the TorchSharp CI scanner, scanner feedback, and libtorch bump automation as repository-local Copilot skills under `.github/skills/`. Local execution uses the operator's authenticated tools and keeps GitHub writes behind explicit approval while preserving the scanner methodology as progressive-disclosure references. This avoids PAT-pool and scheduled workflow configuration until repository billing and authentication are available. The deferred CI implementation remains in https://github.com/dotnet/TorchSharp/pull/1575 for a later follow-up.

### DIFF
--- a/.github/skills/ci-scan/references/playbook.md
+++ b/.github/skills/ci-scan/references/playbook.md
@@ -54,21 +54,6 @@ curl -s "$url" | tee /tmp/torchsharp-ci-scan/timeline.json | jq '.records | leng
 
 Reconstruct `Stage -> Phase -> Job -> Task` using `parentId`. A failed record with non-null `log.id` is a leaf.
 
-| Leg | Where the signature comes from |
-|---|---|
-| `Ubuntu_x64` | xUnit test log or compile error |
-| `MacOS_arm64` | xUnit test log or compile error |
-| `Windows_x64_NetCore` | xUnit test log or compile error |
-| `Windows_x64_NetFX` | xUnit test log or compile error |
-| `Windows_arm64` | compile error |
-| `Linux_Native_Build_For_Packages` | native CMake or compiler error |
-| `Windows_Native_Build_For_Packages` | native CMake or compiler error |
-| `Windows_arm64_Native_Build_For_Packages` | native CMake or compiler error |
-| `MacOS_arm64_Native_Build_For_Packages` | native CMake or compiler error |
-| `Build_TorchSharp_And_libtorch_cpu_Packages` | managed build or packaging error |
-| `Build_libtorch_cuda_linux_Packages`, `Build_libtorch_cuda_win_Packages` | usually large-download infrastructure noise |
-| `Push_*`, `CodeSign_*` | infrastructure noise |
-
 ## Step 3 - Classify each failure
 
 Classify every failed leaf using [Failure classification](ci-scan.instructions.md#classification). Save the canonical task log to `/tmp/torchsharp-ci-scan/failure.log`, count occurrences by distinct build ID, and apply the stability gate from [Occurrence counting and window widening](ci-scan.instructions.md#occurrence-counting).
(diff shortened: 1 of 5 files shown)


.github/skills/ci-scan/references/playbook.md
```diff
@@ -0,0 +1,100 @@
...
+| `Windows_x64_NetCore` | xUnit test log or compile error |
+| `Windows_x64_NetFX` | xUnit test log or compile error |
+| `Windows_arm64` | compile error |
+| `Linux_Native_Build_For_Packages` | native CMake or compiler error |
+| `Windows_Native_Build_For_Packages` | native CMake or compiler error |
+| `Windows_arm64_Native_Build_For_Packages` | native CMake or compiler error |
+| `MacOS_arm64_Native_Build_For_Packages` | native CMake or compiler error |
+| `Build_TorchSharp_And_libtorch_cpu_Packages` | managed build or packaging error |
+| `Build_libtorch_cuda_linux_Packages`, `Build_libtorch_cuda_win_Packages` | usually large-download infrastructure noise |
+| `Push_*`, `CodeSign_*` | infrastructure noise |
```
matouskozak: Out of curiosity, what's the purpose of listing these?

.github/skills/libtorch-bump/SKILL.md
```diff
@@ -0,0 +1,52 @@
...
+# LibTorch version bump
+
+Detect a new stable libtorch release and prepare either a patch-level code change or a major or minor upgrade report.
+
+## Hard rules
+
+1. Do not modify `src/Native/` or `src/TorchSharp/PInvoke/`.
+2. Automate only patch updates within the current `major.minor`. Major or minor changes require human-driven bindings work.
+3. Do not duplicate an existing bump PR or a recently closed major-version tracking issue.
+4. Never change the macOS x64 conditional pins. They currently use `2.2.2` and `2.2.2.0`.
```
matouskozak, line 15: What's the background here to not update it?

matouskozak (approved):

.github/skills/ci-scan/references/playbook.md
kotlarmilos: It is a lookup that drives signature extraction. Each CI leg surfaces its useful failure signal in a different place, so this tells the scanner where to read the signature for that leg

.github/skills/libtorch-bump/SKILL.md
kotlarmilos: PyTorch stopped shipping official macOS x86_64 binaries after 2.2.2, so 2.3.0 and later have no Intel Mac libtorch packs

.github/skills/ci-scan/references/playbook.md
matouskozak: Did we find it necessary for quality outputs? I'm just asking as this is going to be something that has to be potentially maintained going forward as things change.

kotlarmilos: Not strictly necessary for correctness based on the past logs. The real classification runs on content in failure classification, which decides build break, test failure, or infra noise from the failing tests, independent of leg names, so the scan still extracts the right signature when a leg is renamed or added. The table is a hint on top of that.

On maintenance, I can remove the table.

kotlarmilos: Removed

matouskozak: I think both option are fine, I just wanted to make sure we don't put too much stuff in which is not necessary. Feel free to keep it if you found success with it.

kotlarmilos: I would rather keep it in case we determine it is needed, so let's proceed without it for now. Applied the same approach in the machinelearning PR.

rosebyte (approved):

